PR #2, opened 2015-09-17
Add unittests for http and tcp.

brian-brazil:
Fix bug in default status codes accepted by http.
Allow for how redirect errors are handled.

@fabxc


fabxc:
Much tested :+1: \o/

PR #7, opened 2015-10-20
Have the example url match the example config.

lyda:


brian-brazil:
Thanks, thought I'd gotten all these.

PR #8, opened 2015-10-20
Switch to go version 1.5.1.

lyda:
Now will match other prometheus projects.


brian-brazil:
Can you take this from the official version at https://github.com/prometheus/utils/ ?

lyda:
Ah! Better still. And since I looked at the node_exporter version of this file for how to change this, I updated it as well.

PR #15, opened 2015-11-24
README.md: Use a proper target URL

RichiH:


README.md
```diff
@@ -58,7 +58,7 @@ scrape_configs:
       module: [http_2xx]  # Look for a HTTP 200 response.
     target_groups:
       - targets:
-        - mywebsite.com   # Target to probe
+        - www.prometheus.io   # Target to probe
```
brian-brazil: prometheus.io is the canonical version. Next prometheus release we can put the http:// in front too.

jimmidyson: Let's hope loads of people don't try this out at once & DDoS prometheus.io :-b

brian-brazil: I think Github can handle it

PR #18, opened 2015-11-24
Feature/expand docs

RichiH:


brian-brazil:
Thanks for the PR, I'm not sure it makes sense to have 3 very similar examples one after another. A common on one example would probably be clearer.

RichiH:
I am not aware of a way to squash this more and neither is @jimmidyson afaik.

If you can share what to change, I will gladly do so. Or you are free to just toss this and update by yourself, whichever is less work.

jimmidyson:
One thing that was confusing for both @RichiH & I when we started with blackbox exporter was the fact you had to create a separate job for each module you wanted to use. I'd like to see that more explicitly stated in the docs, even in this PR.

brian-brazil:
> I am not aware of a way to squash this more and neither is @jimmidyson afaik.

I meant to have a single example as there is now, but with comments.

> I when we started with blackbox exporter was the fact you had to create a separate job for each module you wanted to use.

That's a general Prometheus thing, you should have a different job for every type of thing you're monitoring. I don't think it's right to cover that bit of knowledge in any one exporter in particular.

jimmidyson:
The only reason is that the blackbox exporter is "special" in the way that it works, unlike most other exporters. Seemed less intuitive, especially when `module` is a param which is multi-valued so could understandably look like it would accept multiple modules in one config.

RichiH:
> Seemed less intuitive, especially when module is a param which is multi-valued so could understandably look like it would accept multiple modules in one config.

That is _precisely_ what had me running against the wall. Well, and the fact that I used 0.16.1 which does not include https://github.com/prometheus/prometheus/pull/1204 .

brian-brazil:
This has been reworked since.

PR #21, opened 2015-12-01
tcp: query_response matching

stapelberg:


brian-brazil:
Generally looks good.

stapelberg:
Added the error message when qr.Expect cannot be compiled into a regular expression. PTAL.

brian-brazil:
Thanks!

Issue #5, opened 2015-09-23
whitelist / blacklist for HTTP and TCP probing targets?


iwinux:
There're some services that should not be exposed outside of `localhost` - if I deploy `blackbox_exporter` on these hosts, maybe it would get exploited somehow to access these internal services?

brian-brazil:
I was thinking a whitelist, mainly for when auth is added and you don't want to send credentials out to arbitrary hosts.

brian-brazil:
The general approach of Prometheus is not to get into authorisation, so to keep consistent with that this is not something we should really add. The new security model docs do mention the risks that come from that with this class of exporter.

PR #23, opened 2015-12-09
Adding support for basic auth

mbrewster:
Implemented support for basic auth for the http probe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ modules:
     http:
       valid_status_codes: []  # Defaults to 2xx
       method: GET
+      basic_auth_username: test_user
+      basic_auth_pwd: test_pwd
       no_follow_redirects: false
       fail_if_ssl: false
       fail_if_not_ssl: false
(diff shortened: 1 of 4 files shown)


brian-brazil:
We should probably factor out the auth support in Prometheus rather than slowly reimplementing it in other components.

mbrewster:
Would you envision the username/pwd being passed through via the scrape config?

brian-brazil:
Yes, the same way it's done in Prometheus. There's some security issues here touched on in #5

mbrewster:
So it would be a new config element unique to the blackbox_exporter that passed a username/pwd in a manner similar to how credentials are passed for the sd configs? I don't see how we get around actually implementing the auth header in the request on the blackbox side.

brian-brazil:
We can have https://github.com/prometheus/prometheus/blob/master/retrieval/target.go#L257 as a part of a shared library between the two systems.

PR #26, opened 2016-01-12
Add path to http request.

caskey:
This patch enables a path to be specified in the configuration.  It also contains a minor refactoring of the flow in evaluating the body of the response.

### DIFF
--- a/blackbox.yml
+++ b/blackbox.yml
@@ -3,6 +3,7 @@ modules:
     prober: http
     timeout: 5s
     http:
+      path: /
   http_post_2xx:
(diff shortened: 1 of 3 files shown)


brian-brazil:
Would you mind splitting this into smaller changes so it's easier to review?

All metrics always need to be returned, otherwise very odd things will happen (http://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics).

caskey:
Okay, I'll sort it out.

caskey:
I'll break it up.

Issue #10, opened 2015-11-11
Add diagnostic information to webserver


brian-brazil:
It'd be handy if the details of the last ~10 probes to each target/module were available via the web UI so you could look at things like HTTP responses, rather than just having the metrics on the Prometheus side

brian-brazil:
@tomwilkie Can you give me a pointer to your in-process Loki/opentracing stuff?

tomwilkie:
Sure!  Its a bit rough right now, but its here: https://github.com/weaveworks-experiments/loki/pull/7

I don't plan on doing any more work on it until post promcon, but it should give yo what you need.

PR #27, opened 2016-01-12
Add configurable path parameter and verify its use in the request.

caskey:
This diff is limited to the new path feature.

### DIFF
--- a/http.go
+++ b/http.go
@@ -76,8 +76,13 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 	if config.Method == "" {
 		config.Method = "GET"
 	}
+	if config.Path == "" {
+		config.Path = "/"
+	}
 
-	request, err := http.NewRequest(config.Method, target, nil)
+	log.Infof("probeHTTP to %s%s", target, config.Path)
+
+	request, err := http.NewRequest(config.Method, target + config.Path, nil)
 	if err != nil {
 		log.Errorf("Error creating request for target %s: %s", target, err)
 		return
--- a/blackbox.yml
+++ b/blackbox.yml
@@ -3,6 +3,7 @@ modules:
     prober: http
     timeout: 5s
     http:
+      path: /
   http_post_2xx:
     prober: http
     timeout: 5s
--- a/http_test.go
+++ b/http_test.go
@@ -55,6 +55,27 @@ func TestHTTPStatusCodes(t *testing.T) {
 	}
 }
 
+func TestConfiguredPathSentInRequest(t *testing.T) {
+	var pathToSend = "/path/to/send?query=string"
+	var pathFound string
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		pathFound = r.URL.Path
+	}))
+	defer ts.Close()
+
+	recorder := httptest.NewRecorder()
+	result := probeHTTP(ts.URL, recorder, Module{HTTP: HTTPProbe{Path: pathToSend}})
+	body := recorder.Body.String()
+	if !result {
+		t.Fatalf("Fetch test failed unexpectedly, got %s", body)
+	}
+	// The path parameter received by the server should be the first part of the path+query string.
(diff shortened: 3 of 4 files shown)


http.go
```diff
@@ -76,8 +76,13 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 	if config.Method == "" {
 		config.Method = "GET"
 	}
+	if config.Path == "" {
+		config.Path = "/"
+	}
+
+	log.Infof("probeHTTP to %s%s", target, config.Path)
```
brian-brazil: This is a debug log at best

caskey: Where I come from we log all service actions as a record of activity (audit log) and to aid in service analysis.

brian-brazil: I consider anything more than 1-2 info logs a minute spammy as it'll hide the interesting logs. This information is primarily useful for debugging, there's no point filling up the logs with it during normal operations.

caskey:
Okay.

brian-brazil:
You seem to have closed this PR by mistake.

caskey:
Sorry I wasn't clear, I closed it because it seems incompatible with your vision for the exporter and what its logging profile should look like. Adding flags to manage multiple levels of 'info' logging would be messy and tiresome.  Rather than add more configuration switches it seems easier to just drop the issue.  Sorry for the distraction.

brian-brazil:
You can still add the path change, which is the main purpose of this exporter.

#10 are my thoughts on how to expose debug information in this particular exporter.

caskey:
Theres no need, I've now fully moved to my environment to using my fork instead. I need it to work for my operational model where effective usage logging, instrumentation, and other features are important.

The change is small so you might want to put the feature in for your users. We find it quite handy, especially when monitoring edge proxy health URLs (root path just gets the proxied service) and matching internal server health status URLs.

It's probably for the best as our code base has been going  through significant refactoring to make the probe api simpler for writing new probes as well as boosting performance, concurrency, and memory usage under load. All of which is specific to our environment.

caskey:
Ugh that first sentence is a train wreck. But you get the idea.

brian-brazil:
We'd appreciate your improvements, everyone benefits from improved performance, concurrency,  memory usage and the various probing features.

We do have a different philosophy on logging, we believe that logs should be useful to a human looking at the log output under full load. Accordingly we believe the signal to noise ratio of logs is important.

You could add this as `Debugf` rather than `Infof`. Logging the module name and target would be more useful, as it covers all the settings.

caskey:
Yeah, we use other tools for managing log volume, querying, and correlation.  At even just a few thousand QPS it wouldn't be possible otherwise, especially when you are cross-correlating and tracing among multiple services. Because of our need to correlate behavior among production instances, by policy we require a minimum of one transaction log line per user request. We understand this is an internal issue and have no desire to force our perspective on your project so we're dealing with it ourselves.

As for your suggestion, debug logs are for a different purpose and giving up or conflating the semantic value of the different log levels would just create needless confusion. Highly undesirable during incident response. Especially when during development debug log lines elsewhere would be useful to the developers. In our environment these are not debugging statements but ordinary production activity logs.

Regarding logging the module and target, we realized that would be better too, but decided to refactor the api first to ensure consistency in log messages among modules before refactoring the transaction logs and initialization messages.

I agree everyone should benefit, and we're committed to not keeping our changes private. We're still publishing our changes back as an OSS project. That way if anyone wants to they can cherry pick the features they do like and incorporate them into the mainline. (I just pushed a bunch today.)

Anyway, with the multiple rounds of refactoring we've already completed, our code base's delta from the baseline has already moved way beyond this change set.

brian-brazil:
We consider these logs to be application logs, not request/transaction logs. The only places in the system it makes sense to have some level of request logs for everyone are in prometheus/alertmanager/pushgateway as they have both state mutating requests which need auditing and potentially expensive requests that need analysis.

I suggested the debug logs as a compromise, they're pretty close to what you want currently.

I look forwards to more PRs in the future.

joshuaspence:
It's a real shame that this PR was abandoned.

PR #29, opened 2016-01-22
Makefile.COMMON: Require Go 1.5.3

RichiH:


juliusv:
So the authoritative copy of this Makefile lives in https://github.com/prometheus/utils (from where the newest version is occasionally copied into individual repos). The version there already requires Go 1.5.3, but also has other changes that this repo here doesn't have yet.

The best course of action would be to copy the `Makefile.COMMON` from there with the commit message saying something like, "Update Makefile.COMMON from utils repo".

RichiH:
Will do. Thanks.

Richard

Sent by mobile; excuse my brevity.

PR #33, opened 2016-02-14
Support TLS in TCP connections

hudashot:
Also, report certificate expiration date just like HTTPS probe does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,14 @@ modules:
   tcp_connect:
     prober: tcp
     timeout: 5s
+  pop3s_banner:
+    prober: tcp
+    tcp:
+      query_response:
+      - expect: "^+OK"
+      tls: true
+      tls_config:
+        insecure_skip_verify: false
   ssh_banner:
     prober: tcp
     timeout: 5s
--- a/tcp.go
+++ b/tcp.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"crypto/tls"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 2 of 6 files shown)


tcp.go
```diff
@@ -11,19 +12,34 @@ import (
 	"github.com/prometheus/log"
 )
 
+func dialTCP(target string, ssl bool, timeout time.Duration) (net.Conn, error) {
+	dialer := &net.Dialer{Timeout: timeout}
+	if !ssl {
+		return dialer.Dial("tcp", target)
+	}
+	config := &tls.Config{InsecureSkipVerify: true}
```
brian-brazil: This needs to be secure by default.

Similar to #23, if we're going to be offering non-default SSL options they should be common across prometheus/blackbox_exporter/alertmanger.

hudashot: Not sure I understand what you mean by "common".

There are some comments in #23 about passing options from prometheus to blackbox_exporter using GET parameters to `/probe`. Is this how you suggest to set InsecureSkipVerify?

brian-brazil: It should all be in the config file.

There's no mention of GET in that PR, the only things that should be passed in are module name and target.

hudashot:
@brian-brazil, PTAL.

I think if `TLSConfig` in prometheus is supposed to be re-usable in other components it might be better to move logic which is currently in `httputil.NewTLSConfig` into a struct method for `TLSConfig`. In `prometheus/config/config.go`:
```
func (c *TLSConfig) GenerateConfig() (tls.Config error) {
...
}
```
If that was the case, I could simply call `module.TCP.TLSConfig.GenerateConfig()` in blackbox_exporter.

brian-brazil:
The idea would be to take all that code that's currently in the prometheus repo and break it out to the common repo. Similar has to be done for #23 for non-ssl auth. Reaching into the internals of other repos is unclean.

hudashot:
I've added prometheus/common#31 to common and made this PR use that code.

hudashot:
@brian-brazil, so what do you think about this?

README.md
```diff
@@ -39,6 +39,14 @@ modules:
...
     prober: tcp
     timeout: 5s
+  pop3s_banner:
+    prober: tcp
+    tcp:
+      query_response:
+      - expect: "^+OK"
+      tls: true
+      tls_config:
+        insecure_skip_verify: true
```
brian-brazil: Our examples should encourage good security, so let's leave the cert verification enabled by default

README.md
```diff
@@ -39,6 +39,14 @@ modules:
   tcp_connect:
     prober: tcp
     timeout: 5s
+  pop3s_banner:
```
brian-brazil: This should be added to blackbox.yml

knyar:
> Our examples should encourage good security, so let's leave the cert verification enabled by default

What would be the right way to document that option? Should I add two more sections into `blackbox.yml`: one with `tls: true` and the second one with `insecure_skip_verify` additionally set?

brian-brazil:
Probably easiest to have the example with `insecure_skip_verify: false`. We'll have to figure out how we're going to document these common options (alertmanager will probably end up needing them too).

brian-brazil:
Please squash the commits too when you're ready.

knyar:
Updated `blackbox.yml`, squashed the commits.

hudashot:
Updated `blackbox.yml`, squashed the commits.

brian-brazil:
Thanks!

PR #35, opened 2016-03-30
http: add "http_host" param to test back-ends and virtual hosts

tux21b:
This CL adds a http_host parameter (similarly to the target parameter) which can be used to set the "Host" header of the HTTP request.

This feature allows the blackbox_exporter to test back-ends and servers with virtual hosts.

### DIFF
--- a/http.go
+++ b/http.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"regexp"
 	"strings"
 
@@ -41,7 +42,7 @@ func matchRegularExpressions(reader io.Reader, config HTTPProbe) bool {
 	return true
 }
 
-func probeHTTP(target string, w http.ResponseWriter, module Module) (success bool) {
+func probeHTTP(target string, w http.ResponseWriter, module Module, params url.Values) (success bool) {
 	var isSSL, redirects int
 	config := module.HTTP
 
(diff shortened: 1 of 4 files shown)


brian-brazil:
Settings like these should come from configuration, not the URL. We should also allow adding of any header, not just Host.

The way to approach this would be to add a `headers` map to the http config that allows for any header to be set.

tux21b:
Hi, thanks for the reply. I agree that supporting a general `headers` map in the config would be a good thing, e.g. for setting User-Agents, Accept-Encodings, etc. since that a general config values that should be reused across nodes. I can prepare a patch for that too if there is interest.

But the `Host` header is different in my opinion. You already support passing the Probe-URL as `target`, so that a user doesn't have to write a separate config for each target. Since the Host header has (depending on the setup) about the same variety, passing it as parameter seems like a good idea.

brian-brazil:
I don't really see it that way, each vhost is a separate type of service and it's normal for each service to get it's own config block as each can have their own settings. You'd then apply that to each VIP/machine.

brian-brazil:
That;s two different checks for the same service, and would be treated the same as two different paths, two different schemes etc.

tux21b:
Sorry for all the questions, but why is target a parameter in the first place? Wouldn't it be nicer to have a standalone blackbox prober with its own config that only exports the results passively to Prometheus?

Anyway, I have created another pull request: https://github.com/prometheus/blackbox_exporter/pull/36 This one is now outdated.

brian-brazil:
> Wouldn't it be nicer to have a standalone blackbox prober with its own config that only exports the results passively to Prometheus?

Service discovery and scheduling belongs in Prometheus, we don't want that duplicated down in exporters.

PR #36, opened 2016-03-31
configureable http headers for the http probe

tux21b:
Another try. This CL makes https://github.com/prometheus/blackbox_exporter/pull/35 obsolete and adds a more general headers map as discussed before.


brian-brazil:
That looks good, can you add an example to the README?

tux21b:
Sure. PTAL :)

brian-brazil:
Great, thanks!

PR #40, opened 2016-05-30
Add notes for privileges necessary for ICMP module

lae:
This exporter does not need to be run as root to use the ICMP module, we should probably avoid implying as such.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ modules:
     timeout: 5s
 ```
 
-HTTP, HTTPS (via the `http` prober), TCP socket and ICMP (v4 only, requires privileged access) are currently supported.
+HTTP, HTTPS (via the `http` prober), TCP socket and ICMP (v4 only, see permissions section) are currently supported.
 Additional modules can be defined to meet your needs.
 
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
`cap_net_raw` in Linux specific as far as I know, I left it as a vague "privileged" to cover both root and capabilities. Do you have suggestions for a better wording that'll cover all this?

lae:
Ah, sorry, I wasn't thinking about other targets.

I think we should probably still specify `cap_net_raw`'s all that's required for Linux distributions, just to point out capabilities exist to those who aren't aware of them. Maybe just a separate note at the bottom? Or something [like this](https://gitlab.com/RoliSoft/Host-Scanner/tree/master#permissions)

brian-brazil:
Something like you linked would be good, probably down the bottom to avoid clutter.

README.md
```diff
@@ -109,6 +109,13 @@ scrape_configs:
         replacement: 127.0.0.1:9115  # Blackbox exporter.
 ```
 
+## Permissions
+
+The ICMP probe requires elevated privileges to function:
+
+* *Linux*: root user _or_ `CAP_NET_RAW` capability is required.
+  * Can be set by executing `setcap cap_net_raw+ep blackbox_exporter`
+* *BSD / OS X*: root user is required.
```
brian-brazil: I guess it's the same on Windows too, or should we just leave it unspecified?

We could just say "Other Unix", which includes Windows technically :)

lae: I had the Windows line still in earlier, but seeing as there isn't a win32 release for blackbox_exporter I wasn't sure if we wanted to imply that the exporter has been tested on Windows or not. Do you want me to include it still?

brian-brazil: I haven't tested it, but I'd presume it works. Best to include it.

lae: Alright, updated.

brian-brazil:
Thanks!

PR #44, opened 2016-06-27
Add DNS support (fix #25)

### LINKED ISSUE #25: dns module?
hi,
is a dns module something you'd be interested in having in blackbox exporter? if so I'm happy to take a stab at it using e.g. https://github.com/miekg/dns

thanks!

knyar:


dns.go
```diff
@@ -0,0 +1,117 @@
...
+
+	response, _, err := client.Exchange(msg, target)
+	if err != nil {
+		log.Warnf("Error while sending a DNS query: %s", err)
+		return false
+	}
+	log.Debugf("Got response: %#v", response)
+
+	// These metrics can be used to build additional alerting based on the number of replies.
+	fmt.Fprintf(w, "probe_dns_answer_rrs %d\n", len(response.Answer))
```
brian-brazil: These should always be returned

README.md
```diff
@@ -73,9 +73,33 @@ modules:
...
     prober: icmp
     timeout: 5s
+  dns_udp:
+    prober: dns
+    timeout: 5s
+    dns:
+      query_name: "www.prometheus.io"
+      query_type: "A"
+      valid_rcodes:
+      - 0
```
brian-brazil: Can these be the names rather than numbers? I expect most people don't know the code for NXDOMAIN.

knyar:
I've addressed your comments in two additional commits. Happy to squash when you think this is ready to be merged.

main.go
```diff
@@ -73,10 +74,26 @@ type TCPProbe struct {
 type ICMPProbe struct {
 }
 
+type DNSProbe struct {
+	Protocol           string         `yaml:"protocol"` // defaults to udp
```
brian-brazil: Comments should start with a capital letter and end with a full stop.

brian-brazil:
That looks generally good, can you squash down to the change and then the vendoring?

knyar:
Fixed comments in `main.go`, squashed to two commits.

dns.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/miekg/dns"
+	"github.com/prometheus/common/log"
+)
+
+// validRRs checks a slice of RRs received from the server against a DNSRRValidator.
+func validRRs(rrs *[]dns.RR, v *DNSRRValidator) bool {
+	// Probe should fail if a DNSRRValidator contains some rules, however there are no
+	// RRs of a given type in the response.
+	if len(*rrs) == 0 {
+		return len(v.FailIfMatchesRegexp) == 0 && len(v.FailIfNotMatchesRegexp) == 0
```
brian-brazil: Shouldn't this just be the 2nd clause? Do we need to special case this?

knyar: Sorry, I am not sure I understand what you mean. Could you please elaborate?

brian-brazil: Say you had an empty result and there's one FailIfMatchesRegexp. It'll incorrectly return false.

knyar: Returning false would be the designed behaviour in this case, as described by the comment. The assumption here is that if you specify `FailIfMatchesRegexp` or `FailIfNotMatchesRegexp`, you expect at least a single RR to be returned (to actually do the matching).

Thinking more about this, I agree that this might not be the most obvious behaviour.

I've changed this to fail only when `FailIfNotMatchesRegexp` is specified, but no RRs are returned, which I believe might be more intuitive.

brian-brazil:
Thanks!

PR #47, opened 2016-07-04
support postgres-style in-band explicit 'STARTTLS'

FauxFaux:
I'm interested in capturing the expiry time of Postgres server's SSL certificates. Postgres uses a `STARTTLS`-style connection setup: you have to speak some Postgres protocol before you can speak TLS.

This change adds support for explicitly talking that protocol, if requested, before doing the normal TLS work. This is similar to `openssl s_client -starttls smtp -connect ..` (although openssl currently doesn't support Postgres).

I have not tested:
- deadlines
- full cert chain verification (this is uncommon in the Postgres world; the official client defaults to limited validation)

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,14 @@ modules:
       protocol: "tcp" # accepts "tcp/tcp4/tcp6/udp/udp4/udp6", defaults to "udp"
       preferred_ip_protocol: "ip4" # used for "udp/tcp", defaults to "ip6"
       query_name: "www.prometheus.io"
+  pg_ssl_connect:
+    prober: tcp
+    timeout: 5s
+    tcp:
+      tls: true
+      tls_config:
+        insecure_skip_verify: true
+        starttls: postgres
 ```
 
 HTTP, HTTPS (via the `http` prober), DNS, TCP socket and ICMP (v4 only, see permissions section) are currently supported.
--- a/tcp.go
+++ b/tcp.go
@@ -16,6 +16,7 @@ package main
 import (
 	"bufio"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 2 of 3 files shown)


brian-brazil:
I'm not sure it's practical to allow for all the different ways different systems have to initiate SSL inside our codebase, as I can see that getting complicated fast. Do you have examples of how other systems do it to see if we can extract out something common that could be configured?

FauxFaux:
The only application I'm aware of which does this kind of "debugging connect to whatever" is `openssl s_client`, which just has a big switch on an enum: https://github.com/openssl/openssl/blob/master/apps/s_client.c#L1815

There are a distinctly limited number of protocols that people could want to support. I don't believe `STARTTLS` is all that popular any more, so it's unlikely that more people will be adding more protocols to the list... hopefully.

brian-brazil:
That's 8, though I'm not sure how many people would know openssl supports that (I didn't) and end up coding their own.

What would you think of using the query/response support in the tcp module for this? It is text rather than binary based.

FauxFaux:
For Postgres (the one I care about), the intital protocol is binary: `[]byte{0, 0, 0, 8, 0x04, 0xd2, 0x16, 0x2f}`.  For some other protocols, you have to have a bit of a chat, so you can't just directly send the `STARTTLS` equivalent (like you can in Postgres); which seems like it would unnecessarily complicate a DSL for communications.

It seems implausible to offer arbitrary extension (without embedding a scripting language or something); maybe it would be better to encourage people to write custom exporters, and expose `blackbox_exporter's` functionality as a library? Plus, it would mean I could get rid of that unbelievably awful relabelling config...

brian-brazil:
I'm not against expanding the query/response to be binary in principle, though I can see that getting complicated fast.

I guess we've found the line where something is custom enough to need it's own exporter. Maybe if a nice library appears in future we can take it in.

> Plus, it would mean I could get rid of that unbelievably awful relabelling config...

Any exporter doing what the blackbox exporter does will have the same style of config, see for example the SNMP exporter. This is due to the general architectural model of Prometheus. The only other option is to run an exporter per postgres.

brian-brazil:
I don't think this is something we'll be taking in.

PR #54, opened 2016-07-28
Added support for posting a body of content

drewhemm:


README.md
```diff
@@ -131,6 +142,15 @@ scrape_configs:
         regex: .*
         target_label: __address__
         replacement: 127.0.0.1:9115  # Blackbox exporter.
+
+  - job_name: 'blackbox-post'
+    metrics_path: /probe
+    params:
+      body: ['{}'] # Content, type must match the Content-Type header configured in the module
```
brian-brazil: All parameters must come from the config.

drewhemm: May I ask why? It's more flexible having it in the scrape config. We have multiple endpoints that we scrape, at the moment with the same body, but if we needed different body content, we would need to have multiple module declarations and could no longer use generic names such as http_json_post_2xx, we'd need http_json_post_2xx_service_a and http_json_post_2xx_service_b etc. 

Kind regards

Andrew

brian-brazil: This is how we approach everything for the blackbox exporter. Each config is meant to be a test of a particular thing such as a service, which you then run against multiple targets.

drewhemm: Okay, I will adjust and be back shortly :)

drewhemm:
I will squash my changes into a single commit, then open a new pull request.

Issue #41, opened 2016-05-31
Please support custom CAs for http prober


FauxFaux:
The `tcp` prober supports specifying a custom certificate authority store, but the `http` one does not.

brian-brazil:
Would you like to send a PR adding a `TLSConfig` to the http module?

lausser:
Hi,
i just played around a bit. Before sending a pull-request, what's better in your opinion?
1) new option for the http prober
modules:
  http_2xx:
    prober: http
    timeout: 5s
    http:
      valid_status_codes: [] # Defaults to 2xx
      .....
      tls_config:
        insecure_skip_verify: true

2) using the tcp prober to tell the module how to handle insecure CAs
modules:
  http_2xx:
    prober: http
    timeout: 5s
    http:
      valid_status_codes: [] # Defaults to 2xx
      ...
      no_follow_redirects: false
    tcp:
      tls: true
      tls_config:
        insecure_skip_verify: true

Gerhard

brian-brazil:
Definitely the former. Probers shouldn't depend on each other.

javefang:
Looks like this is supported already. This is not documented in this repository, but I noticed you can set "ca_file" under the "tls_config".

e.g.
```
modules:
  http_2xx:
    probe: http
    timeout: 5s
    http:
      method: GET
      tls_config:
        ca_file: <path_to_custom_ca.pem>
```

PR #57, opened 2016-08-20
http: implement TLS and InsecureSkipVerify

lausser:
As we talked yesterday in: 
[https://github.com/prometheus/blackbox_exporter/issues/41](url) "Please support custom CAs for http prober"


lausser:
Oh, sorry for that. My "go test" suceeded so i thought it's fine. Will have to fix this and then come back.

PR #58, opened 2016-08-20
http: implement TLS and InsecureSkipVerify

lausser:
Allow InsecureSkipVerify for self-signed certificates (Issue  #41)

### DIFF
--- a/http.go
+++ b/http.go
@@ -62,6 +62,15 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 		Timeout: module.Timeout,
 	}
 
+	tlsconfig, err := module.HTTP.TLSConfig.GenerateConfig()
+	if err != nil {
+		log.Errorf("Error generating TLS config: %s", err)
+		return false
+	}
+	client.Transport = &http.Transport{
+		TLSClientConfig: tlsconfig,
+	}
+
 	client.CheckRedirect = func(_ *http.Request, via []*http.Request) error {
 		redirects = len(via)
 		if redirects > 10 || config.NoFollowRedirects {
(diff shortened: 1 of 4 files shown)


http.go
```diff
@@ -61,6 +61,16 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 	client := &http.Client{
 		Timeout: module.Timeout,
 	}
+	if module.HTTP.TLS {
```
brian-brazil: There's no need for this, always generate the tls config.

http.go
```diff
@@ -61,6 +61,16 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
...
+		config, err := module.HTTP.TLSConfig.GenerateConfig()
+		if err != nil {
+			log.Errorf("Error configtls request for target %s", err)
+			return
```
brian-brazil: Explicitly return false here

http.go
```diff
@@ -61,6 +61,16 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 	client := &http.Client{
 		Timeout: module.Timeout,
 	}
+	if module.HTTP.TLS {
+		config, err := module.HTTP.TLSConfig.GenerateConfig()
+		if err != nil {
+			log.Errorf("Error configtls request for target %s", err)
```
brian-brazil: This reads oddly. "Error generating TLS config"

main.go
```diff
@@ -58,6 +58,8 @@ type HTTPProbe struct {
 	Headers                map[string]string `yaml:"headers"`
 	FailIfMatchesRegexp    []string          `yaml:"fail_if_matches_regexp"`
 	FailIfNotMatchesRegexp []string          `yaml:"fail_if_not_matches_regexp"`
+	TLS                    bool              `yaml:"tls"`
```
brian-brazil: This is no longer needed.

lausser:
I'm not sure if the additional test (http-webserver and insecure_skip_verify=false) is too much, i felt like "even if there's no benefit, at least there is no harm"

brian-brazil:
More tests are always good. Can you squash your commits?

syepes:
+1 A must have ;-)

brian-brazil:
Could you add a insecure_skip_verify: false to the http_2xx example in the readme?

brian-brazil:
Thanks!

Issue #24, opened 2015-12-21
IPv6 support


hasso:
- ICMP module is IPv4 only.
- While TCP/HTTP modules seem to support IPv6, it's unclear how the choice is made between IPv4 and IPv6. For monitoring it's often good idea to force to is it either one.

RichiH:
For all of the blackbox_exporters, I would prefer to be able to force v4, v6, or v4 _and_ v6. Obviously, at least the last will need to have a label attached to avoid duplicates.

PR #59, opened 2016-08-30
icmp: support IPv6 addresses

stapelberg:
For hosts which have both an A and an AAAA record, the default stays
IPv4, so there’s no change in behavior.

In addition to being able to specify IPv4 addresses, you can now specify
IPv6 addresses. Tested using curl
http://localhost:9115/probe?module=ping_1s&target=127.0.0.1 and
http://localhost:9115/probe?module=ping_1s&target=::1

related to #24

### DIFF
--- a/icmp.go
+++ b/icmp.go
@@ -15,13 +15,15 @@ package main
 
 import (
 	"bytes"
-	"golang.org/x/net/icmp"
-	"golang.org/x/net/ipv4"
 	"net"
 	"net/http"
 	"os"
 	"sync"
 	"time"
+
+	"golang.org/x/net/icmp"
+	"golang.org/x/net/ipv4"
+	"golang.org/x/net/ipv6"
 
 	"github.com/prometheus/common/log"
 )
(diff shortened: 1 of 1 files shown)


icmp.go
```diff
@@ -40,24 +42,38 @@ func getICMPSequence() uint16 {
...
 		return
 	}
-	defer socket.Close()
 
-	ip, err := net.ResolveIPAddr("ip4", target)
+	var (
+		socket                 *icmp.PacketConn
+		requestType, replyType icmp.Type
+	)
+	if ip.IP.To4() == nil {
```
brian-brazil: We should export whether v4 or v6 is being used

stapelberg: Do you have a suggestion as to how to export that?

brian-brazil: As a gauge with either the value 4 or 6.

hasso:
While I like the code, I disagree with logic – there is no way to probe by name and force it to IPv6. I borrowed some of your code to enhance the patch I'm using here quite some time already and which adds support for forcing probes to IPv4 or IPv6 to all modules - #64 .

stapelberg:
Alright then. Closing this one in favor of your PR #64

PR #64, opened 2016-09-14
IPv6 support for all modules

hasso:
Introduce protocol field for every module which can be used to force
probe to IPv4 (tcp4 or icmp4) or IPv6 (tcp6 or icmp6). Default for tcp
and http modules is current behaviour - let stack decide. Default for
icmp is also current behaviour - IPv4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,14 @@ modules:
       - "Download the latest version here"
       tls_config:
         insecure_skip_verify: false
+      protocol: "tcp" # accepts "tcp/tcp4/tcp6", defaults to "tcp"
+      preferred_ip_protocol: "ip4" # used for "tcp", defaults to "ip6"
   tcp_connect:
     prober: tcp
     timeout: 5s
+    tcp:
+      protocol: "tcp"
+      preferred_ip_protocol: "ip4"
   pop3s_banner:
     prober: tcp
     tcp:
@@ -75,6 +80,9 @@ modules:
   icmp:
     prober: icmp
     timeout: 5s
+    icmp:
+      protocol: "icmp"
+      preferred_ip_protocol: "ip4"
   dns_udp:
     prober: dns
     timeout: 5s
@@ -97,7 +105,8 @@ modules:
   dns_tcp:
     prober: dns
     dns:
-      protocol: "tcp"  # can also be something like "udp4" or "tcp6"
+      protocol: "tcp" # accepts "tcp/tcp4/tcp6/udp/udp4/udp6", defaults to "udp"
+      preferred_ip_protocol: "ip4" # used for "udp/tcp", defaults to "ip6"
       query_name: "www.prometheus.io"
 ```
 
--- a/dns.go
+++ b/dns.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"regexp"
 
@@ -82,6 +83,7 @@ func validRcode(rcode int, valid []string) bool {
 
 func probeDNS(target string, w http.ResponseWriter, module Module) bool {
 	var numAnswer, numAuthority, numAdditional int
+	var dialProtocol, fallbackProtocol string
 	defer func() {
 		// These metrics can be used to build additional alerting based on the number of replies.
 		// They should be returned even in case of errors.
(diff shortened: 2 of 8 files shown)


brian-brazil:
I can see users wanting to being able to prefer v4 or v6 but fallback to the other too. Could that be added?

Which was used also needs to be exported.

hasso:
Preference could be added, but I'd avoid that - that would introduce at least yet another configure option, complicating both configuration and the code and which would not be used much anyway. And fallback when? According to answers from resolver? Or when preferred one times out? Or run them in parallel? Which time values should be returned in this case?

This is also true for exporting protocol info. It could be done, but it should be done for all probes then - ie for HTTP, TCP and UDP as well. And it really makes sense only if you let the stack decide which one to use. If you use tcp[4|6] and udp[4|6], the info is already there in blackbox_exporter modules.

PS. I don't see any use of fallbacks and such automatics myself. If I'm monitoring something, I want to know exactly what. Fallbacks and address family selection automatics make perfect sense for users of the services, but not for administrators.

Summary: is it worth of effort?

brian-brazil:
> And fallback when?

Prefer v4 would mean that if a resolver returns v4 and v6, use v4 - but if it only returns v6 use v6.

> This is also true for exporting protocol info. It could be done, but it should be done for all probes then - ie for HTTP, TCP and UDP as well.

Yes, it should be done for them all.

> I want to know exactly what. Fallbacks and address family selection automatics make perfect sense for users of the services, but not for administrators.

I've seen other similar systems end up supporting this, so if it's something we're going to need I'd prefer to design for it now.

hasso:
This is a RFC. The idea is that "protocol: proto[4|6]" there is no fallback, in case of "protocol: proto", there is fallback to other protocol and preferred one can be configured by user (IPv6 is default).

DNS is not yet implemented (should it, as there is no way to specify a server anyway?), no docs and tests yet. And this is all much more complicated than it should be - there is no way to specify your address family preference in Go and there is no way to access underlaying net.Conn from higher level interfaces (http.Client and dns.Client). So, this is the way I see it could be done.

brian-brazil:
That generally seems like an okay idea.

hasso:
Here it is. Tests are for DNS and TCP only, as I have no idea how to control what exactly httptest.Server listens and icmp testing requires root privileges.

And for a future reference - the code can be simplified a lot if https://github.com/golang/go/issues/14760 will be implemented.

dns.go
```diff
@@ -90,8 +92,45 @@ func probeDNS(target string, w http.ResponseWriter, module Module) bool {
...
+
+		if ip.IP.To4() == nil {
+			dialProtocol = module.DNS.Protocol + "6"
+		} else {
+			dialProtocol = module.DNS.Protocol + "4"
+		}
+	}
+
+	if dialProtocol[len(dialProtocol)-1] == '6' {
+		fmt.Fprintf(w, "probe_protocol 6\n")
```
brian-brazil: probe_ip_protocol would be clearer

icmp.go
```diff
@@ -39,15 +41,63 @@ func getICMPSequence() uint16 {
...
+		socket           *icmp.PacketConn
+		requestType      icmp.Type
+		replyType        icmp.Type
+		fallbackProtocol string
+	)
+
 	deadline := time.Now().Add(module.Timeout)
-	socket, err := icmp.ListenPacket("ip4:icmp", "0.0.0.0")
+
+	// Defaults to IPv4 to be compatible with older versions
```
brian-brazil: I'm okay with this changing to be consistent with other modules

README.md
```diff
@@ -45,9 +45,14 @@ modules:
       - "Download the latest version here"
       tls_config:
         insecure_skip_verify: false
+      protocol: "tcp" # accepts "tcp/tcp4/tcp6", defaults to "tcp"
+      preferred_ip_protocol: "ip4" # preferred in case of "protocol: tcp", accepts "ip4/ip6", defaults to "ip6"
```
brian-brazil: This is a bit wordy, considering the rest is self-explanatory. Can it be trimmed?

dns.go
hasso: Agree, done.

icmp.go
hasso: OK, but as this one could potentially break monitoring for users, warning into release announcement, please.

README.md
hasso: OK, should be better now.

README.md
```diff
@@ -98,6 +106,7 @@ modules:
     prober: dns
     dns:
       protocol: "tcp"  # can also be something like "udp4" or "tcp6"
```
brian-brazil: Can you make this like the above comment?

dns_test.go
```diff
@@ -318,3 +319,138 @@ func TestServfailDNSResponse(t *testing.T) {
 		}
 	}
 }
+
+func TestDNSProtocol(t *testing.T) {
+	// This test assumes that listening "tcp" listens both IPv6 and IPv4 traffic and
+	// localhost resolves to both 127.0.0.1 and ::1. we must skip the test if either
+	// of these isn't true. This should be true for modern Linux systems.
+	if runtime.GOOS == "dragonfly" || runtime.GOOS == "openbsd" {
+		t.Skip("IPv6 socket isn't able to asccept IPv4 traffic in the system.")
```
brian-brazil: accept

README.md
hasso: Done.

dns_test.go
hasso: Fixed.

brian-brazil:
Thanks!

agsmith:
I'm very interested in seeing this get into a release.  Any idea when v0.3.0 would be cut?

http.go
```diff
@@ -56,8 +58,55 @@ func matchRegularExpressions(reader io.Reader, config HTTPProbe) bool {
...
+		target_url, err := url.Parse(target)
+		if err != nil {
+			return false
+		}
+		target_host, _, err := net.SplitHostPort(target_url.Host)
+		// If split fails, assuming it's a hostname without port part
+		if err != nil {
+			target_host = target_url.Host
+		}
+		ip, err := net.ResolveIPAddr(module.HTTP.PreferredIpProtocol, target_host)
```
discordianfish: I was reading the code and came across this and tracked it to this PR:
This causes another DNS roundtrip for probe, correct? Or do I miss something?

PR #78, opened 2016-11-10
Improve gzip support

carlpett:
So, this PR adds `probe_http_content_compressed` and also calculates the `content_length` of the compressed content. Additionally, it exposes a new option on the HTTP module for disabling the built-in support for gzip altogether.

### DIFF
--- a/http.go
+++ b/http.go
@@ -123,8 +123,9 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 		return net.Dial(dialProtocol, address)
 	}
 	client.Transport = &http.Transport{
-		TLSClientConfig: tlsconfig,
-		Dial:            dial,
+		TLSClientConfig:    tlsconfig,
+		Dial:               dial,
+		DisableCompression: module.HTTP.DisableGzipEncoding,
 	}
 
 	client.CheckRedirect = func(_ *http.Request, via []*http.Request) error {
@@ -154,6 +155,9 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 	}
 
 	resp, err := client.Do(request)
+	content_length := resp.ContentLength
+	content_compressed := 0
+
 	// Err won't be nil if redirects were turned off. See https://github.com/golang/go/issues/3795
 	if err != nil && resp == nil {
 		log.Warnf("Error for HTTP request to %s: %s", target, err)
@@ -170,6 +174,13 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 			success = true
 		}
 
+		// If the response was gzipped and was transparently uncompressed
+		if resp.Uncompressed {
+			content_compressed = 1
+			if body, err := ioutil.ReadAll(resp.Body); err == nil {
+				content_length = int64(len(body))
+			}
+		}
 		if success && (len(config.FailIfMatchesRegexp) > 0 || len(config.FailIfNotMatchesRegexp) > 0) {
 			success = matchRegularExpressions(resp.Body, config)
 		}
@@ -190,7 +201,8 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 		success = false
 	}
 	fmt.Fprintf(w, "probe_http_status_code %d\n", resp.StatusCode)
-	fmt.Fprintf(w, "probe_http_content_length %d\n", resp.ContentLength)
+	fmt.Fprintf(w, "probe_http_content_compressed %d\n", content_compressed)
+	fmt.Fprintf(w, "probe_http_content_length %d\n", content_length)
 	fmt.Fprintf(w, "probe_http_redirects %d\n", redirects)
 	fmt.Fprintf(w, "probe_http_ssl %d\n", isSSL)
 	return
--- a/main.go
+++ b/main.go
@@ -61,6 +61,7 @@ type HTTPProbe struct {
 	TLSConfig              config.TLSConfig  `yaml:"tls_config"`
 	Protocol               string            `yaml:"protocol"`              // Defaults to "tcp".
 	PreferredIpProtocol    string            `yaml:"preferred_ip_protocol"` // Defaults to "ip6".
+	DisableGzipEncoding    bool              `yaml:"disable_gzip_encoding"`
 }
 
 type QueryResponse struct {
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ modules:
         insecure_skip_verify: false
       protocol: "tcp" # accepts "tcp/tcp4/tcp6", defaults to "tcp"
       preferred_ip_protocol: "ip4" # used for "tcp", defaults to "ip6"
+      disable_gzip_encoding: false
   tcp_connect_v4_example:
     prober: tcp
     timeout: 5s
--- a/http_test.go
+++ b/http_test.go
@@ -14,8 +14,10 @@
 package main
 
 import (
+	"compress/gzip"
 	"fmt"
 	"github.com/prometheus/common/config"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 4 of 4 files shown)


carlpett:
Background: The Go net/http client automatically adds `Accept-Encoding: gzip` unless you specifically opt out by either setting `DisableGzipEncoding` on the transport, or set a `Accept-Encoding` header yourself. 
In the default mode, this causes the `Content-Length` header to be discarded. If setting `Accept-Encoding: gzip` yourself, it keeps the length header, but doesn't gunzip the body, so you cannot use the `fail_if_regexp*` methods.

carlpett:
I had some thoughts about adding another series for `probe_http_content_compressed_length` which I apparently did not remove entirely (the ExpectedLength of the test cases). My idea was to re-compress the body with `compress/gzip` [here](https://github.com/prometheus/blackbox_exporter/pull/78/files#diff-31dcde2645a9daeff4e0064653331de4R182), but I didn't get it to mirror the actual sizes I could see in the wild while testing, so I dropped it. Would that be an interesting addition? Otherwise I should probably just make sure that length is not reported as -1 in the test.

carlpett:
Oh. Apparently `Response.Uncompressed` is a 1.7 feature... Would a 1.7 bump be possible, or are there compatbility issues?

grobie:
The Prometheus builder [promu](https://github.com/prometheus/promu) already
uses 1.7.3 as default go version. This shouldn't be an issue.

carlpett:
@grobie Hm, it does? Looking at https://github.com/prometheus/promu/blob/master/circle.yml#L3, I'd say it uses 1.6 as well? Am I looking in the wrong place?

main.go
```diff
@@ -61,6 +61,7 @@ type HTTPProbe struct {
 	TLSConfig              config.TLSConfig  `yaml:"tls_config"`
 	Protocol               string            `yaml:"protocol"`              // Defaults to "tcp".
 	PreferredIpProtocol    string            `yaml:"preferred_ip_protocol"` // Defaults to "ip6".
+	DisableGzipEncoding    bool              `yaml:"disable_gzip_encoding"`
```
brian-brazil: This doesn't feel like a clean API, as we're leaking implementation details of the Go http api. Maybe we should just disable this for everyone?

http.go
```diff
@@ -170,6 +174,13 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
 			success = true
 		}
 
+		// If the response was gzipped and was transparently uncompressed
+		if resp.Uncompressed {
+			content_compressed = 1
+			if body, err := ioutil.ReadAll(resp.Body); err == nil {
+				content_length = int64(len(body))
```
brian-brazil: content_length is the content-length as reported by the HTTP header, it shouldn't be changed. A new metric should cover the actual length, and cases where there is no content length header.

main.go
carlpett: Agee about the naming. Initially I named it `AcceptGzipCompression`, which feels better, but it was a bit troublesome to figure out how to handle the default value (ie, distinguish between the user setting `AcceptGzipCompression: false` and not setting anything at all).
I don't think we should disable it for everyone, though. At least Akamai, and possibly others, has some weird behavior where they drop headers if the origin server sends compressed and the client requests non-compressed. Being clear about what is going on and allowing it to be toggled seems much better.

carlpett: Very much so that I wouldn't expect it to set the header unless I asked it, that was what led me to investigate this in the first place... 
My point about Akamai was that the blackbox_exporter gets hard to use for that usecase if we force-disable gzip. And while I agree with that vendor bug (although they claim it is "working as intended"...) workarounds shouldn't be in the code, I think there is a legitimate case for having gzip supported.

http.go
carlpett: Agree as well. However, the net/http client actually discards the Content-Length header when it has set accept gzip. There are basically two possibilities that I can see:
1. Manage gunzipping the body in blackbox_exporter if `Content-Encoding: gzip` is set (note that this is also thrown away by net/http if they set the accept header)
2. Try to estimate the original length by gzipping it ourselves. This is quite a hack, and won't give correct results. 

As for the new metric, `probe_http_content_decompressed_length`?

carlpett: Here's the code: https://github.com/golang/go/blob/release-branch.go1.7/src/net/http/transport.go#L1524

And from the documentation, https://golang.org/pkg/net/http/#Transport:
>         // DisableCompression, if true, prevents the Transport from
>         // requesting compression with an "Accept-Encoding: gzip"
>         // request header when the Request contains no existing
>         // Accept-Encoding value. If the Transport requests gzip on
>         // its own and gets a gzipped response, it's transparently
>         // decoded in the Response.Body. However, if the user
>         // explicitly requested gzip it is not automatically
>         // uncompressed.

brian-brazil: So the removal of the header sounds like a bug then, though I can see how it would lead to confusion for some as it's about the frame rather than the packet.

carlpett: I doubt it is a bug, it looks pretty deliberate:

```go
resp.Body = &gzipReader{body: body}
resp.Header.Del("Content-Encoding")
resp.Header.Del("Content-Length")
resp.ContentLength = -1
resp.Uncompressed = true
```

I'm guessing it is as you say, they remove them so that noone tries to double-gunzip etc.

carlpett:
So, what is the path forward here? I could change so that `transport.DisableCompression` is true by default, which would address the fact that blackbox_exporter adds headers unasked for (which would be a breaking change). 
I would however strongly prefer that it is possible to enable it with a configuration flag, both because of my specific requirements, but also since it in general will reduce scrape time and bandwidth requirements. For the cases where gzip is then enabled, we need to decide how to handle the content length. `net/http` will set it to -1. One way is to have a separate metric for calculated length, which is what I've currently implemented, but the name would then be changed. Any other alternatives?

brian-brazil:
I think we should be disabling gzip, and then have a separate metric for the actual content length - to cover cases where the header isn't set.

carlpett:
So basically `DisableCompression: true` always, and then if `resp.ContentLength` is 0 or -1 add a different metric `probe_http_content_calculated_length`?

I wouldn't object to this being the default behavior. But could you elaborate on why you think it should not be configurable?
My argument for it being overridable is that a user of blackbox_exporter cannot set a `Accept-Encoding: gzip` header manually. This is different from any other header which can be set in blackbox_exporter configuration with the `headers` option (or rather, I _can_ set the header, but then `resp.Body` is not decompressed, and the regex-matching functionality is lost).

brian-brazil:
> But could you elaborate on why you think it should not be configurable?

First we should setup things so they work as users expect, then we can consider making them configurable for more advanced use cases.

> My argument for it being overridable is that a user of blackbox_exporter cannot set a Accept-Encoding: gzip header manually. 

All headers are configurable.

> I can set the header, but then resp.Body is not decompressed, and the regex-matching functionality is lost

You can use the regex against the exact data that's returned. There's likely use cases for this.

carlpett:
Alright, I'll rework the PR to just disable the "automagic" gzipping first, then.

xkilian:
@carlpett Any chance you can get a version of the PR out the door, this issue is close to being canned and would be a great benefit (as you know). Related to #337
Cheers

carlpett:
@xkilian Had completely forgotten about this... I'll see if I can get some time to fix it.

carlpett:
Looked into this a bit, it is not quite as simple as one could hope. The `Transport` setup has been moved into prometheus/common since I made this PR. In that upstream repo, compression has actually been disabled, but the vendored copy does not allow setting the flag.
The vendoring cannot be updated, however, because it now sets another undesired flag for using keepalives. There's a PR on prom/common [here](https://github.com/prometheus/common/pull/133) which let's this be configured, but it seems to have grown a bit stale too. 

Perhaps @brian-brazil or @discordianfish could give that one a look and see if it is ready to go? Then #321 can be updated and merged, which would actually, if I read them correctly, remove the need for this PR.

avonwyss:
The upgrade of 0.14.0 to 0.15.1 has led to one of our monitored targets to fail. Upon inspection I noticed this response header from the server:

```
> GET / HTTP/1.1
> Host: reflemotion.ch
> Accept: */*

< HTTP/1.1 403 Forbidden
< Cache-Control: no-cache
< Content-Type: text/html
< X-Deny-Reason: gz
< Date: Thu, 10 Oct 2019 17:35:43 GMT
< Content-Length: 7077
```

But with GZIP it works fine:
```
> GET / HTTP/1.1
> Host: reflemotion.ch
> Accept-Encoding: gz
> Accept: */*

< HTTP/1.1 200 OK
< Cache-Control: private
< Content-Type: text/html; charset=ISO-8859-1
< Vary: Accept-Encoding
< P3P: policyref="/w3c/p3p.xml", CP="NOI DSP COR CURa OUR STP"
< X-Backend: portfolio_http_back
< Date: Thu, 10 Oct 2019 17:37:24 GMT
< Content-Length: 34512
```

**Based on this I'd ask for the compression to be configurable**, since there are in fact sites which require compression in the wild.

brian-brazil:
I'm not sure that we should be supporting sites that don't implement HTTP properly, compression is optional.

brian-brazil:
This has gone stale, and we do have a metric for the content size these days.

PR #101, opened 2017-01-11
Added support for custom SSL certificate authorities

pj42:


brian-brazil:
This should come via TLSConfig which is over in common.

pj42:
Ok yep so I see I have duplicated some effort here. So the following does work:

    http:
      tls_config:
        ca_file: "custom_root_ca.pem"

But it doesn't support multiple files? And by specifying a single ca_file you then lose all the system default public root certificate authories. Thus in this pull request I supprt multiple, so that I can query normal public trusted SSL sites as well as my custom trusted ssl site?

brian-brazil:
Have you considered having two modules?

pj42:
Great point :) Ok so that makes this pull request unnecessary. Is it worth adding ca_file to the README for blackbox exporter in the initial "showing all options" section? It would have helped me anyway.

brian-brazil:
Yeah, that should be documented. There's other options too.

PR #103, opened 2017-01-17
Adds feature to proxy metric endpoints

simonswine:
This helps to proxy local metrics endpoint (which could possibly need some custom TLS configuration to access them) to expose to the prometheus server.

Would this be a worth feature?

I can create some tests in that case


brian-brazil:
I'd suggest using a standard proxy server such as haproxy to handle this. The blackbox exporter is for probing rather than data transfer.

PR #108, opened 2017-01-27
Set default timeouts for all probes

qybl:
Set default timeouts taken from example blackbox.yml, because
people can delete/forget them, or mess up the indentation making some probes fail instantly.

Happy to chat about my solution, if you'd rather have updated documentation emphasizing the importance of timeouts one could do that as well.

In my opinion this is newcomer friendlier though.

### DIFF
--- a/main.go
+++ b/main.go
@@ -41,7 +41,7 @@ type Config struct {
 
 type Module struct {
 	Prober  string        `yaml:"prober"`
-	Timeout time.Duration `yaml:"timeout"`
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks, look at https://github.com/prometheus/prometheus/blob/master/config/config.go to see how we usually do defaults.

qybl:
Okay, will do.

brian-brazil:
Obseleted by #181

PR #109, opened 2017-02-01
Add an external url support for reverse proxy configurations

hasso:
The code is adapted from Alertmanager code.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,8 +17,12 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
+	"net/url"
 	"os"
+	"path"
+	"strings"
 	"time"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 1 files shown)


hasso:
I actually think that all components might benefit from this feature, so maybe it makes sense to standardise it somehow.

brian-brazil:
External urls are for UIs intended to be used by a browser. The blackbox exporter is not such a system.

hasso:
So, how I should use blackbox exporter behind https://hostname/blackbox_exporter reverse proxy?

brian-brazil:
I'd presume it'd be https://hostname/blackbox_exporter/probe

hasso:
That's the part I understand obviously ;). But it might need some serious rewrite stuff depending on proxy in use and leaves front page links broken. Yes, I understand that these are for demo mostly, but still.

brian-brazil:
This is standard reverse proxy usage, there shouldn't be anything too complicated.

The front page links are more for those using it for the first time, them breaking in a more complex use isn't the end of the world.

hasso:
OK. Let it be then.

PR #114, opened 2017-02-23
multi targets support and add target label

bagel:


brian-brazil:
The Prometheus way is to have a target per target, not to bunch them up into one target.

bagel:
That's true, but if I have 10 thousands targets it will need 10 thousands jobs. If it supports multi targets then it can split in little jobs.

bagel:
Thanks, that works.

PR #128, opened 2017-03-12
Add error to probe failures

SuperQ:
When a probe fails, include the error in as a label so
different errors can be classified separately.

Also, be explicit about returning `true` and `false` for all return cases.

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -81,7 +81,7 @@ func validRcode(rcode int, valid []string) bool {
 	return false
 }
 
-func probeDNS(target string, w http.ResponseWriter, module Module) bool {
+func probeDNS(target string, w http.ResponseWriter, module Module) (success bool, probe_error string) {
 	var numAnswer, numAuthority, numAdditional int
 	var dialProtocol, fallbackProtocol string
 	defer func() {
@@ -112,7 +112,7 @@ func probeDNS(target string, w http.ResponseWriter, module Module) bool {
 		if err != nil {
 			ip, err = net.ResolveIPAddr(fallbackProtocol, targetAddress)
 			if err != nil {
-				return false
+				return false, "Failed to resolve"
 			}
 		}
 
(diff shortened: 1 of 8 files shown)


brian-brazil:
This will cause all sorts of breakage (beyond being a major breaking change) and doesn't generally follow the rules for how to use labels.

This sort of information can't really be sanely represented with labels, you'd need a event logging system.

discordianfish:
So I'd like to at least differentiate between: "Probe ran successfully and discovered that the target failed" and "Probe failed and has no idea whether target could have succeeded".

I'm also not sure about having full error messages in the label.. Even though they are bounded, it just doesn't feel right.

I still don't understand the reason of exposing everything no matter what might have failed. If I would design this from scratch I (naively?) would expose probe_success as well as probe_duration_seconds only if the probe itself was successful. In general it would mean we only return them if we could resolve DNS, open/establish(*) an connection and successfully send out the request.
Now additionally to that we might want a separate failure latency. But not sure if this is needed.


*) Depends on the prober. ICMP/UDP would need to consider things working if they can send out something, with TCP it would only returns success/duration if we can actually establish a connection. With HTTP, only if we actually get any response.

brian-brazil:
>  as well as probe_duration_seconds only if the probe itself was successful.

That'd be incorrect given the name, the duration is useful even if the probe fails. For example if there's a regex mismatch or unexpected DNS result, the timing can tell you if a cache is involved.

That also wouldn't work with current staleness semantics.

discordianfish:
@brian-brazil Regex mismatch would be a case where the probe itself was successful, so it would return timings. SERVFAIL DNS response after your recursor timed out with exactly $timeout as duration_seconds still doesn't make sense to me. But as long as I can join the duration metric with something that differentiates between 'probe itself failed' vs 'probe determined target failed', I'm good.

brian-brazil:
> Regex mismatch would be a case where the probe itself was successful

Regex mismatch means the probe failed, that's the whole purpose of having the regex feature.

discordianfish:
@brian-brazil I know what it means right now, I argue for better semantics. I want to differentiate between the probe failing to do it's probing and the target not responding as the probe expects.

brian-brazil:
#161 covers adding a metric to determine if the regexes failed, as we're missing that currently.

PR #129, opened 2017-03-14
added http.http_proxy setting

tux21b:
added a setting to set an http proxy that defaults to the proxy from the environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ modules:
         insecure_skip_verify: false
       protocol: "tcp" # accepts "tcp/tcp4/tcp6", defaults to "tcp"
       preferred_ip_protocol: "ip4" # used for "tcp", defaults to "ip6"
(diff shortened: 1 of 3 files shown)


brian-brazil:
This should follow the same proxy_url naming as Prometheus, and have the same validation etc.

We also don't promise anything in terms of reading from the enviroment.

tux21b:
Sounds reasonable. Thanks for your feedback! I'll update the pull request tomorrow.

dbaumgarten:
Is anyone still working on this? This feature would be extremely useful to me right now.
I need to probe the availability of different http proxys and without this PR this seems not to be possible.
@tux21b if you don't have the time for this at the moment, I would happily continue working on this.

tux21b:
@dbaumgarten feel free to continue working on this change - I am afraid I will not have time for this until the weekend.

brian-brazil:
This was added in another PR today.

Issue #135, opened 2017-04-17
Rewrite http header with specified targets


Colstuwjx:
Hi,  is there any way to rewrite http header while the target is in specified list?

brian-brazil:
No, you can create modules with the headers you need.

Colstuwjx:
To support this, blackbox exporter may need add some modules and prometheus use `relabel` to determine the real `module` param. Thus, it also need to be reloadable.

brian-brazil:
It already supports modules via params. Please read the documentation.

PR #136, opened 2017-04-18
Implement reloadable config

Colstuwjx:
Associate with #135 , we need a reloadable mechanism.

### DIFF
--- a/main.go
+++ b/main.go
@@ -176,6 +176,7 @@ func main() {
 	reloadConfig(*configFile, cfg)
 
 	hup := make(chan os.Signal)
+	reloadCh := make(chan chan error)
 	signal.Notify(hup, syscall.SIGHUP)
 	go func() {
 		for {
@@ -183,6 +184,13 @@ func main() {
 			case <-hup:
 				if err := reloadConfig(*configFile, cfg); err != nil {
 					log.Errorf("Error reloading config: %s", err)
+				}
+			case rc := <-reloadCh:
+				if err := reloadConfig(*configFile, cfg); err != nil {
+					log.Errorf("Error reloading config: %s", err)
+					rc <- err
+				} else {
+					rc <- nil
 				}
 			}
 		}
@@ -192,6 +200,20 @@ func main() {
 	http.HandleFunc("/probe",
 		func(w http.ResponseWriter, r *http.Request) {
 			probeHandler(w, r, cfg)
+		})
+	http.HandleFunc("/reload",
+		func(w http.ResponseWriter, r *http.Request) {
+			if r.Method != "POST" {
+				w.WriteHeader(http.StatusMethodNotAllowed)
+				fmt.Fprintf(w, "This endpoint requires a POST request.\n")
+				return
+			}
+
+			rc := make(chan error)
+			reloadCh <- rc
+			if err := <-rc; err != nil {
+				http.Error(w, fmt.Sprintf("failed to reload config: %s", err), http.StatusInternalServerError)
+			}
 		})
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte(`<html>
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"gopkg.in/yaml.v2"
+	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/config"
@@ -33,6 +34,11 @@ import (
 
 type Config struct {
 	Modules map[string]Module `yaml:"modules"`
+}
+
+type SafeConfig struct {
+	sync.RWMutex
+	C *Config
 }
 
 type Module struct {
@@ -101,17 +107,22 @@ var (
 		"icmp": probeICMP,
 		"dns":  probeDNS,
 	}
-	cfg = &Config{}
+	sc = &SafeConfig{
+		C: &Config{},
+	}
 )
 
-func reloadConfig(confFile string, conf *Config) (err error) {
+func (this *SafeConfig) reloadConfig(confFile string) (err error) {
 	yamlFile, err := ioutil.ReadFile(confFile)
 	if err != nil {
 		log.Fatalf("Error reading config file: %s", err)
 		return err
 	}
 
-	if err := yaml.Unmarshal(yamlFile, &conf); err != nil {
+	this.Lock()
+	defer this.Unlock()
+
+	if err := yaml.Unmarshal(yamlFile, this.C); err != nil {
 		log.Fatalf("Error parsing config file: %s", err)
 		return err
 	}
@@ -173,7 +184,9 @@ func main() {
 	log.Infoln("Starting blackbox_exporter", version.Info())
 	log.Infoln("Build context", version.BuildContext())
 
-	reloadConfig(*configFile, cfg)
+	if err := sc.reloadConfig(*configFile); err != nil {
+		log.Errorf("Error loading config: %s", err)
+	}
 
 	hup := make(chan os.Signal)
 	reloadCh := make(chan chan error)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -92,14 +100,38 @@ type DNSRRValidator struct {
...
+	}
+	sc = &SafeConfig{
+		C: &Config{},
+	}
+)
+
+func (this *SafeConfig) reloadConfig(confFile string) (err error) {
+	yamlFile, err := ioutil.ReadFile(confFile)
+	if err != nil {
+		log.Fatalf("Error reading config file: %s", err)
```
brian-brazil: We should continue running if the reload fails

main.go
```diff
@@ -152,20 +184,49 @@ func main() {
 	log.Infoln("Starting blackbox_exporter", version.Info())
 	log.Infoln("Build context", version.BuildContext())
 
-	yamlFile, err := ioutil.ReadFile(*configFile)
-	if err != nil {
-		log.Fatalf("Error reading config file: %s", err)
+	if err := sc.reloadConfig(*configFile); err != nil {
```
brian-brazil: the config shouldn't be reloaded on every probe, just on the signal/POST

main.go
```diff
@@ -152,20 +184,49 @@ func main() {
...
+		}
+	}()
 
 	http.Handle("/metrics", prometheus.Handler())
 	http.HandleFunc("/probe",
 		func(w http.ResponseWriter, r *http.Request) {
-			probeHandler(w, r, &config)
+			probeHandler(w, r, sc.C)
+		})
+	http.HandleFunc("/reload",
```
brian-brazil: This should be `/-/reload` per Prometheus

main.go
Colstuwjx: Yeah, it will be loaded at the first time, and then, just fire **ONLY** by signal/POST.

main.go
```diff
@@ -92,14 +100,41 @@ type DNSRRValidator struct {
...
+
+	this.Lock()
+	defer this.Unlock()
+
+	if err := yaml.Unmarshal(yamlFile, &c); err != nil {
+		log.Errorf("Error parsing config file: %s", err)
+		return err
+	}
+
+	*this.C = c
```
brian-brazil: `this.C = &c` would be faster, as you're copying a pointer rather than the entire struct.

main.go
```diff
@@ -92,14 +100,41 @@ type DNSRRValidator struct {
...
+	var c = Config{}
+
+	yamlFile, err := ioutil.ReadFile(confFile)
+	if err != nil {
+		log.Fatalf("Error reading config file: %s", err)
+		return err
+	}
+
+	this.Lock()
+	defer this.Unlock()
```
brian-brazil: You can take the lock just before the assignment

main.go
```diff
@@ -92,14 +100,41 @@ type DNSRRValidator struct {
...
+
+	if err := yaml.Unmarshal(yamlFile, &c); err != nil {
+		log.Errorf("Error parsing config file: %s", err)
+		return err
+	}
+
+	*this.C = c
+	log.Infoln("Loading config file")
```
brian-brazil: It is loaded by now

main.go
```diff
@@ -92,14 +100,41 @@ type DNSRRValidator struct {
...
+		"tcp":  probeTCP,
+		"icmp": probeICMP,
+		"dns":  probeDNS,
+	}
+	sc = &SafeConfig{
+		C: &Config{},
+	}
+)
+
+func (this *SafeConfig) reloadConfig(confFile string) (err error) {
```
brian-brazil: `sc` would be the usual instance name here

main.go
```diff
@@ -152,20 +187,49 @@ func main() {
...
+				}
+			}
+		}
+	}()
 
 	http.Handle("/metrics", prometheus.Handler())
 	http.HandleFunc("/probe",
 		func(w http.ResponseWriter, r *http.Request) {
-			probeHandler(w, r, &config)
+			probeHandler(w, r, sc.C)
```
brian-brazil: You need a mutex around the access to sc.C here

Colstuwjx: Thanks a lot for your nice reply :)

main.go
```diff
@@ -215,7 +215,9 @@ func main() {
 	http.Handle("/metrics", prometheus.Handler())
 	http.HandleFunc("/probe",
 		func(w http.ResponseWriter, r *http.Request) {
+			sc.RLock()
 			probeHandler(w, r, sc.C)
+			sc.RUnlock()
```
brian-brazil: This will hold the lock for the entire duration of a probe, which could be a while. Take the lock only to copy the pointer to a local variable.

main.go
```diff
@@ -215,9 +215,13 @@ func main() {
 	http.Handle("/metrics", prometheus.Handler())
 	http.HandleFunc("/probe",
 		func(w http.ResponseWriter, r *http.Request) {
+			var c Config
+
 			sc.RLock()
-			probeHandler(w, r, sc.C)
+			c = *sc.C
```
brian-brazil: You can do `c := sc.C`

main.go
```diff
@@ -215,9 +215,13 @@ func main() {
 	http.Handle("/metrics", prometheus.Handler())
 	http.HandleFunc("/probe",
 		func(w http.ResponseWriter, r *http.Request) {
+			var c Config
+
 			sc.RLock()
```
brian-brazil: Don't forget to unlock.

Colstuwjx: Yes, I have already done the RUnlock.

main.go
```diff
@@ -152,20 +187,53 @@ func main() {
 	log.Infoln("Starting blackbox_exporter", version.Info())
 	log.Infoln("Build context", version.BuildContext())
 
-	yamlFile, err := ioutil.ReadFile(*configFile)
-	if err != nil {
-		log.Fatalf("Error reading config file: %s", err)
+	if err := sc.reloadConfig(*configFile); err != nil {
+		log.Errorf("Error loading config: %s", err)
```
brian-brazil: We should fatal if there's a bad config at startup

main.go
```diff
@@ -92,14 +100,41 @@ type DNSRRValidator struct {
...
+	sc.Lock()
+	sc.C = &c
+	sc.Unlock()
+
+	log.Infoln("Loaded config file")
+	return nil
 }
 
-func probeHandler(w http.ResponseWriter, r *http.Request, config *Config) {
+func probeHandler(w http.ResponseWriter, r *http.Request, conf Config) {
```
brian-brazil: This should remain a *Config, so that we're not copying the entire config every time.

The less garbage we create the better.

main.go
```diff
@@ -92,14 +100,41 @@ type DNSRRValidator struct {
...
-	"icmp": probeICMP,
-	"dns":  probeDNS,
+var (
+	Probers = map[string]func(string, http.ResponseWriter, Module) bool{
+		"http": probeHTTP,
+		"tcp":  probeTCP,
+		"icmp": probeICMP,
+		"dns":  probeDNS,
+	}
+	sc = &SafeConfig{
```
brian-brazil: This can exist in main, and start off uninitilised.

main.go
Colstuwjx: @brian-brazil  Thanks a lot for your code comment, and I prefer taking a new branch to do real merge after all is ok :)

main.go
brian-brazil: You should still change this to `c := sc.C`. What you're currently doing is creating a fully copy of the config.

Colstuwjx:
I am just confused about whether the pointer `c` read would be unsafe. Because `c` will read the entire configs in the probe function and has no lock here.

brian-brazil:
It's safe as there's no modification of the pointed to config once it's created.

brian-brazil:
On line 111 you create a new config object that's read into, which doesn't affect the previous config object.

Colstuwjx:
Yes, you're right, as I always make a new config object while do write, everything is clear and no unsafe lock here.

brian-brazil:
Thanks!

PR #142, opened 2017-05-08
probe url adjustment according to example config

dwdraju:


brian-brazil:
The URL is correct per the included config.

Issue #76, opened 2016-11-07
Use custom registry


brian-brazil:
Now that the go client supports them, we should switch to custom registries rather than hand-creating the output.

PR #148, opened 2017-05-15
Switch from handling output by hand to using a custom registry

Conorbro:
Completes #76


dns.go
```diff
@@ -81,15 +81,36 @@ func validRcode(rcode int, valid []string) bool {
...
+	})
+	probeDNSAuthorityRRSGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_dns_authority_rrs",
+		Help: "Returns number of entries in the authority resource record list",
+	})
+	probeDNSAdditionalRRSGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_dns_additional_rrs",
+		Help: "Returns number of entries in the additional resource record list",
+	})
+	registry.Register(probeIPProtocolGauge)
```
brian-brazil: This should use MustRegister, so we'll catch future bugs.

dns_test.go
```diff
@@ -249,14 +259,30 @@ func TestAuthoritativeDNSResponse(t *testing.T) {
...
+			}
+			var buf bytes.Buffer
+			for _, mf := range mfs {
+				if _, err := expfmt.MetricFamilyToText(&buf, mf); err != nil {
+					t.Fatal(err)
+				}
+			}
+
+			for _, re := range []*regexp.Regexp{
+				regexp.MustCompile("\nprobe_dns_answer_rrs 1"),
```
brian-brazil: Why the leading \n?

http.go
```diff
@@ -56,10 +56,54 @@ func matchRegularExpressions(reader io.Reader, config HTTPProbe) bool {
...
+		})
+
+		redirectsGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_redirects",
+			Help: "The number of redirects",
+		})
+
+		isSSLGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_ssl",
+			Help: "Checks if SSL cert is valid",
```
brian-brazil: This is if SSL was used

http.go
```diff
@@ -56,10 +56,54 @@ func matchRegularExpressions(reader io.Reader, config HTTPProbe) bool {
...
+		})
+
+		probeIPProtocolGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_ip_protocol",
+			Help: "Specifies whether probe ip protocl is IP4 or IP6",
+		})
+
+		probeSSLEarliestCertExpiryGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_ssl_earliest_cert_expiry",
+			Help: "Returns earliest SSL cert expiry date",
```
brian-brazil: This is a unixtime, not a date

http.go
```diff
@@ -56,10 +56,54 @@ func matchRegularExpressions(reader io.Reader, config HTTPProbe) bool {
...
+			Name: "probe_http_ssl",
+			Help: "Checks if SSL cert is valid",
+		})
+
+		statusCodeGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "status_code",
+			Help: "Response HTTP status code",
+		})
+
+		sslExpiry = prometheus.NewGauge(prometheus.GaugeOpts{
```
brian-brazil: This is unused

http.go
```diff
@@ -56,10 +56,54 @@ func matchRegularExpressions(reader io.Reader, config HTTPProbe) bool {
...
+			Help: "The number of redirects",
+		})
+
+		isSSLGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_ssl",
+			Help: "Checks if SSL cert is valid",
+		})
+
+		statusCodeGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "status_code",
```
brian-brazil: probe_http_

http.go
```diff
@@ -188,17 +233,18 @@ func probeHTTP(target string, w http.ResponseWriter, module Module) (success boo
...
 	}
-	fmt.Fprintf(w, "probe_http_status_code %d\n", resp.StatusCode)
-	fmt.Fprintf(w, "probe_http_content_length %d\n", resp.ContentLength)
-	fmt.Fprintf(w, "probe_http_redirects %d\n", redirects)
-	fmt.Fprintf(w, "probe_http_ssl %d\n", isSSL)
+
+	statusCodeGauge.Set(float64(resp.StatusCode))
+	contentLengthGauge.Set(float64(resp.ContentLength))
+	redirectsGauge.Set(float64(redirects))
+	isSSLGauge.Set(float64(isSSL))
```
brian-brazil: This can go above now, eliminating the isSSL variable.

main.go
```diff
@@ -61,8 +63,8 @@ type HTTPProbe struct {
 	FailIfMatchesRegexp    []string          `yaml:"fail_if_matches_regexp"`
 	FailIfNotMatchesRegexp []string          `yaml:"fail_if_not_matches_regexp"`
 	TLSConfig              config.TLSConfig  `yaml:"tls_config"`
-	Protocol               string            `yaml:"protocol"`              // Defaults to "tcp".
-	PreferredIPProtocol    string            `yaml:"preferred_ip_protocol"` // Defaults to "ip6".
+	Protocol               string            `yaml:"protocol"`              // Defaults to TCP.
```
brian-brazil: This defaults to "tcp", not TCP

main.go
```diff
@@ -153,20 +163,26 @@ func probeHandler(w http.ResponseWriter, r *http.Request, conf *Config) {
...
+	registry.MustRegister(probeSuccessGauge)
+	registry.MustRegister(probeDurationGauge)
+	success := prober(target, w, module, registry)
+	probeDurationGauge.Set(time.Since(start).Seconds())
 	if success {
-		fmt.Fprintln(w, "probe_success 1")
+		probeSuccessGauge.Set(1)
 	} else {
-		fmt.Fprintln(w, "probe_success 0")
+		probeSuccessGauge.Set(0)
```
brian-brazil: It's initialised to 0, so you don't need the else

http.go
```diff
@@ -56,10 +56,48 @@ func matchRegularExpressions(reader io.Reader, config HTTPProbe) bool {
...
+		})
+
+		redirectsGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_redirects",
+			Help: "The number of redirects",
+		})
+
+		isSSLGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_ssl",
+			Help: "Checks if SSL was used",
```
brian-brazil: Indicates, not checks. It's also for the final redirect.

grobie:
Have you measured the impact of the overhead of creating all metrics on the fly during each probe? Why isn't it possible to just use const metrics here?

PR #149, opened 2017-05-15
Switch from handling output by hand to using a custom registry

Conorbro:
Completes #76

(Screwed up rebasing in the last PR #148, apologies!)


tcp.go
```diff
@@ -75,9 +76,19 @@ func dialTCP(target string, w http.ResponseWriter, module Module) (net.Conn, err
 	return tls.DialWithDialer(dialer, dialProtocol, target, config)
 }
 
-func probeTCP(target string, w http.ResponseWriter, module Module) bool {
+func probeTCP(target string, w http.ResponseWriter, module Module, registry *prometheus.Registry) bool {
+	probeIPProtocolGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_ip_protocol",
+		Help: "Specifies whether probe ip protocl is IP4 or IP6",
```
brian-brazil: protocol

grobie:
Have you measured the impact of the overhead of creating all metrics on the fly during each probe? Why isn't it possible to just use const metrics here?

Conorbro:
@grobie @brian-brazil 

`process_cpu_seconds_total`

**When handling output by hand:**

10,000 tcp probes: 2.44s
1,000 tcp probes: 0.24s
100 tcp probes: 0.02s

10,000 http probes: 3.46s
1,000 http probes: 0.35s 
100 http probes: 0.03s

**When handling output via custom registry:**

100 tcp probes: 0.08s
1,000 tcp probes: 0.82s
10,000 tcp probes: 8.06s

100 http probes: 0.09s
1,000 http probes: 1.01s
10,000 http probes: 9.6s

grobie:
Oh, so the impact is quite noticeable. The overhead should be a lot smaller when we implementing the modules as Collector and using const metrics.

brian-brazil:
I'd presume that is largely the overhead of the registry generally. Printf is going to be faster than that logic.

Conorbro:
**tcp - with registering and setting of metrics**
10000 - 7.9s/7.72s/7.98s/7.8s
1000 - 0.76s/0.84s/0.79s/0.84s
100 - 0.07s/0.07s/0.07s/0.08s

**tcp - no registering or setting of metrics**
10000 - 7.38s/7.47/7.27s/7.58
1000 - 0.74s/0.73s/0.74s/0.75
100 - 0.09/0.08/0.07/0.08s

brian-brazil:
That's a ~5% slowdown from having metrics at all, which seems fine to me from a standpoint of keeping the code clean.

grobie:
Agree, thanks for the numbers.

Issue #53, opened 2016-07-21
Question regarding probing remote instances having basic auth


svenmueller:
Hi,

Some of our instances (e.g. staging servers) are protected using BASIC AUTH. Is there a way using blackbox-exporter to probe those servers (e.g. using module `http_2xx`)

Thx in Advance,
Sven

brian-brazil:
You can pass in headers with the http module, which would let you do basic auth.

brian-brazil:
Yes, you'd need a module config per different set of credentials.

I'd suggest looking into having a monitoring set of credentials that works across targets.

svenmueller:
Yes, that would be one way of limiting the amount of required modules.

Btw, thx for your awesome work here (and all other Prometheus related project). I really appreciate it!

Cheers Sven

PR #150, opened 2017-05-16
Add a Basic Auth example to the README.

SuperQ:


SuperQ:
But [here ](https://github.com/prometheus/blackbox_exporter/issues/53#issuecomment-234250146) you said to use headers

SuperQ:
Otherwise, we should simply implement it like https://github.com/prometheus/blackbox_exporter/pull/23

brian-brazil:
Ah, I was looking for #23 and couldn't find it.

SuperQ:
A variation on #23 would work just fine for me.  Let me know what you're thinking, this doesn't seem like a complicated code change.

brian-brazil:
It's more factoring out the relevant bit of code into common, and then reusing it here. We'll be wanting the same in the AM too.

SuperQ:
For now, can we document a workaround?

brian-brazil:
Considering how we currently treat secrets in configs, I don't think we should be encouraging users to do things that'd bypass those protections (not that the blackbox exporter exposes its config yet).

SuperQ:
Secrets in configs seems fine to me, it's basically the same as community strings in the snmp exporter.

brian-brazil:
Secrets in places where you don't expect them to be is a problem though. Random headers are not where you expect secrets.

SuperQ:
I wouldn't call `Authorization` obscure or random.

discordianfish:
I think it's perfectly fine to store secrets in a config (are we really discussing this?). I really don't get what the problem is. People want to use basic auth, we support basic auth (by setting headers) so we should tell people how to do it..?
Maybe add a sentence like "Basic auth is not officially supported but this is how you can do it anyway"?

brian-brazil:
> I think it's perfectly fine to store secrets in a config (are we really discussing this?).

It's fine if it's in a place we expect to have secrets. HTTP headers are not such a place.

SuperQ:
There's nothing surprising about using headers for basic auth.

It says it in in the docs for the [golang helper](https://golang.org/pkg/net/http/#Request.SetBasicAuth).  Which is just a wrapper for [request.Set.Header()](https://golang.org/src/net/http/request.go?s=28267:28324#L860).

discordianfish:
@brian-brazil The 'Authorization' header is there since forever and it was always meant to include credentials. Is there any form of proof or argument that would change your mind?

brian-brazil:
There's a difference between implementation details and user-facing APIs. In Prometheus we obtain this information via the `basic_auth` fields, and the blackbox exporter should follow that.

brian-brazil:
Basic auth (and the other auth options) are now supported in the usual Prometheus, though not released yet. There's also an example in the readme (at least for a little bit, that's about to get rejigged).

SuperQ:
Nice! Thanks!

PR #151, opened 2017-05-16
Use url.params as parameter for probes

guilhem:
This let probes adding their own configuration

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"regexp"
 
 	"github.com/miekg/dns"
@@ -81,7 +82,7 @@ func validRcode(rcode int, valid []string) bool {
 	return false
 }
 
-func probeDNS(target string, w http.ResponseWriter, module Module) bool {
+func probeDNS(params url.Values, w http.ResponseWriter, module Module) bool {
 	var numAnswer, numAuthority, numAdditional int
 	var dialProtocol, fallbackProtocol string
 	defer func() {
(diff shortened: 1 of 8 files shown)


brian-brazil:
Why do you want to make this change? Probes are configured entirely from the target and the config.

guilhem:
@brian-brazil we plan to add another probes like s3 / cassandra.
This probes need more options than only `target`.

guilhem:
That may be a bad way to do it. We are really open to any suggestion :)
That's why this PR is open

SuperQ:
Cassandra should be monitored with the [jmx_exporter](https://github.com/prometheus/jmx_exporter/blob/master/example_configs/cassandra.yml).

I'm not sure what S3 monitoring you're looking for, but it's probably better done without blackbox probing.

guilhem:
indeed cassandra is monitored with jmx_exporter. But it's clearly not a "blackbox" test.
We are very sensitive to latency and we would like to have an external monitoring of all our services

SuperQ:
What about application-side metrics for Cassandra requests?  You'll catch a lot more latency issues with an application-side histogram of Cassandra requests.

guilhem:
That's perfectly true, but you can't trust every application metrics.
As a service provider, we have to create our own metrics to expose a true SLA.

SuperQ:
Yes, good point.  Expressing different probes in the configuration file is still preferred for the blackbox exporter.  Although it's not really designed for something as complex as Cassandra filesystem operations.

This may be the case where a specialized prober is required.

brian-brazil:
I think that supporting anything beyond very simple and standard protocols would become unmaintainable. While you should be able to some basic S3 monitoring via HTTP, something like C* is too application specific.

As SuperQ says, a separate exporter is the way to go here.

guilhem:
@BarthV is already working on a S3 probe https://github.com/BarthV/blackbox_exporter/tree/s3v2

We don't know if cassandra patches can be upstream here, but this patch will ease maintenance of custom probes (like an "addon").

brian-brazil:
I don't think any probers for specific applications or services will be accepted, you're free to maintain your own versions though.

guilhem:
@brian-brazil I really think it's a good thing :)

This patch just creates a "simpler" interface when an application needs more parameters than only "target".

We may this in the future probes as a lib that can import in our own "blackbox_exporter"

brian-brazil:
> This patch just creates a "simpler" interface when an application needs more parameters than only "target".

It is an explicit design decision of the blackbox exporter not to allow that, the config file has all the config.

PR #154, opened 2017-05-17
Added utility methods for checking probe results

Conorbro:


utils_test.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// Check if expected results are in the input, return error if not
+func probeLookup(expRes map[string]float64, mfs []*dto.MetricFamily) error {
+	var err error
+	for i := range mfs {
+		if _, ok := expRes[mfs[i].GetName()]; ok {
+			if expRes[mfs[i].GetName()] == mfs[i].Metric[0].GetGauge().GetValue() {
+			} else {
+				err = fmt.Errorf("Expected %v %v, got %v %v", mfs[i].GetName(), expRes[mfs[i].GetName()], mfs[i].GetName(), mfs[i].Metric[0].GetGauge().GetValue())
```
brian-brazil: This should take the `t` object and fail the test directly

utils_test.go
```diff
@@ -0,0 +1,29 @@
...
+			}
+		}
+	}
+	return err
+}
+
+// Error checking
```
brian-brazil: I'd leave this as-is.

utils_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+// Check if expected results are in the input, return error if not
+func probeLookup(expRes map[string]float64, mfs []*dto.MetricFamily) error {
```
brian-brazil: This function doesn't do anything with probes, checkRegistryResults or similar would be better.

utils_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"fmt"
+	"testing"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+// Check if expected results are in the registry
+func checkRegistryResults(expRes map[string]float64, mfs []*dto.MetricFamily, t *testing.T) {
+	for i := range mfs {
+		if _, ok := expRes[mfs[i].GetName()]; ok {
```
brian-brazil: What if a time series isn't returned?

main.go
```diff
@@ -52,8 +52,14 @@ type Module struct {
 	DNS     DNSProbe      `yaml:"dns"`
 }
 
+type Probe struct {
+	Protocol            string `yaml:"protocol"`
+	PreferredIPProtocol string `yaml:"preferred_ip_protocol"`
+}
+
 type HTTPProbe struct {
 	// Defaults to 2xx.
```
brian-brazil: This comment is for the next line

main.go
```diff
@@ -52,8 +52,14 @@ type Module struct {
 	DNS     DNSProbe      `yaml:"dns"`
 }
 
+type Probe struct {
+	Protocol            string `yaml:"protocol"`
```
brian-brazil: This would be clearer as l4_protocol or transport_protocol, however the TCP and HTTP modules only support TCP - so this should be a setting only in the DNS module.

Conorbro:
@brian-brazil

PR #157, opened 2017-05-23
Fillout the dns default port `53` if there is no setup.

### LINKED ISSUE #155: Fillout a default port for dns probe
We should make a default port number, e.g. 53, for dns probe target. It could fix the following issue:

```
WARN[0012] Error while sending a DNS query: dial udp6: missing port in address internal.dns-server.co  source=dns.go:152
```

Thanks.

Colstuwjx:
Fixed #155 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,12 @@ modules:
       protocol: "tcp" # accepts "tcp/tcp4/tcp6/udp/udp4/udp6", defaults to "udp"
       preferred_ip_protocol: "ip4" # used for "udp/tcp", defaults to "ip6"
       query_name: "www.prometheus.io"
+  dns_default_target_port_example:
+    prober: dns
+    dns:
+      default_target_port: "53" # set default dns server port while `target` doesn't setup it
+      preferred_ip_protocol: "ip4" # used for "udp/tcp", defaults to "ip6"
+      query_name: "www.prometheus.io"
 ```
 
 HTTP, HTTPS (via the `http` prober), DNS, TCP socket and ICMP (v4 only, see permissions section) are currently supported.
--- a/dns.go
+++ b/dns.go
@@ -134,7 +134,7 @@ func probeDNS(target string, w http.ResponseWriter, module Module, registry *pro
 	if targetPort == "" {
 		// missingPort could be fixed by fallbacking to the default port.
 		targetAddress = target
-		targetPort = "53"
+		targetPort = module.DNS.DefaultTargetPort
 		log.Debugf("Fixing missingPort error, targetAddress %v, targetPort %v", targetAddress, targetPort)
 		target = net.JoinHostPort(targetAddress, targetPort)
 	}
--- a/dns.go
+++ b/dns.go
@@ -82,8 +82,12 @@ func validRcode(rcode int, valid []string) bool {
 }
 
 func probeDNS(target string, w http.ResponseWriter, module Module, registry *prometheus.Registry) bool {
-	var numAnswer, numAuthority, numAdditional int
-	var dialProtocol, fallbackProtocol string
+	var (
+		numAnswer, numAuthority, numAdditional int
+		dialProtocol, fallbackProtocol         string
+		targetAddress, targetPort              string
+	)
+
 	probeIPProtocolGauge := prometheus.NewGauge(prometheus.GaugeOpts{
 		Name: "probe_ip_protocol",
 		Help: "Specifies whether probe ip protocl is IP4 or IP6",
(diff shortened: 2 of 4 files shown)


dns.go
```diff
@@ -126,9 +130,17 @@ func probeDNS(target string, w http.ResponseWriter, module Module, registry *pro
 		fallbackProtocol = "ip6"
 	}
 
+	targetAddress, targetPort, _ = net.SplitHostPort(target)
+	if targetPort == "" {
+		// missingPort could be fixed by fallbacking to the default port.
```
Colstuwjx: I'm not sure about this line, as [SplitHostPort](https://golang.org/src/net/ipsock.go?s=3572:3638#L104) func has `missingPort ` and `tooManyColons ` error, will the wrong ipv6 address trouble our probe?

I'm afraid about whether there would be potential panic here...

SuperQ: I think it should function the same way, as we previously didn't check the error return at all, and would simply send the empty string target returned by `net.SplitHostPort()` to the DNS lookup.  At this point we would catch the error.

Colstuwjx: Yeah, you're right. Let me know if there is any problem either.

SuperQ:
Please update the `dns_test.go` with the new use case(s).

Colstuwjx:
Woops! I plan to always start the dns server at 53 port, but it seems not okay...

Colstuwjx:
I'd like to setup a `fail` test case to confirm the process.

dns_test.go
```diff
@@ -374,17 +374,47 @@ func TestDNSProtocol(t *testing.T) {
 
 		_, port, _ := net.SplitHostPort(addr.String())
 
-		// Force IPv4
+		// Default DNS Server port
```
Colstuwjx: Aha.. Fixed! Maybe a separate Test func could be better? Please feel free to let me know.

dns_test.go
```diff
@@ -374,17 +374,47 @@ func TestDNSProtocol(t *testing.T) {
...
 		}
 		recorder := httptest.NewRecorder()
 		registry := prometheus.NewRegistry()
-		result := probeDNS(net.JoinHostPort("localhost", port), recorder, module, registry)
+		result := probeDNS("localhost", recorder, module, registry)
+		if result {
+			t.Fatalf("DNS protocol: \"%v\" with localhost:53 succeeded, expected failure.", protocol)
+		}
+
+		// Set DNS Server target with port
```
brian-brazil: Comments should end with a full stop.

dns_test.go
```diff
@@ -374,17 +374,47 @@ func TestDNSProtocol(t *testing.T) {
...
-				Protocol:  protocol + "4",
+				Protocol:  protocol,
 			},
 		}
 		recorder := httptest.NewRecorder()
 		registry := prometheus.NewRegistry()
-		result := probeDNS(net.JoinHostPort("localhost", port), recorder, module, registry)
+		result := probeDNS("localhost", recorder, module, registry)
+		if result {
+			t.Fatalf("DNS protocol: \"%v\" with localhost:53 succeeded, expected failure.", protocol)
```
brian-brazil: What if the test machine has a local resolver?

Colstuwjx: I'm sorry, I don't understand that, as the [test assume](https://github.com/Colstuwjx/blackbox_exporter/blob/5fc97cca1dae3824b1a598b122d11a7bdd38d891/dns_test.go#L361) said that `localhost` resolves both `127.0.0.1` & `::1`

brian-brazil: This test is not hitting your test server, it's hitting localhost:53 - which may or may not work. It'd work on most Linux laptops I'd expect for example.

Colstuwjx: Yes, the test server almost would not start up at `localhost:53`, due to [test server](https://github.com/Colstuwjx/blackbox_exporter/blob/5fc97cca1dae3824b1a598b122d11a7bdd38d891/dns_test.go#L37) listen to a random port `:0`. So I designed that case to a `failure`.

Is there another pretty way to test it? I've tried `startDNSServer` with 53 port, but it didn't work...
Maybe [add default port support](https://github.com/miekg/dns/issues/491) with the upstream go dns would be better, but I'm not sure whether upstream will accept this dirty judgement, as the upstream repo is using `addr` at all place.

brian-brazil: > if there is a local resolver localhost:53 on the test machine, but that's almost not.

A local resolver is very common, the machine I'm working on has one for example.

> Is there another better way to test it?

I can't think of one.

Colstuwjx: Aha.. I introduced a new way to set default port via `DefaultTargetPort` argument in config.

README.md
```diff
@@ -101,6 +101,12 @@ modules:
       protocol: "tcp" # accepts "tcp/tcp4/tcp6/udp/udp4/udp6", defaults to "udp"
       preferred_ip_protocol: "ip4" # used for "udp/tcp", defaults to "ip6"
       query_name: "www.prometheus.io"
+  dns_default_target_port_example:
+    prober: dns
+    dns:
+      default_target_port: "53" # set default dns server port while `target` doesn't setup it
```
brian-brazil: I don't think this is something that should be configurable. If the user wants a non-standard port they can configure it with relabelling on the Prometheus end.

Colstuwjx:
Okay, it seems that adding default port is not really good via this way, so I'd like to close this PR, and prefer relabeling it.

brian-brazil:
This has been added in #169

Colstuwjx:
Cool!

Issue #159, opened 2017-05-30
probe_success fails with https only site


kaihendry:
I have a test site, letterly.com that only responds on https://letterly.com

I tried updating the regex from :80 to :443. Same issue.

```
  - job_name: 'bb'
    scrape_interval: 5m
    scrape_timeout: 10s
    metrics_path: /probe
    params:
      module: [http_2xx]  # Look for a HTTP 200 response.
    file_sd_configs:
      - files:
        - /etc/prometheus/targets.yml
    relabel_configs:
      - source_labels: [__address__]
        regex: (.*)(:443)?
        target_label: __param_target
        replacement: ${1}
      - source_labels: [__param_target]
        regex: (.*)
        target_label: instance
        replacement: ${1}
      - source_labels: []
        regex: .*
        target_label: __address__
        replacement: blackboxprober:9115  # Blackbox exporter.
```

So I'm incorrectly getting probe_success 0 https://prom.dabase.com/graph?g0.range_input=2h&g0.expr=probe_success&g0.tab=0

SuperQ:
Without the contents of the targets.yml, it's not possible to debug the problem.

kaihendry:
Here you go: https://github.com/kaihendry/pingprom/blob/master/prometheus/targets.yml

SuperQ:
In order to use https with bare target hostnames, you need to specify the protocol as `https`.  The default assumes `http`.

Either prefix the targets like `https://natalian.org`, or adjust the replacement in the target label like this:

```
    relabel_configs:
      - source_labels: [__address__]
        regex: (.*)(:443)?
        target_label: __param_target
        replacement: "https://${1}"
```

kaihendry:
Thank you, your suggestion appears to work. https://prom.dabase.com/graph?g0.range_input=1h&g0.expr=probe_success&g0.tab=1

Would be great if this could be documented? https is like the norm nowadays... :D

SuperQ:
Yes, this should be better documented.  I've filed a PR to simplify the http example.

PR #160, opened 2017-05-30
Improve example prometheus.yml config

SuperQ:
To reduce confusion about how to use http probes, simplify the `http_2xx` probe example.
* Make the example use full URLs for targets.
* Remove relabel config params that are defaults.

Closes: https://github.com/prometheus/blackbox_exporter/issues/159

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,19 +121,15 @@ scrape_configs:
       module: [http_2xx]  # Look for a HTTP 200 response.
     static_configs:
       - targets:
-        - prometheus.io   # Target to probe
+        - http://prometheus.io    # Target to probe with http.
+        - https://prometheus.io   # Target to probe with https.
+        - http://example.com:8080 # Target to probe with http on port 8080.
     relabel_configs:
       - source_labels: [__address__]
-        regex: (.*)(:80)?
         target_label: __param_target
-        replacement: ${1}
       - source_labels: [__param_target]
-        regex: (.*)
         target_label: instance
-        replacement: ${1}
-      - source_labels: []
-        regex: .*
-        target_label: __address__
+      - target_label: __address__
         replacement: 127.0.0.1:9115  # Blackbox exporter.
 ```
 


README.md
```diff
@@ -121,19 +121,15 @@ scrape_configs:
       module: [http_2xx]  # Look for a HTTP 200 response.
     static_configs:
       - targets:
-        - prometheus.io   # Target to probe
+        - http://prometheus.io    # Target to probe with http.
+        - https://prometheus.io   # Target to probe with https.
+        - http://example.com:8080 # Target to probe with http on port 8080.
     relabel_configs:
       - source_labels: [__address__]
-        regex: (.*)(:80)?
```
brian-brazil: I think this is still needed

kaihendry:
Tried incorporating your change to: https://prom.dabase.com/config

I was getting some `https://https://` duplication initially, but now it seems OK: https://prom.dabase.com/graph?g0.range_input=1h&g0.expr=probe_success&g0.tab=1

README.md
SuperQ: This is only to optionally strip `:80` from the target.. for example `http://example.com:80` would be come `http://example.com`.  This is unnecessary for an example config with the full URLs as targets.

brian-brazil: The config will also be used with DNS and ICMP, which requires stripping the :80 that Prometheus defaults in.

SuperQ: This doesn't seem to be a case at the relabel_configs phase:
```
  - job_name: 'blackbox-icmp'
    metrics_path: /probe
    params:
      module: [icmp]  # Look for a HTTP 200 response.
    static_configs:
      - targets:
        - prometheus.io
    relabel_configs:
        target_label: __param_target
      - source_labels: [__param_target]
        target_label: instance
      - target_label: __address__
```

Results in:
```
probe_success{instance="prometheus.io",job="blackbox-icmp"} 1
```

brian-brazil: I must have already fixed that then on the Prometheus side.

PR #164, opened 2017-06-06
Register probeSSLEarliestCertExpiryGauge

kblin:
This makes the gauge show up in the probe output when a TLS endpoint is
probed.

Signed-off-by: Kai Blin <kblin@biosustain.dtu.dk>


brian-brazil:
Thanks!

PR #169, opened 2017-06-21
Set default port of 53 if not specified for DNS target

Conorbro:


dns.go
```diff
@@ -115,7 +115,10 @@ func probeDNS(target string, module Module, registry *prometheus.Registry) bool
 	}
 
 	if module.DNS.TransportProtocol == "udp" || module.DNS.TransportProtocol == "tcp" {
-		targetAddr, _, _ := net.SplitHostPort(target)
+		targetAddr, port, _ := net.SplitHostPort(target)
```
brian-brazil: Hmm, the existing logic isn't correct. We should be using the `ip` below and then joining the port with it rather than using target down in client.Exchange

Conorbro:
@brian-brazil

PR #180, opened 2017-07-04
Ensure duration of lookup is set if ip resolution fails

Conorbro:


utils.go
```diff
@@ -47,7 +47,7 @@ func chooseProtocol(preferredIPProtocol string, target string, registry *prometh
 		}
 	}
 
-	probeDNSLookupTimeSeconds.Add(time.Since(resolveStart).Seconds())
+	defer probeDNSLookupTimeSeconds.Add(time.Since(resolveStart).Seconds())
```
brian-brazil: You still don't reach this line if the above return happens, this needs to move up.

Conorbro:
@brian-brazil

brian-brazil:
Thanks!

PR #181, opened 2017-07-05
Added timeout to blackbox exporter

Conorbro:
Added timeout set by Prometheus header to blackbox exporter.


main.go
```diff
@@ -108,26 +107,38 @@ func init() {
...
 	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds-*timeoutOffset)*time.Second)
+	defer cancel()
+	r = r.WithContext(ctx)
+	probeHandler(w, r, c)
+}
 
-	log.Infoln("Starting blackbox_exporter", version.Info())
-	log.Infoln("Build context", version.BuildContext())
+var timeoutOffset = flag.Float64("timeout-offset", 0.5, "Offset to subtract from timeout in seconds.")
```
brian-brazil: All the flags should be together.

config_test.go
```diff
@@ -25,3 +29,23 @@ func TestLoadBadConfig(t *testing.T) {
 		t.Errorf("\nexpected:\n%v\ngot:\n%v", expected, err.Error())
 	}
 }
+
+func TestPrometheusTimeout(t *testing.T) {
+	initConfig()
```
brian-brazil: This is pulling the file off the filesystem, it'd be better in the test to pass in the config to be used to probeRequestHandler. It's already setup to allow you to inject a config.

config_test.go
```diff
@@ -25,3 +29,23 @@ func TestLoadBadConfig(t *testing.T) {
...
+
+	req, err := http.NewRequest("GET", "/probe?module=http_2xx&target=http%3A%2F%2Fprometheus.io", nil)
```
brian-brazil: We shouldn't presume internet access, see the http tests for how to setup a test server

main.go
```diff
@@ -52,6 +54,10 @@ func (sc *SafeConfig) reloadConfig(confFile string) (err error) {
 		return err
 	}
 
+	if err := checkOverflow(c.XXX, "config"); err != nil {
```
brian-brazil: This belongs in Unmarshal

config.go
```diff
@@ -98,6 +99,18 @@ type DNSRRValidator struct {
 	XXX map[string]interface{} `yaml:",inline"`
 }
 
+func getModule(c *Config, r *http.Request, w http.ResponseWriter) Module {
```
brian-brazil: This is not related to parsing the config, it doesn't belong in this file

config_test.go
```diff
@@ -25,3 +29,23 @@ func TestLoadBadConfig(t *testing.T) {
 		t.Errorf("\nexpected:\n%v\ngot:\n%v", expected, err.Error())
 	}
 }
+
+func TestPrometheusTimeout(t *testing.T) {
```
brian-brazil: This doesn't belong in this file

config_test.go
```diff
@@ -25,3 +29,23 @@ func TestLoadBadConfig(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	req.Header.Set("X-Prometheus-Scrape-Timeout-Seconds", "5")
+
+	rr := httptest.NewRecorder()
+	handler := http.HandlerFunc(probeRequestHandler)
+	handler.ServeHTTP(rr, req)
+
+	if status := rr.Code; status != http.StatusOK {
```
brian-brazil: Make a endpoint that's purposefully slow and see it timeout

main.go
```diff
@@ -76,15 +82,8 @@ func probeHandler(w http.ResponseWriter, r *http.Request, conf *Config) {
...
-	moduleName := params.Get("module")
-	if moduleName == "" {
-		moduleName = "http_2xx"
-	}
-	module, ok := conf.Modules[moduleName]
-	if !ok {
-		http.Error(w, fmt.Sprintf("Unknown module %q", moduleName), 400)
-		return
-	}
+	module := getModule(conf, r, w)
```
brian-brazil: I don't think factoring this out is gaining much

main.go
```diff
@@ -108,26 +107,38 @@ func init() {
 	prometheus.MustRegister(version.NewCollector("blackbox_exporter"))
 }
 
-func main() {
+func probeRequestHandler(w http.ResponseWriter, r *http.Request) {
```
brian-brazil: I'd leave this inline with no changes. Splitting the logic across two functions is confusing.

main.go
```diff
@@ -108,26 +107,38 @@ func init() {
...
-		fmt.Fprintln(os.Stdout, version.Print("blackbox_exporter"))
-		os.Exit(0)
+	module := getModule(c, r, w)
+
+	if module.Timeout.Seconds() < timeoutSeconds {
+		timeoutSeconds = module.Timeout.Seconds()
 	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds-*timeoutOffset)*time.Second)
+	defer cancel()
+	r = r.WithContext(ctx)
```
brian-brazil: You need to use this context down in the probes

Conorbro:
@brian-brazil

dns_test.go
```diff
@@ -118,7 +119,10 @@ func TestRecursiveDNSResponse(t *testing.T) {
 			test.Probe.TransportProtocol = protocol
 			registry := prometheus.NewPedanticRegistry()
 			registry.Gather()
-			result := probeDNS(addr.String(), Module{Timeout: time.Second, DNS: test.Probe}, registry)
+
+			testCTX, cancel := context.WithTimeout(context.Background(), time.Duration(10)*time.Second)
```
brian-brazil: you can do `10 * time.Second`

main.go
```diff
@@ -76,15 +78,15 @@ func probeHandler(w http.ResponseWriter, r *http.Request, conf *Config) {
 		return
 	}
 
-	moduleName := params.Get("module")
+	moduleName := r.URL.Query().Get("module")
```
brian-brazil: this is available as params from above

main.go
```diff
@@ -108,15 +112,55 @@ func init() {
 	prometheus.MustRegister(version.NewCollector("blackbox_exporter"))
 }
 
+func probeRequestHandler(w http.ResponseWriter, r *http.Request) {
```
brian-brazil: You're still arbitrarily splitting logic across probeRequestHandler and probeHandler. All this logic should be in one function, with a small shim anonymous function in the HandleFunc.

main.go
```diff
@@ -108,15 +112,55 @@ func init() {
...
+	moduleName := r.URL.Query().Get("module")
+	if moduleName == "" {
+		moduleName = "http_2xx"
+	}
+	module, ok := c.Modules[moduleName]
+	if !ok {
+		http.Error(w, fmt.Sprintf("Unknown module %q", moduleName), 400)
+	}
+
+	if module.Timeout.Seconds() < timeoutSeconds && module.Timeout.Seconds() > 0 {
```
brian-brazil: Keep the timeout logic together

main.go
```diff
@@ -108,15 +112,55 @@ func init() {
...
+		timeoutSeconds = module.Timeout.Seconds()
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds-*timeoutOffset)*time.Second)
+	defer cancel()
+	r = r.WithContext(ctx)
+
+	probeHandler(ctx, w, r, c)
+}
+
+var timeoutOffset = flag.Float64("timeout-offset", 0.5, "Offset to subtract from timeout in seconds.")
```
brian-brazil: Keep all the flags together, they should be up near the top of this file in a var section.

main_test.go
```diff
@@ -0,0 +1,41 @@
...
+		"http_2xx": Module{
+			Prober:  "http",
+			Timeout: time.Duration(10 * time.Second),
+		},
+	},
+}
+
+func TestPrometheusTimeout(t *testing.T) {
+
+	sc.C = c
```
brian-brazil: Changing globals in tests is best avoided where possible. Pass in what you need to the function.

main_test.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	req.Header.Set("X-Prometheus-Scrape-Timeout-Seconds", "1")
+
+	rr := httptest.NewRecorder()
+	handler := http.HandlerFunc(probeRequestHandler)
+	handler.ServeHTTP(rr, req)
+
+	if status := rr.Code; status != http.StatusInternalServerError {
```
brian-brazil: This should return a 200.

Conorbro:
@brian-brazil

main.go
```diff
@@ -176,6 +199,10 @@ func main() {
 				http.Error(w, fmt.Sprintf("failed to reload config: %s", err), http.StatusInternalServerError)
 			}
 		})
+	http.Handle("/metrics", prometheus.Handler())
+	http.HandleFunc("/probe", func(w http.ResponseWriter, r *http.Request) {
+		probeHandler(w, r, nil)
```
brian-brazil: The config should continue to be passed in here

main.go
```diff
@@ -60,7 +72,44 @@ func (sc *SafeConfig) reloadConfig(confFile string) (err error) {
...
+
+	// If a timeout is configured via the Prometheus header, add it to the request.
+	var prometheusTimeout string
+	if r.Header["X-Prometheus-Scrape-Timeout-Seconds"] != nil {
+		prometheusTimeout = r.Header["X-Prometheus-Scrape-Timeout-Seconds"][0]
+	}
+
+	timeoutSeconds, err := strconv.ParseFloat(prometheusTimeout, 64)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to parse timeout from Prometheus header: %s", err), http.StatusInternalServerError)
```
brian-brazil: if you're using http.Error you need to `return` right after it.

main.go
```diff
@@ -60,7 +72,44 @@ func (sc *SafeConfig) reloadConfig(confFile string) (err error) {
...
+	}
+	if timeoutSeconds == 0 {
+		timeoutSeconds = 10
+	}
+
+	if module.Timeout.Seconds() < timeoutSeconds && module.Timeout.Seconds() > 0 {
+		timeoutSeconds = module.Timeout.Seconds()
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds-*timeoutOffset)*time.Second)
```
brian-brazil: You're using precision here as time.Duration is an int64. Multiplying it by 1e9 before casting would keep the precision.

main.go
```diff
@@ -76,17 +125,9 @@ func probeHandler(w http.ResponseWriter, r *http.Request, conf *Config) {
...
-		moduleName = "http_2xx"
-	}
-	module, ok := conf.Modules[moduleName]
-	if !ok {
-		http.Error(w, fmt.Sprintf("Unknown module %q", moduleName), 400)
-		return
-	}
 	prober, ok := Probers[module.Prober]
 	if !ok {
+		fmt.Println("JHere")
```
brian-brazil: test code

tcp.go
```diff
@@ -48,23 +48,26 @@ func dialTCP(target string, module Module, registry *prometheus.Registry) (net.C
...
 	if !module.TCP.TLS {
-		return dialer.Dial(dialProtocol, dialTarget)
+		return dialer.DialContext(ctx, dialProtocol, dialTarget)
 	}
 	tlsConfig, err := config.NewTLSConfig(&module.TCP.TLSConfig)
 	if err != nil {
 		return nil, err
 	}
+	timeoutDeadline, _ := ctx.Deadline()
+	dialer.Timeout = timeoutDeadline.Sub(time.Now())
```
brian-brazil: You can use dialer.Deadline to make it simpler

tcp.go
```diff
@@ -48,23 +48,26 @@ func dialTCP(target string, module Module, registry *prometheus.Registry) (net.C
...
 	return tls.DialWithDialer(dialer, dialProtocol, dialTarget, tlsConfig)
 }
 
-func probeTCP(target string, module Module, registry *prometheus.Registry) bool {
+func probeTCP(ctx context.Context, target string, module Module, registry *prometheus.Registry) bool {
 	probeSSLEarliestCertExpiry := prometheus.NewGauge(prometheus.GaugeOpts{
 		Name: "probe_ssl_earliest_cert_expiry",
 		Help: "Returns earliest SSL cert expiry date",
 	})
 	registry.MustRegister(probeSSLEarliestCertExpiry)
```
brian-brazil: Hmm, this should only be registered if we set it

Conorbro:
@brian-brazil

main.go
```diff
@@ -60,7 +72,45 @@ func (sc *SafeConfig) reloadConfig(confFile string) (err error) {
 	return nil
 }
 
-func probeHandler(w http.ResponseWriter, r *http.Request, conf *Config) {
+func probeHandler(w http.ResponseWriter, r *http.Request, c *Config) {
+
+	if c == nil {
```
brian-brazil: You don't need this.

main.go
```diff
@@ -60,7 +72,45 @@ func (sc *SafeConfig) reloadConfig(confFile string) (err error) {
...
+		sc.RUnlock()
+	}
+
+	moduleName := r.URL.Query().Get("module")
+	if moduleName == "" {
+		moduleName = "http_2xx"
+	}
+	module, ok := c.Modules[moduleName]
+	if !ok {
+		http.Error(w, fmt.Sprintf("Unknown module %q", moduleName), 400)
```
brian-brazil: Return after sending a http error

main.go
```diff
@@ -60,7 +72,45 @@ func (sc *SafeConfig) reloadConfig(confFile string) (err error) {
...
+		return
+	}
+	if timeoutSeconds == 0 {
+		timeoutSeconds = 10
+	}
+
+	if module.Timeout.Seconds() < timeoutSeconds && module.Timeout.Seconds() > 0 {
+		timeoutSeconds = module.Timeout.Seconds()
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration((timeoutSeconds-*timeoutOffset)*1e9)*time.Second)
```
brian-brazil: Don't multiply by time.Second, the 1e9 already covers that

main.go
```diff
@@ -176,6 +199,10 @@ func main() {
 				http.Error(w, fmt.Sprintf("failed to reload config: %s", err), http.StatusInternalServerError)
 			}
 		})
+	http.Handle("/metrics", prometheus.Handler())
+	http.HandleFunc("/probe", func(w http.ResponseWriter, r *http.Request) {
+		probeHandler(w, r, sc.C)
```
brian-brazil: You need to protect this access to sc.C with the mutex

brian-brazil:
This will need to be documented in the README too

Conorbro:
@brian-brazil

README.md
```diff
@@ -39,6 +39,8 @@ Additionally, an [example configuration](https://github.com/prometheus/blackbox_
 HTTP, HTTPS (via the `http` prober), DNS, TCP socket and ICMP (see permissions section) are currently supported.
 Additional modules can be defined to meet your needs.
 
+The timeout of each probe may be configured using `timeout` in the Blackbox exporter config file or preferably by setting the `scrape_timeout` in the [Prometheus config](https://prometheus.io/docs/operating/configuration/#configuration-file) itself.
```
brian-brazil: This doesn't quite capture the logic. 

The timeout of each probe is automatically determined from the `scrape_timeout` in the Prometheus config, slightly reduced to allow for network delays. This can be further limited by the `timeout` in the Blackbox exporter config file. If neither is specified, it defaults to 10 seconds.

main.go
```diff
@@ -176,6 +194,12 @@ func main() {
 				http.Error(w, fmt.Sprintf("failed to reload config: %s", err), http.StatusInternalServerError)
 			}
 		})
+	http.Handle("/metrics", prometheus.Handler())
+	http.HandleFunc("/probe", func(w http.ResponseWriter, r *http.Request) {
+		sc.RLock()
```
brian-brazil: The mutex is taken for too long here, this will block reloads while a probe is in progress. Only hold the mutex for enough time to copy over C.

Conorbro:
@brian-brazil

main.go
brian-brazil: This comment remains

brian-brazil:
The CI doesn't like this, can you run make format?

Conorbro:
@brian-brazil

brian-brazil:
Thanks!

PR #185, opened 2017-07-07
Added expose loaded blackbox configuration on /config

Conorbro:
/config now reveals loaded Blackbox exporter configuration in YAML format.

### DIFF
--- a/config.go
+++ b/config.go
@@ -22,12 +22,12 @@ type SafeConfig struct {
 }
 
 type Module struct {
-	Prober  string        `yaml:"prober"`
-	Timeout time.Duration `yaml:"timeout"`
-	HTTP    HTTPProbe     `yaml:"http"`
-	TCP     TCPProbe      `yaml:"tcp"`
-	ICMP    ICMPProbe     `yaml:"icmp"`
-	DNS     DNSProbe      `yaml:"dns"`
+	Prober  string        `yaml:"prober,omitempty"`
+	Timeout time.Duration `yaml:"timeout,omitempty"`
+	HTTP    HTTPProbe     `yaml:"http,omitempty"`
+	TCP     TCPProbe      `yaml:"tcp,omitempty"`
+	ICMP    ICMPProbe     `yaml:"icmp,omitempty"`
+	DNS     DNSProbe      `yaml:"dns,omitempty"`
 
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline"`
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -17,82 +17,82 @@ type Config struct {
...
 type SafeConfig struct {
-	sync.RWMutex
-	C *Config
+	sync.RWMutex `yaml:",omitempty"`
+	C            *Config `yaml:"config"`
 }
 
 type Module struct {
 	Prober  string        `yaml:"prober"`
 	Timeout time.Duration `yaml:"timeout"`
```
brian-brazil: this should be omit empty too

main.go
```diff
@@ -176,15 +176,24 @@ func main() {
...
-            <h1>Blackbox Exporter</h1>
-            <p><a href="/probe?target=prometheus.io&module=http_2xx">Probe prometheus.io for http_2xx</a></p>
-            <p><a href="/metrics">Metrics</a></p>
-            </body>
-            </html>`))
+			<head><title>Blackbox Exporter</title></head>
+			<body>
+			<h1>Blackbox Exporter</h1>
+			<p><a href="/probe?target=prometheus.io&module=http_2xx">Probe prometheus.io for http_2xx</a></p>
+			<p><a href="/metrics">Metrics</a></p>
```
brian-brazil: Add a link to the config

main.go
```diff
@@ -176,15 +176,24 @@ func main() {
...
+			</body>
+			</html>`))
+	})
+
+	http.HandleFunc("/config", func(w http.ResponseWriter, r *http.Request) {
+		c, err := yaml.Marshal(sc)
+		if err != nil {
+			log.Fatalf("Error marshalling configuration: %v", err)
```
brian-brazil: While this should be impossible, kill the process is probably a bit much. Log it and return a 500 I'd say.

Conorbro:
@brian-brazil

blackbox.yml
```diff
@@ -3,6 +3,10 @@ modules:
     prober: http
     timeout: 5s
     http:
+      method: POST
```
brian-brazil: This file is the default config, it shouldn't be changed for tests.

I'd put what you need in blackbox-good.yml

config.go
```diff
@@ -17,82 +19,82 @@ type Config struct {
 }
 
 type SafeConfig struct {
-	sync.RWMutex
-	C *Config
+	sync.RWMutex `yaml:",omitempty"`
```
brian-brazil: Why are you adding yaml tags here?

config.go
brian-brazil: This comment stands

config.go
```diff
@@ -109,6 +111,14 @@ func checkOverflow(m map[string]interface{}, ctx string) error {
 	return nil
 }
 
+func (sc *SafeConfig) marshalYAML() ([]byte, error) {
```
brian-brazil: You don't need this, the default marshalling is fine

config_test.go
```diff
@@ -25,3 +28,24 @@ func TestLoadBadConfig(t *testing.T) {
...
+		C: &Config{},
+	}
+
+	err := sc.reloadConfig("testdata/blackbox-good.yml")
+	if err != nil {
+		t.Errorf("Error loading config %v: %v", "testdata/blackbox-good.yml", err)
+	}
+
+	// String method must not reveal authentication credentials.
+	s, err := sc.marshalYAML()
```
brian-brazil: Marshall the config, not the wrapper around it.

Conorbro:
@brian-brazil

main.go
```diff
@@ -176,15 +176,27 @@ func main() {
...
+			<p><a href="/config">Configuration</a></p>
+			</body>
+			</html>`))
+	})
+
+	http.HandleFunc("/config", func(w http.ResponseWriter, r *http.Request) {
+		c, err := yaml.Marshal(sc.C)
+		if err != nil {
+			log.Warnf("Error marshalling configuration: %v", err)
+			http.Error(w, "", 500)
```
brian-brazil: Put the error in the http status also

main.go
```diff
@@ -176,15 +176,27 @@ func main() {
...
+			<h1>Blackbox Exporter</h1>
+			<p><a href="/probe?target=prometheus.io&module=http_2xx">Probe prometheus.io for http_2xx</a></p>
+			<p><a href="/metrics">Metrics</a></p>
+			<p><a href="/config">Configuration</a></p>
+			</body>
+			</html>`))
+	})
+
+	http.HandleFunc("/config", func(w http.ResponseWriter, r *http.Request) {
+		c, err := yaml.Marshal(sc.C)
```
brian-brazil: You need to protect this read with the mutex

Conorbro:
@brian-brazil

brian-brazil:
Thanks!

Issue #162, opened 2017-06-04
Expose HTTP/2 support


cceckman:
As the title says: it'd be nice to be able to tell whether a probe used HTTP/2, and even nicer to condition success / failure on it.

It's possible to explicitly configure probes for other "is this modern" dimensions. HTTP vs. HTTPS gives the fail_if_{not_,}ssl options, and specifying the scheme in the probe request allows explicit control over which gets probed. Even if left to the default, the probe_http_ssl variable exposes whether a request was or wasn't HTTPS. Similarly, a module can be configured for tcp, tcp4, or tcp6, and probe_ip_protocol exposes which was used.

HTTP/2 is another one of those dimensions of modernity, but I'm not finding anywhere where it gets percolated up- either through configuration of probes or through exported variables. As with the above, it would be nice to have both: a "force http/2" or "force http/1.1" option, and a variable showing which protocol was used.

PR #189, opened 2017-07-11
Expose HTTP version of probe

Conorbro:
Implements #162.

Exposes HTTP version via `probe_http_version` gauge and probes can now have set valid HTTP versions that will fail if not met.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -38,6 +38,9 @@ The other placeholders are specified separately.
   # Accepted status codes for this probe. Defaults to 2xx.
   [ valid_status_codes: <string>, ... | default = "2xx" ]
 
+  # Accepted HTTP versions for this probe.
+  [ valid_http_versions: <string>, ... ]
+
   # The HTTP method the probe will use.
   [ method: <string> | default = "GET" ]
 
--- a/config.go
+++ b/config.go
@@ -35,17 +35,18 @@ type Module struct {
 
 type HTTPProbe struct {
 	// Defaults to 2xx.
-	ValidStatusCodes       []int                   `yaml:"valid_status_codes"`
-	PreferredIPProtocol    string                  `yaml:"preferred_ip_protocol"`
-	NoFollowRedirects      bool                    `yaml:"no_follow_redirects"`
-	FailIfSSL              bool                    `yaml:"fail_if_ssl"`
-	FailIfNotSSL           bool                    `yaml:"fail_if_not_ssl"`
-	Method                 string                  `yaml:"method"`
-	Headers                map[string]string       `yaml:"headers"`
-	FailIfMatchesRegexp    []string                `yaml:"fail_if_matches_regexp"`
-	FailIfNotMatchesRegexp []string                `yaml:"fail_if_not_matches_regexp"`
-	Body                   string                  `yaml:"body"`
-	HTTPClientConfig       config.HTTPClientConfig `yaml:"http_client_config,inline"`
+	ValidStatusCodes       []int                   `yaml:"valid_status_codes,omitempty"`
+	ValidHTTPVersions      []string                `yaml:"valid_http_versions,omitempty"`
+	PreferredIPProtocol    string                  `yaml:"preferred_ip_protocol,omitempty"`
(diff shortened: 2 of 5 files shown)


CONFIGURATION.md
```diff
@@ -38,6 +38,9 @@ The other placeholders are specified separately.
   # Accepted status codes for this probe. Defaults to 2xx.
   [ valid_status_codes: <string>, ... | default = "2xx" ]
 
+  # Accepted HTTP versions for this probe.
+  [ valid_http_versions: <float64>, ... ]
```
brian-brazil: I think a string might make more sense here

http.go
```diff
@@ -85,12 +95,18 @@ func probeHTTP(target string, module Module, registry *prometheus.Registry) (suc
 			Name: "probe_ssl_earliest_cert_expiry",
 			Help: "Returns earliest SSL cert expiry in unixtime",
 		})
+
+		probeHTTPVersionGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_version",
+			Help: "Returns the version of HTTP the probe used",
```
brian-brazil: It's about the response, not the request

http.go
```diff
@@ -57,6 +58,15 @@ func matchRegularExpressions(reader io.Reader, httpConfig HTTPProbe) bool {
 	return true
 }
 
+func isValidHTTPVersion(probeVersion float64, validHTTPVersions []float64) bool {
```
brian-brazil: I'm not sure this needs to be a separate function

http.go
```diff
@@ -180,6 +195,18 @@ func probeHTTP(target string, module Module, registry *prometheus.Registry) (suc
...
+
+		httpVersion, err := strconv.ParseFloat(strings.TrimPrefix(resp.Proto, "HTTP/"), 64)
+		if err == nil {
+			probeHTTPVersionGauge.Set(httpVersion)
+		} else {
+			log.Errorf("Error parsing HTTP version from probe response: %v", err)
+		}
+
+		if len(httpConfig.ValidHTTPVersions) != 0 {
+			success = isValidHTTPVersion(httpVersion, httpConfig.ValidHTTPVersions)
```
brian-brazil: This could override a previous success = false, only set if it this fails.

Conorbro:
@brian-brazil

config.go
```diff
@@ -35,16 +35,17 @@ type Module struct {
...
-	PreferredIPProtocol    string                  `yaml:"preferred_ip_protocol,omitempty"`
-	NoFollowRedirects      bool                    `yaml:"no_follow_redirects,omitempty"`
-	FailIfSSL              bool                    `yaml:"fail_if_ssl,omitempty"`
-	FailIfNotSSL           bool                    `yaml:"fail_if_not_ssl,omitempty"`
-	Method                 string                  `yaml:"method,omitempty"`
-	Headers                map[string]string       `yaml:"headers,omitempty"`
-	FailIfMatchesRegexp    []string                `yaml:"fail_if_matches_regexp,omitempty"`
-	FailIfNotMatchesRegexp []string                `yaml:"fail_if_not_matches_regexp,omitempty"`
-	Body                   string                  `yaml:"body,omitempty"`
+	ValidStatusCodes       []int                   `yaml:"valid_status_codes"`
```
brian-brazil: These should still be omitempty

http.go
```diff
@@ -180,6 +186,31 @@ func probeHTTP(target string, module Module, registry *prometheus.Registry) (suc
 		if success && (len(httpConfig.FailIfMatchesRegexp) > 0 || len(httpConfig.FailIfNotMatchesRegexp) > 0) {
 			success = matchRegularExpressions(resp.Body, httpConfig)
 		}
+
+		if err == nil {
```
brian-brazil: What err is this?

example.yml
```diff
@@ -3,6 +3,7 @@ modules:
     prober: http
     timeout: 5s
     http:
+      valid_http_versions: []
```
brian-brazil: I'd include an example that made it clean the valid values were 1.1 etc.

Conorbro:
@brian-brazil

brian-brazil:
Thanks!

cceckman:
Thanks a bunch!

PR #191, opened 2017-07-12
Added metric for http/tcp regex match

Conorbro:
Implements #161


tcp.go
```diff
@@ -112,8 +117,10 @@ func probeTCP(ctx context.Context, target string, module Module, registry *prome
 				return false
 			}
 			if match == nil {
+				probeFailedDueToRegex.Set(0)
```
brian-brazil: The logic here doesn't seem right, isn't this a failure due to a mismatch?

Conorbro: My mistake, I had it backwards. 

It's fixed now!

PR #195, opened 2017-07-17
doc: Rename protocol to transport_protocol

sbadia:
Hi here, this PR fixes this issue with default config. « Error parsing config file: unknown fields in dns probe: protocol »

Refs: https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md#dns_probe


brian-brazil:
Thanks!

PR #197, opened 2017-07-18
Add an issue template.

SuperQ:


brian-brazil:
We should also ask for the module config, log output, and relevant bit of the Prometheus config.

SuperQ:
@brian-brazil Updated template.

brian-brazil:
Thanks!

PR #198, opened 2017-07-20
Add large link to config docs

SuperQ:
Make the Configuration section header link to the configuration doc.
* Use relative link[0].

[0]: https://help.github.com/articles/about-readmes/#relative-links-and-image-paths-in-readme-files


README.md
```diff
@@ -22,7 +22,7 @@ will return metrics for a HTTP probe against google.com. The `probe_success` met
     docker build -t blackbox_exporter .
     docker run -d -p 9115:9115 --name blackbox_exporter -v `pwd`:/config blackbox_exporter -config.file=/config/blackbox.yml
 
-## Configuration
+## [Configuration](CONFIGURATION.md)
 
 Blackbox exporter is configured via a configuration file and command-line flags (such as what configuration file to load, what port to listen on, and the logging format and level).
```
brian-brazil: Maybe also link "configuration file" here and remove the bit below?

I don't think we're using headers as links elsewhere, so it might be missed.

SuperQ: PTAL

brian-brazil:
We'd talked about further improving this, but given its been a while I'm going to merge this as it's already a notable improvement, rather than waiting for it to be perfect.

PR #199, opened 2017-07-25
Change to Kingpin flags

carlpett:
As per [Prometheus #2455](https://github.com/prometheus/prometheus/issues/2455), this is a conversion from `flag` to `kingpin`.

I tried keeping it as close to the dev-2.0 branch of Prometheus as possible. However, there everything is put into a configuration object, while blackbox_exporter currently uses package scope variables. I kept it that way, but it could also be changed by either creating a new config struct, or merging it into the yaml-backed struct, if that seems preferable.


brian-brazil:
I think the dockerfile and readme also need updating.

brian-brazil:
The new dependency also needs vendoring.

carlpett:
Yes, I just wanted to validate that this is in line with how it should be done first :) If it looks ok I'll go ahead and update the rest. The changelog should contain some mention about it being a breaking change, also.

main.go
```diff
@@ -37,10 +38,11 @@ var (
 	sc = &SafeConfig{
 		C: &Config{},
 	}
-	configFile    = flag.String("config.file", "blackbox.yml", "Blackbox exporter configuration file.")
-	listenAddress = flag.String("web.listen-address", ":9115", "The address to listen on for HTTP requests.")
-	showVersion   = flag.Bool("version", false, "Print version information.")
-	timeoutOffset = flag.Float64("timeout-offset", 0.5, "Offset to subtract from timeout in seconds.")
+
+	// Flags
+	configFile    string
```
brian-brazil: Could you use this style: https://github.com/alecthomas/kingpin#simple-example
rather than spreading each flag definition around?

carlpett: I did initially, but it does not combine with the usage in the main Prometheus repo, where a Kingpin `application` is defined, and the flags need to be created with calls on this application instance.
I agree that pre-declaring the flags isn't very nice, which is what I referred to in the initial comment. To avoid it, we can either skip using the application (but diverge from how it is done in Prometheus), or move the variables into a config object, which allows us to exactly follow the Prometheus usage. But then I wonder if you'd prefer to merge it into the existing Config or SafeConfig types, or create a new one? This struct will need to be passed into the probeHandler func, as well.

brian-brazil: I like my flags simple, and this is a simple application. There's no need to get fancy.

carlpett: Great :) I've pushed a new simpler version now.

README.md
```diff
@@ -20,7 +20,7 @@ will return metrics for a HTTP probe against google.com. The `probe_success` met
 ### Building with Docker
 
     docker build -t blackbox_exporter .
-    docker run -d -p 9115:9115 --name blackbox_exporter -v `pwd`:/config blackbox_exporter -config.file=/config/blackbox.yml
+    docker run -d -p 9115:9115 --name blackbox_exporter -v `pwd`:/config blackbox_exporter --config.file=/config/blackbox.yml
```
brian-brazil: On line 34 the flag needs to be updated

carlpett: Had that unsaved in the editor... Pushed now.

brian-brazil:
Thanks!

PR #201, opened 2017-07-27
Fix logging flags

carlpett:
The logging flags are no longer automatically set, which was missed in the kingpin conversion.


brian-brazil:
I've the same comments as for https://github.com/prometheus/influxdb_exporter/pull/15

carlpett:
Updated to the newly merged common/log and call `log.AddFlags` instead.

brian-brazil:
Thanks!

PR #206, opened 2017-08-08
Fix wrong config flag name in 0.8.0 (--config-flag=..)

fabianlaule:
Some minutes ago, my monitoring system notified me about problems with the blackbox_exporter which updates automatically on new releases. It seems that in version 0.8.0, the flag for the config file was accidentally renamed to `--config-flag`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,7 +38,7 @@ var (
 		C: &Config{},
 	}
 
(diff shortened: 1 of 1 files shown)


grobie:
Thank you!

PR #207, opened 2017-08-08
Fix config.file command line option name.

sysadmind:
This appears to have been incorrectly modified when updating the configuration flags in ae0f6bff1498237a185e0a7496cdba09ced22d4b


sysadmind:
Looks like this was fixed in #206

PR #209, opened 2017-08-14
Added check for empty query_name and where targetAddr may be empty

Conorbro:


dns.go
```diff
@@ -109,9 +109,13 @@ func probeDNS(ctx context.Context, target string, module Module, registry *prome
 		probeDNSAdditionalRRSGauge.Set(float64(numAdditional))
 	}()
 
+	if module.DNS.QueryName == "" {
```
brian-brazil: This should be part of the config validation, we shouldn't accept a config with this missing.

dns.go
```diff
@@ -120,6 +124,9 @@ func probeDNS(ctx context.Context, target string, module Module, registry *prome
 		if port == "" {
 			port = "53"
 		}
+		if targetAddr == "" {
+			targetAddr = target
```
brian-brazil: Combine this with the previous if, and check err instead

Conorbro:
@brian-brazil

dns.go
```diff
@@ -110,15 +110,16 @@ func probeDNS(ctx context.Context, target string, module Module, registry *prome
...
 	}
 	if module.DNS.TransportProtocol == "udp" || module.DNS.TransportProtocol == "tcp" {
-		targetAddr, port, _ := net.SplitHostPort(target)
-		if port == "" {
+		targetAddr, port, err := net.SplitHostPort(target)
+		if err != nil {
+			// Target only contains host so fallback to default port and set targetAddr as target.
 			port = "53"
+			targetAddr = target
+			log.Error(err)
```
brian-brazil: This is an expected thing, so don't log it.

brian-brazil:
Thanks!

PR #213, opened 2017-08-26
Update build

SuperQ:
* Update Makefile.
* Update travis config.
* Fix staticcheck errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,10 @@ sudo: false
 language: go
 
 go:
-- 1.7.5
-- 1.8
+- 1.8.x
+- master
+
+go_import_path: github.com/prometheus/blackbox_exporter
 
 script:
 - make
--- a/Makefile
+++ b/Makefile
@@ -11,10 +11,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-GO     := GO15VENDOREXPERIMENT=1 go
-GOPATH := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))
-PROMU  ?= $(GOPATH)/bin/promu
-pkgs    = $(shell $(GO) list ./... | grep -v /vendor/)
+GO           := GO15VENDOREXPERIMENT=1 go
+FIRST_GOPATH := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))
+PROMU        := $(FIRST_GOPATH)/bin/promu
+STATICCHECK  := $(FIRST_GOPATH)/bin/staticcheck
+pkgs         = $(shell $(GO) list ./... | grep -v /vendor/)
 
 PREFIX                  ?= $(shell pwd)
 BIN_DIR                 ?= $(shell pwd)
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -3,8 +3,10 @@ sudo: false
 language: go
 
 go:
-- 1.7.5
-- 1.8
+- 1.8.x
```
mdlayher: Now that 1.9 is out, do you want to just do `1.x` or also add `1.9.x`?

tcp_test.go
```diff
@@ -35,7 +35,7 @@ func TestTCPConnection(t *testing.T) {
 	go func() {
 		conn, err := ln.Accept()
 		if err != nil {
-			t.Fatalf("Error accepting on socket: %s", err)
+			panic(fmt.Sprintf("Error accepting on socket: %s", err))
```
mdlayher: I sometimes make a `panicf` helper for these cases, but it isn't super necessary for just a handful.

brian-brazil: Why are these changing? They look correct to me.

Makefile
```diff
@@ -52,7 +59,16 @@ docker:
 	@echo ">> building docker image"
 	@docker build -t "$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" .
 
-$(PROMU) promu:
+$(GOPATH)/bin/promu promu:
 	@GOOS= GOARCH= $(GO) get -u github.com/prometheus/promu
 
-.PHONY: all style format build test vet tarball docker promu
+$(GOPATH)/bin/staticcheck:
```
brian-brazil: This is not in the standard makefile, please keep in sync with upstream

tcp_test.go
SuperQ:
```
the goroutine calls T.Fatalf, which must be called in the same goroutine as the test (SA2002)
```

Makefile
SuperQ: [staticcheck](https://staticcheck.io/) is used in several other exporters.

brian-brazil: It's not in the promu or prometheus makefiles, I don't want exporters to get their makefiles out of sync.

Makefile
```diff
@@ -11,18 +11,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-GO     := GO15VENDOREXPERIMENT=1 go
+GO     ?= GO15VENDOREXPERIMENT=1 go
 GOPATH := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))
-PROMU  ?= $(GOPATH)/bin/promu
-pkgs    = $(shell $(GO) list ./... | grep -v /vendor/)
+GOARCH := $(shell $(GO) env GOARCH)
```
brian-brazil: There's FIRST_GOPATH stuff in promu, we should keep this all in sync

SuperQ:
@brian-brazil Ok, I've synced up the Makefile with https://github.com/prometheus/prometheus/pull/3125.

brian-brazil:
CI is failing, looks like the Makefile from promu isn't all there.

SuperQ:
Yea, working on it.

.travis.yml
SuperQ: Looks like we still need 1.9 in our tools.

PR #214, opened 2017-09-01
Move prober and config to own package

### LINKED ISSUE #205: Extract prober from exporter
Hi,

I've created my own fork of the blackbox_exporter to modify it for my needs but would like to re-use as much as possible from the original exporter as well as contribute back prober and prober changes/fixes which is hard right now.

Therefor I propose to extract the probers into their own package that can be reused.
We could create an interface `ProberFn(ctx context.Context, target string, config Config, registry *prometheus.Registry)` where Config would be a prober specific config struct that is referenced in the blackbox_exporter config struct.

This change would be fairly light on the blackbox_exporter and arguably it's main namespace is quite big already so IMO moving probers in their own package makes sense anyway. It would also make the exporter a bit more testable.

Thoughts?

discordianfish:
This closes #205


prober/dns.go
```diff
@@ -11,21 +11,23 @@
...
+package prober
 
 import (
 	"context"
 	"net"
 	"regexp"
 	"time"
 
-	"github.com/miekg/dns"
+	"github.com/prometheus/blackbox_exporter/config"
```
brian-brazil: This is an internal dependency, so should be separated from the other 3

main.go
```diff
@@ -27,52 +26,32 @@ import (
...
-	if err != nil {
-		log.Errorf("Error reading config file: %s", err)
-		return err
-	}
 
-	if err := yaml.Unmarshal(yamlFile, c); err != nil {
-		log.Errorf("Error parsing config file: %s", err)
-		return err
+	Probers = map[string]prober.ProbeFn{
+		"http": prober.ProbeFn(prober.ProbeHTTP),
```
brian-brazil: Why do you wrap these in ProbeFn here?

discordianfish: Oh, though I had to. Fixing.

discordianfish:
@brian-brazil ok, moved the internal dependencies down.

PR #215, opened 2017-09-01
Add httptrace based prober

discordianfish:
Hi,

this is pretty much the prober I use for the latency.at exporters. I've added it as a new module here although in my exporters I replaced the http module which might make sense, given that it basically only adds one new metric.

Beside that I still need refactor and move over some tests. But wanted to get feedback early. I expected especially the redirect stuff to something to discuss.

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -338,6 +338,10 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		if i != 0 {
 			durationGaugeVec.WithLabelValues("resolve").Add(trace.dnsDone.Sub(trace.start).Seconds())
 		}
+		// Skip trace if we never got a connection because a request failed.
+		if trace.gotConn.IsZero() {
+			continue
+		}
 		if trace.tls {
 			durationGaugeVec.WithLabelValues("connect").Add(trace.connectDone.Sub(trace.dnsDone).Seconds())
 			durationGaugeVec.WithLabelValues("tls").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
--- a/prober/http.go
+++ b/prober/http.go
@@ -131,7 +131,7 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	var redirects int
 	var (
 		durationGaugeVec = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-			Name: "duration_seconds",
+			Name: "probe_http_duration_seconds",
 			Help: "Duration of http request by phase, summed over all redirects",
 		}, []string{"phase"})
 		contentLengthGauge = prometheus.NewGauge(prometheus.GaugeOpts{
@@ -168,10 +168,6 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Name: "probe_failed_due_to_regex",
 			Help: "Indicates if probe failed due to regex",
 		})
-		probeIPProtocolGauge = prometheus.NewGauge(prometheus.GaugeOpts{
-			Name: "probe_ip_protocol",
-			Help: "Specifies whether probe ip protocol is IP4 or IP6",
-		})
 	)
 
 	registry.MustRegister(durationGaugeVec)
@@ -181,7 +177,6 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	registry.MustRegister(statusCodeGauge)
 	registry.MustRegister(probeHTTPVersionGauge)
 	registry.MustRegister(probeFailedDueToRegex)
-	registry.MustRegister(probeIPProtocolGauge)
 
 	httpConfig := module.HTTP
 
@@ -200,11 +195,12 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		targetHost = targetURL.Host
 	}
 
-	ip, err := chooseProtocolMetrics(module.HTTP.PreferredIPProtocol, targetHost, durationGaugeVec.WithLabelValues("resolve"), probeIPProtocolGauge, logger)
+	ip, lookupTime, err := chooseProtocol(module.HTTP.PreferredIPProtocol, targetHost, registry, logger)
 	if err != nil {
 		level.Error(logger).Log("msg", "Error resolving address", "err", err)
 		return false
 	}
+	durationGaugeVec.WithLabelValues("resolve").Add(lookupTime)
 
 	httpClientConfig := module.HTTP.HTTPClientConfig
 	if len(httpClientConfig.TLSConfig.ServerName) == 0 {
@@ -334,25 +330,13 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if resp == nil {
 		resp = &http.Response{}
 	}
-	var (
-		earlierstCertExpiry time.Time
-		i                   = 0
-	)
+	i := 0
 	for _, trace := range tt.traces {
 		// We get the duration for the first request from chooseProtocol
 		if i != 0 {
 			durationGaugeVec.WithLabelValues("resolve").Add(trace.dnsDone.Sub(trace.start).Seconds())
 		}
 		if resp.TLS != nil {
-			exp := getEarliestCertExpiry(resp.TLS)
-			if earlierstCertExpiry.IsZero() || earlierstCertExpiry.Before(exp) {
-				earlierstCertExpiry = exp
-			}
-
-			probeSSLEarliestCertExpiryGauge.Set(float64(getEarliestCertExpiry(resp.TLS).UnixNano() / 1e9))
-			if httpConfig.FailIfSSL {
-				success = false
-			}
 			durationGaugeVec.WithLabelValues("connect").Add(trace.connectDone.Sub(trace.dnsDone).Seconds())
 			durationGaugeVec.WithLabelValues("tls").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
 		} else {
--- a/prober/http.go
+++ b/prober/http.go
@@ -349,17 +349,22 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		if i != 0 {
 			durationGaugeVec.WithLabelValues("resolve").Add(trace.dnsDone.Sub(trace.start).Seconds())
 		}
-		// Skip trace if we never got a connection because a request failed.
+		// Continue here if we never got a connection because a request failed.
 		if trace.gotConn.IsZero() {
 			continue
 		}
 		if trace.tls {
+			// dnsDone must be set if gotConn was set
 			durationGaugeVec.WithLabelValues("connect").Add(trace.connectDone.Sub(trace.dnsDone).Seconds())
 			durationGaugeVec.WithLabelValues("tls").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
 		} else {
 			durationGaugeVec.WithLabelValues("connect").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
 		}
 
+		// Continue here if we never got a response from the server.
+		if trace.responseStart.IsZero() {
+			continue
+		}
 		durationGaugeVec.WithLabelValues("processing").Add(trace.responseStart.Sub(trace.gotConn).Seconds())
 		durationGaugeVec.WithLabelValues("transfer").Add(trace.end.Sub(trace.responseStart).Seconds())
 		i++
--- a/prober/http.go
+++ b/prober/http.go
@@ -80,14 +80,16 @@ type trace struct {
 // transport is a custom transport keeping traces for each http roundtrip.
 type transport struct {
 	Transport http.RoundTripper
+	logger    log.Logger
 	traces    map[*http.Request]*trace
 	current   *trace
 }
 
-func newTransport(rt http.RoundTripper) *transport {
+func newTransport(rt http.RoundTripper, logger log.Logger) *transport {
 	return &transport{
-		traces:    make(map[*http.Request]*trace),
 		Transport: rt,
+		logger:    logger,
+		traces:    make(map[*http.Request]*trace),
 	}
 }
 
@@ -98,6 +100,7 @@ func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 	t.traces[req] = &trace{}
 	t.current = t.traces[req]
+	level.Debug(t.logger).Log("msg", "Processing request", "url", req.URL.String())
 	defer func() { t.current.end = time.Now() }()
 	return t.Transport.RoundTrip(req)
 }
@@ -215,7 +218,7 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	}
 
 	// Inject transport that tracks trace for each redirect
-	tt := newTransport(client.Transport)
+	tt := newTransport(client.Transport, logger)
 	client.Transport = tt
 
 	client.CheckRedirect = func(r *http.Request, via []*http.Request) error {
--- a/prober/http.go
+++ b/prober/http.go
@@ -67,8 +67,9 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
 	return true
 }
 
-// trace holds timings for a single http roundtrip.
+// trace holds timings for a single HTTP roundtrip.
 type trace struct {
+	tls           bool
 	start         time.Time
 	dnsDone       time.Time
 	connectDone   time.Time
@@ -77,7 +78,7 @@ type trace struct {
 	end           time.Time
 }
 
-// transport is a custom transport keeping traces for each http roundtrip.
+// transport is a custom transport keeping traces for each HTTP roundtrip.
 type transport struct {
 	Transport http.RoundTripper
 	logger    log.Logger
@@ -95,12 +96,11 @@ func newTransport(rt http.RoundTripper, logger log.Logger) *transport {
 
 // RoundTrip switches to a new trace, then runs embedded RoundTripper.
 func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
-	if _, ok := t.traces[req]; ok {
-		return nil, errors.New("Redirect loop detected")
-	}
 	t.traces[req] = &trace{}
 	t.current = t.traces[req]
-	level.Debug(t.logger).Log("msg", "Processing request", "url", req.URL.String())
+	if req.URL.Scheme == "https" {
+		t.current.tls = true
+	}
 	defer func() { t.current.end = time.Now() }()
 	return t.Transport.RoundTrip(req)
 }
@@ -113,14 +113,13 @@ func (t *transport) DNSDone(_ httptrace.DNSDoneInfo) {
 }
 func (ts *transport) ConnectStart(_, _ string) {
 	t := ts.current
-	// No DNS resolution, e.g connecting to IP directly
+	// No DNS resolution, e.g. connecting to IP directly
 	if t.dnsDone.IsZero() {
 		t.start = time.Now()
 		t.dnsDone = t.start
 	}
 }
 func (t *transport) ConnectDone(net, addr string, err error) {
-	// t.current.addr = addr
 	t.current.connectDone = time.Now()
 }
 func (t *transport) GotConn(_ httptrace.GotConnInfo) {
@@ -217,7 +216,7 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		return false
 	}
 
-	// Inject transport that tracks trace for each redirect
+	// Inject transport that tracks trace for each redirect.
 	tt := newTransport(client.Transport, logger)
 	client.Transport = tt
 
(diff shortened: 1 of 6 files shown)


brian-brazil:
Yeah, it doesn't make sense to have this as a new module. Metrics should Just Work from the user standpoint.
My inclination would be not to use labels for these, as the existing dns metric doesn't have them and it'd make us less likely to forget to set one.
We should mention in the help strings how this interacts with redirects, e.g. the stats are for the first request.

discordianfish:
@brian-brazil I've change the PR to extend the existing http probe. From a user perspective it should be just a new metric that appears. I also simplified the tracing a bit.

Which labels do you mean? The duration labels should be fine, given it's a new metrics, right?
For the SSL metrics in the redirect we could either keep it as it is (which is stats for *last* request though, or do I miss something?) and document it in the help string, or we could add the redirect number label too but it would be a breaking change. I don't feel strongly either way.

brian-brazil:
> Which labels do you mean? The duration labels should be fine, given it's a new metrics, right?

I mean the labels on durationGaugeVec so that it matches the existing dns duration metric, no need to duplicate that information and cause confusion.

> which is stats for last request though, or do I miss something?

Yes, it's the last request we get there as that's what the library gives us.

prober/http.go
```diff
@@ -128,7 +200,7 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		targetHost = targetURL.Host
 	}
 
-	ip, err := chooseProtocol(module.HTTP.PreferredIPProtocol, targetHost, registry, logger)
+	ip, err := chooseProtocolMetrics(module.HTTP.PreferredIPProtocol, targetHost, durationGaugeVec.WithLabelValues("resolve"), probeIPProtocolGauge, logger)
```
brian-brazil: This removes the existing duration gauge, which needs to stay around. If you need the duration, get the function to pass it back rather than splitting the logic in two and being inconsistent.

prober/http.go
```diff
@@ -65,9 +67,73 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+func (t *transport) GotFirstResponseByte() {
+	t.current.responseStart = time.Now()
+}
+
 func ProbeHTTP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var redirects int
 	var (
+		durationGaugeVec = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "duration_seconds",
+			Help: "Duration of http request by phase, summed over all redirects",
```
brian-brazil: if it's http then http should be in the metric name

prober/http.go
```diff
@@ -243,6 +330,38 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if resp == nil {
 		resp = &http.Response{}
 	}
+	var (
+		earlierstCertExpiry time.Time
+		i                   = 0
+	)
+	for _, trace := range tt.traces {
+		// We get the duration for the first request from chooseProtocol
```
brian-brazil: It's missing that from here then, so isn't giving a full view

prober/http.go
```diff
@@ -243,6 +330,38 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if resp == nil {
 		resp = &http.Response{}
 	}
+	var (
+		earlierstCertExpiry time.Time
+		i                   = 0
+	)
+	for _, trace := range tt.traces {
```
brian-brazil: This should all be logged

prober/http.go
```diff
@@ -243,6 +330,38 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+		// We get the duration for the first request from chooseProtocol
+		if i != 0 {
+			durationGaugeVec.WithLabelValues("resolve").Add(trace.dnsDone.Sub(trace.start).Seconds())
+		}
+		if resp.TLS != nil {
+			exp := getEarliestCertExpiry(resp.TLS)
```
brian-brazil: This is conflicting with the logic below, as there's two different sets of TLS expiration logic. Keep just the existing one that looks only at the last request, for simplicity

prober/http.go
```diff
@@ -243,6 +330,38 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+			if earlierstCertExpiry.IsZero() || earlierstCertExpiry.Before(exp) {
+				earlierstCertExpiry = exp
+			}
+
+			probeSSLEarliestCertExpiryGauge.Set(float64(getEarliestCertExpiry(resp.TLS).UnixNano() / 1e9))
+			if httpConfig.FailIfSSL {
+				success = false
+			}
+			durationGaugeVec.WithLabelValues("connect").Add(trace.connectDone.Sub(trace.dnsDone).Seconds())
+			durationGaugeVec.WithLabelValues("tls").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
```
brian-brazil: This is going to be inconsistent with the name of the equivalent metric for the tcp probe

prober/http.go
discordianfish: Yes, the idea was to use the http durationGauge with 'resolve' label for that instead.
I could expose both?
I definitely would want to have the resolution in the time in the overall http duration metric. It's just so much more convienient to consume.

prober/http.go
discordianfish: Not sure what you mean. At this point we already recorded the time to resolve the first redirect in chooseProtocolMetrics, so that should be right, no?

prober/http.go
discordianfish: Oh right, missed that when rebasing.

prober/http.go
discordianfish: Yes.. As I said, I feel like the http duration metric is with these labels much more convinient to use.
We could also make the tcp probe and all other probes put these timings into one duration gauge for consistency. What do you think about that? That would be a breaking change though.

brian-brazil: I'm not sure what the best way forward is here, http having multiple requests messes things up a bit. Http phases don't also apply to other probes.

discordianfish: > I'm not sure what the best way forward is here, http having multiple requests messes things up a bit

I'm already summing up the times for the redirects, so there is only one timeseries per phase.


> Http phases don't also apply to other probes.

Every probe has some sort of phases, right? They are even similar: resolve/connect/tls/processing/transfer should apply 1:1 to tcp probe. A subset of that would work for udp too. For ICMP we could still have resolve/waiting or something like that. DNS is tricky though given that Exchange() is pretty high level. But that could still use the same gauge with a single label.

In the end I'm fine with having 'inconsistent' metrics or just expose times like tls and dns twice.

brian-brazil: > resolve/connect/tls/processing/transfer should apply 1:1 to tcp probe

Not quite, HTTP is a single query and response whereas TCP can have multiple queries and responses in what HTTP would consider "transfer".

discordianfish: Oh right now I see what you mean, didn't use the tcp probes much.
There is probably no way to get something consistent here. I'm leaning towards just exposing the tls and dns timings twice.

brian-brazil: That's probably the best route, we'll just need to be clear in the HELP strings about how they're different.

discordianfish:
Okay, I think I addressed all feedback. This also logs each request on debug level.

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
 	return true
 }
 
+// trace holds timings for a single http roundtrip.
```
brian-brazil: HTTP

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+type trace struct {
+	start         time.Time
+	dnsDone       time.Time
+	connectDone   time.Time
+	gotConn       time.Time
+	responseStart time.Time
+	end           time.Time
+}
+
+// transport is a custom transport keeping traces for each http roundtrip.
```
brian-brazil: Here too

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+	return &transport{
+		Transport: rt,
+		logger:    logger,
+		traces:    make(map[*http.Request]*trace),
+	}
+}
+
+// RoundTrip switches to a new trace, then runs embedded RoundTripper.
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if _, ok := t.traces[req]; ok {
```
brian-brazil: There's already a limit on the number of redirects, so we don't also need this.

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+func (ts *transport) ConnectStart(_, _ string) {
+	t := ts.current
+	// No DNS resolution, e.g connecting to IP directly
+	if t.dnsDone.IsZero() {
+		t.start = time.Now()
+		t.dnsDone = t.start
+	}
+}
+func (t *transport) ConnectDone(net, addr string, err error) {
+	// t.current.addr = addr
```
brian-brazil: Old debug code?

prober/http.go
```diff
@@ -146,6 +217,10 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		return false
 	}
 
+	// Inject transport that tracks trace for each redirect
```
brian-brazil: Full stop

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+}
+
+// RoundTrip switches to a new trace, then runs embedded RoundTripper.
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if _, ok := t.traces[req]; ok {
+		return nil, errors.New("Redirect loop detected")
+	}
+	t.traces[req] = &trace{}
+	t.current = t.traces[req]
+	level.Debug(t.logger).Log("msg", "Processing request", "url", req.URL.String())
```
brian-brazil: We already have this logged from the redirect logic

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+
+func (t *transport) DNSStart(_ httptrace.DNSStartInfo) {
+	t.current.start = time.Now()
+}
+func (t *transport) DNSDone(_ httptrace.DNSDoneInfo) {
+	t.current.dnsDone = time.Now()
+}
+func (ts *transport) ConnectStart(_, _ string) {
+	t := ts.current
+	// No DNS resolution, e.g connecting to IP directly
```
brian-brazil: e.g.

prober/http.go
```diff
@@ -247,6 +333,26 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if resp == nil {
 		resp = &http.Response{}
 	}
+	i := 0
+	for _, trace := range tt.traces {
+		// We get the duration for the first request from chooseProtocol
+		if i != 0 {
+			durationGaugeVec.WithLabelValues("resolve").Add(trace.dnsDone.Sub(trace.start).Seconds())
+		}
+		if resp.TLS != nil {
```
brian-brazil: Just because the last response wasn't TLS doesn't mean previous ones were, you always should add in tls

prober/http.go
```diff
@@ -247,6 +333,26 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+			durationGaugeVec.WithLabelValues("connect").Add(trace.connectDone.Sub(trace.dnsDone).Seconds())
+			durationGaugeVec.WithLabelValues("tls").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
+		} else {
+			durationGaugeVec.WithLabelValues("connect").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
+			if httpConfig.FailIfNotSSL {
```
brian-brazil: This logic is already handled below

prober/utils.go
```diff
@@ -11,9 +11,8 @@ import (
 )
 
 // Returns the preferedIPProtocol, the dialProtocol, and sets the probeIPProtocolGauge.
```
brian-brazil: This comment needs updating

discordianfish:
Note to myself: Looks like something is broken with the trace metrics for IPv6:
```
```
level=info msg="Resolving target address" preferred_ip_protocol=ip6
level=info msg="Resolved target address" ip=2a00:1450:4016:80a::200e
level=info msg="Making HTTP request" url=http://[2a00:1450:4016:80a::200e] host=google.com
level=debug msg="Processing request" url=http://[2a00:1450:4016:80a::200e]
level=error msg="Error for HTTP request" err="Get http://[2a00:1450:4016:80a::200e]: dial tcp [2a00:1450:4016:80a::200e]:80: connect: network is unreachable"
```

```
probe_http_duration_seconds{phase="connect"} -9.223372036854776e+09
probe_http_duration_seconds{phase="processing"} 0
probe_http_duration_seconds{phase="resolve"} 0.021426744
probe_http_duration_seconds{phase="transfer"} 9.223372036854776e+09
```

File: prober/http.go (hunk 14 above)
Comment (discordianfish) [reply prober/http.go]: Oh good catch. I think we need to track which trace involved TLS.

**Comment by discordianfish** [comment]
I've addressed the comments but still need to look into the IPV6 issue.

**Comment by discordianfish** [comment]
@brian-brazil How is the chooseProtocol logic suppose to work?

It looks like by default it sets
```
                preferredIPProtocol = "ip6"
                fallbackProtocol = "ip4"
```

in which case it just resolves the name and returns it. I expected it somehow checks whether IPv6 is working? I guess that's as intended and people need to explicitly set preferred protocol to v4 if their hosts have no IPv6 support?

**Comment by brian-brazil** [comment]
> I guess that's as intended and people need to explicitly set preferred protocol to v4 if their hosts have no IPv6 support?

Yes, that's how it works.

**Comment by discordianfish** [comment]
The problem with ipv6 was due to the traces being empty when the request failed. I'm not skipping traces where gotConn is zero.

**Comment by brian-brazil** [comment]
What if there's resolution time and then the request fails?

**Comment by discordianfish** [comment]
@brian-brazil Not sure if you saw the earlier version but I moved the skipping below recording of resolution, so you would still get resolution times even if the request itself fails.
Arguably the request could also fail later and it would make sense to expose timings for connect etc but I feel that's not worth the more complicated code. But if you feel different, I can look into doing that.

I've also added a test. This only checks if the metric+values exist though. I don't think that sleeping and fuzzy matching the timings would be the right way. What do you think?

File: prober/http.go [hunk 17]
```diff
@@ -65,9 +67,75 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+	gotConn       time.Time
+	responseStart time.Time
+	end           time.Time
+}
+
+// transport is a custom transport keeping traces for each HTTP roundtrip.
+type transport struct {
+	Transport http.RoundTripper
+	logger    log.Logger
+	traces    map[*http.Request]*trace
```
brian-brazil: This should be a list rather than a map

prober/http.go
```diff
@@ -65,9 +67,75 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+
+func (t *transport) DNSStart(_ httptrace.DNSStartInfo) {
+	t.current.start = time.Now()
+}
+func (t *transport) DNSDone(_ httptrace.DNSDoneInfo) {
+	t.current.dnsDone = time.Now()
+}
+func (ts *transport) ConnectStart(_, _ string) {
+	t := ts.current
+	// No DNS resolution, e.g. connecting to IP directly
```
brian-brazil: Full stop at the end of comments

prober/http.go
```diff
@@ -247,6 +332,27 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if resp == nil {
 		resp = &http.Response{}
 	}
+	i := 0
+	for _, trace := range tt.traces {
+		// We get the duration for the first request from chooseProtocol
```
brian-brazil: You should log the times too, so the user can get back to the individual request times

prober/http.go
discordianfish: Good point, I've changed it.

prober/http.go
discordianfish: I've added logging the raw timestamps which is probably most useful.

discordianfish:
@brian-brazil Okay, so from my perspective this is ready to get merged. Let me know if there is more to change, preferably in one batch so I can address all remaining changes at once.

discordianfish:
Okay hold off on merging this. Look like we have similar issues with the other timings: A connection can break at any point, so we always need to check if a timing is available or not. Will update the PR in a bit.

prober/http.go
brian-brazil: I think the diffs would be more useful to a human.

Rudd-O:
This looks quite useful.

discordianfish:
I think this should do it. I've deployed it here and will report back if things look good in prod too.

discordianfish:
I've been running this for the past 20 days and things look good. Going to rebase now. Can we then get this merged?

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
 	return true
 }
 
+// trace holds timings for a single HTTP roundtrip.
```
brian-brazil: roundTripTrace

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+
+func (t *transport) DNSStart(_ httptrace.DNSStartInfo) {
+	t.current.start = time.Now()
+}
+func (t *transport) DNSDone(_ httptrace.DNSDoneInfo) {
+	t.current.dnsDone = time.Now()
+}
+func (ts *transport) ConnectStart(_, _ string) {
+	t := ts.current
+	// No DNS resolution, e.g. connecting to IP directly.
```
brian-brazil: s/e.g.//

e.g. means "for example" and this is the only case where there'd be no DNS

prober/http.go
```diff
@@ -247,6 +333,42 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+		// We get the duration for the first request from chooseProtocol.
+		if i != 0 {
+			durationGaugeVec.WithLabelValues("resolve").Add(trace.dnsDone.Sub(trace.start).Seconds())
+		}
+		// Continue here if we never got a connection because a request failed.
+		if trace.gotConn.IsZero() {
+			continue
+		}
+		if trace.tls {
+			// dnsDone must be set if gotConn was set
```
brian-brazil: Comments end with a full stop

prober/http.go
```diff
@@ -65,9 +67,76 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+	t.current.gotConn = time.Now()
+}
+func (t *transport) GotFirstResponseByte() {
+	t.current.responseStart = time.Now()
+}
+
 func ProbeHTTP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var redirects int
 	var (
+		durationGaugeVec = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
brian-brazil: We should initialise this with all the potential phases.

prober/http.go
```diff
@@ -247,6 +333,42 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if resp == nil {
 		resp = &http.Response{}
 	}
+	i := 0
```
brian-brazil: You don't need a separate i now that tt.traces is a list, it's the first variable from the range.

prober/utils.go
```diff
@@ -63,5 +62,7 @@ func chooseProtocol(preferredIPProtocol string, target string, registry *prometh
 	}
 
 	level.Info(logger).Log("msg", "Resolved target address", "ip", ip)
-	return ip, nil
+	lookupTime := time.Since(resolveStart).Seconds()
+	probeDNSLookupTimeSeconds.Add(lookupTime)
```
brian-brazil: This is going to double count with the defer above. We should just have the defer, and can use a named return value for it.

prober/http.go
discordianfish: How?

prober/http.go
discordianfish: Good catch. Still a left over from earlier where this wasn't needed.

discordianfish:
Ok, made these changes except the initializing. Not sure what you mean by that one..?

prober/http.go
brian-brazil: Call WithLabelValues with all the potential values.

discordianfish: Like this?

prober/http.go
```diff
@@ -173,6 +173,11 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		})
 	)
 
+	// Initialize durationGaugeVec
```
brian-brazil: You don't really need this comment.

discordianfish:
Okay, removed the comment. Ready to merge now?

PR #218, opened 2017-09-04
Make per-probe logs more accessible and useful

brian-brazil:
@discordianfish 

There's more to be done here. Main thing of note user wise is that the application logs no longer contain errors/info when an individual probe fails, that's all debug now. However failure are now much easier to debug, as you can pass in `&debug=true` as a url parameter and see what's going wrong.


Conorbro:
Looks good to me 👍

PR #220, opened 2017-09-07
add tls-upgrade for tcp-prober (STARTTLS)

thz:
This change adds tls-upgrade feature for tcp-probing also known as STARTTLS. This allows probing protocols like SMTP, IMAP, POP3 (also for certificate expiry) with the tcp-prober by adding a starttls directive to the query_response configuration.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -101,10 +101,15 @@ The other placeholders are specified separately.
 [ preferred_ip_protocol: <string> | default = "ip6" ]
 
 # The query sent in the TCP probe and the expected associated response.
+# starttls upgrades TCP connection to TLS.
 query_response:
-  [ - [ [ expect: <string> ], [ send: <string> ] ], ... ]
+  [ - [ [ expect: <string> ],
+        [ send: <string> ],
+        [ starttls: <boolean | default = false> ]
+      ], ...
+  ]
 
-# Whether or not TLS is used.
+# Whether or not TLS is used when the connection is initiated.
 [ tls: <boolean | default = false> ]
 
 # Configuration for TLS protocol of TCP probe.
--- a/config/config.go
+++ b/config/config.go
@@ -76,8 +76,9 @@ type HTTPProbe struct {
 }
 
 type QueryResponse struct {
-	Expect string `yaml:"expect,omitempty"`
-	Send   string `yaml:"send,omitempty"`
+	Expect   string `yaml:"expect,omitempty"`
+	Send     string `yaml:"send,omitempty"`
+	StartTLS bool   `yaml:"starttls,omitempty"`
 
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline"`
--- a/prober/tcp.go
+++ b/prober/tcp.go
@@ -150,7 +150,6 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 				targetAddress, _, _ := net.SplitHostPort(target) // Had succeeded in dialTCP already.
 				tlsConfig.ServerName = targetAddress
 			}
-
 			tlsConn := tls.Client(conn, tlsConfig)
 			defer tlsConn.Close()
 
--- a/prober/tcp.go
+++ b/prober/tcp.go
@@ -138,6 +138,36 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 				return false
 			}
 		}
+		if qr.StartTLS {
+			// Upgrade TCP connection to TLS.
+			tlsConfig, err := pconfig.NewTLSConfig(&module.TCP.TLSConfig)
+			if err != nil {
+				level.Error(logger).Log("msg", "Failed to create TLS configuration", "err", err)
+				return false
+			}
+			if tlsConfig.ServerName == "" {
+				// Use target-hostname as default for TLS-servername.
+				targetAddress, _, _ := net.SplitHostPort(target) // Had succeeded in dialTCP already.
+				tlsConfig.ServerName = targetAddress
+			}
+
+			tlsConn := tls.Client(conn, tlsConfig)
+			defer tlsConn.Close()
+
+			// Initiate TLS handshake (required here to get TLS state).
+			if err := tlsConn.Handshake(); err != nil {
+				level.Error(logger).Log("msg", "TLS Handshake (client) failed", "err", err)
+				return false
+			}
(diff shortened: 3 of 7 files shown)


brian-brazil:
Would this work for other protocols that do non standard things with TLS, such as postgres?

thz:
@brian-brazil what postgres does is actually very similar but the "STARTTLS" is less readable. We would either need to extend the send/expect directives to support binary or create a custom psql_request_ssl directive (postgres expects some 0x04D2162F plus a length, and will confirm by sending a 0x53). I just gave it a try and it works with a new directive like this:

```     
          if qr.PSQLSSLREQ {
            buf := make([]byte, 8)
            binary.BigEndian.PutUint32(buf, 8)
            binary.BigEndian.PutUint32(buf[4:], 0x04D2162F)
            conn.Write(buf) // write SSLRequest message
            if n, err := conn.Read(buf); err != nil {
                log.Errorf("Error reading back psql SSLRequest reply: %v", err)
                return false 
            } else if n != 1 {
                log.Errorf("Error: unexpected reply size for psql SSLRequest: %d", n)
                return false 
            } else if buf[0] == 0x4e { // 'N' - ssl not supported
                log.Infof("psql: postgres target does not support ssl")
                return false 
            } else if buf[0] == 0x53 { // 'S' - ssl startup confirmed
                log.Info("psql: postgres target confirmed tls-upgrade")
            } else {
                log.Errorf("Error: unexpected reply psql SSLRequest: %x", buf[0])
                return false
            }
        }
```

This is very specific for a generic TCP prober, so I would prefer either a dedicated prober for psql or a send/receive which supports binary. With the latter it would look more like this in the config:
```
    tcp:
      query_response:
        - binarysend: 0x04d2162f
        - binaryexpect: 0x53
        - starttls: yes
```
That looks a lot better and should be doable.

RFC!

thz:
@brian-brazil so to answer your question: yes, it just starts the tls-upgrade. Making the protocol ready for it depends on the protocol and is out of this PR's scope.

brian-brazil:
What about the other ones beyond these two? I don't want to end up having to support tens of different custom non-standard ways of doing TLS, of which SMTP and postgres are only two. Binary query/response is another discussion again.

thz:
This upgrade mechanism in in no way a custom way of doing TLS.

The directive/command just upgrades a plain-TCP connection to a TLS connection by initiating the TLS handshake (technically by swapping the conn interface). This is useful for all protocols which start in plain mode and upgrade to encrypted communication later. Examples are SMTP, IMAP, POP3, IRC, LDAP (rfc4511 / 4.14.), FTP (rfc4217), XMPP (rfc6120 / 5.) , NNTP (rfc4642 / 2.1.). As far as I know even HTTP had such a mode. 

Actually it should even work with protocols which are encrypted from the beginning by simply putting the starttls-config-statement at the top of the query_response.
Perhaps "tlsupgrade" would be a better name for the config-directive to not let people confuse it with the protocol specific "STARTTLS" statement which _prepares_ for the TLS-upgrade.

thz:
@brian-brazil does that address your question/concern? Would you prefer "tlsupgrade" as the directive?

SuperQ:
While I think `tls_upgrade` is the "correct" term, the common name for it is `STARTTLS`, so most users are going to be aware of it named `starttls` or `start_tls` in the config,

thz:
@brian-brazil what is your opinion on this?

CONFIGURATION.md
```diff
@@ -101,10 +101,11 @@ The other placeholders are specified separately.
 [ preferred_ip_protocol: <string> | default = "ip6" ]
 
 # The query sent in the TCP probe and the expected associated response.
+# starttls upgrades TCP connection to TLS.
 query_response:
-  [ - [ [ expect: <string> ], [ send: <string> ] ], ... ]
+  [ - [ [ expect: <string> ], [ send: <string> ], [ starttls: yes ] ], ... ]
```
brian-brazil: This will need to be put on multiple lines now. The default is not yes, and this is a bool.

CONFIGURATION.md
```diff
@@ -101,10 +101,11 @@ The other placeholders are specified separately.
...
 
-# Whether or not TLS is used.
+# Whether or not TLS is used (from the beginning).
```
brian-brazil: "when the connection is initiated" would be more common I think

config/config.go
```diff
@@ -76,8 +76,9 @@ type HTTPProbe struct {
 }
 
 type QueryResponse struct {
-	Expect string `yaml:"expect,omitempty"`
-	Send   string `yaml:"send,omitempty"`
+	Expect   string `yaml:"expect,omitempty"`
+	Send     string `yaml:"send,omitempty"`
+	Starttls bool   `yaml:"starttls,omitempty"`
```
brian-brazil: StartTLS as your uppercase acronyms in Go

prober/tcp.go
```diff
@@ -131,6 +131,30 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 				return false
 			}
 		}
+		if qr.Starttls {
+			// TLS-upgrade
```
brian-brazil: Period at the end of comments, and they should start with a capital letter.

prober/tcp.go
```diff
@@ -131,6 +131,30 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
...
+				tlsConfig.ServerName = targetAddress
+			}
+
+			tlsConn := tls.Client(conn, tlsConfig)
+			defer tlsConn.Close()
+			conn = net.Conn(tlsConn)
+			scanner = bufio.NewScanner(conn)
+
+			// get certificate expiry
+			tlsConn.Handshake()
```
brian-brazil: You need to check the error here.

thz:
@brian-brazil, thanks for review,

-  addressed your comments,
-  added a test (including cert-expiry metric),
-  squashed

thz:
ping @brian-brazil 
Any more remarks?

thz:
friendly reminder @brian-brazil

prober/tcp_test.go
```diff
@@ -15,14 +15,19 @@ package prober
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
+	"io/ioutil"
 	"net"
+	"os"
 	"runtime"
+
```
brian-brazil: This blank line shouldn't be here

brian-brazil:
Thanks!

PR #227, opened 2017-09-14
Add last 100 probes to the front page, with logs

brian-brazil:
screenshot_2017-09-14_15-46-32


SuperQ:
Neat, would it be worth making the number of buffered probes a flag?

brian-brazil:
For v1 I'm just going to hardcode it. We'd need paging and the like to support high numbers of results, and I'm thinking that having a separate 100 last failures and merging them would cover most cases.

PR #231, opened 2017-09-19
[prober/tcp] get servername for TLS from target

### LINKED ISSUE #225: TCP+TLS tries to validate IP address instead of hostname
It seems plain TCP TLS mode doesn't correctly validate the hostname. It appears to be resolving the hostname then passing the IP down to the TLS layer for hostname verification, which is wrong.

This works:
`/probe?target=https://google.com&module=http_2xx`

But this doesn't:
`/probe?target=google.com:443&module=tls_connect`

`level=error msg="Error dialing TCP: x509: cannot validate certificate for 216.58.197.174 because it doesn't contain any IP SANs" source="tcp.go:78"`

This isn't the same as #224, as google.com serves a valid cert even without SNI. It isn't something weird about Google's cert either: the same thing happens with euskalencounter.org, which by default (without SNI) serves a valid Let's Encrypt cert for that hostname.

Version: 0.8.1
Config:
```
  http_2xx:
    prober: http
    timeout: 5s
    http:
      preferred_ip_protocol: ip4
      tls_config:
        insecure_skip_verify: false
  tls_connect:
    prober: tcp
    timeout: 5s
    tcp:
      preferred_ip_protocol: ip4
      tls: true
      tls_config:
        insecure_skip_verify: false

### PR DESCRIPTION
**Pull request by thz**
Because dialTCP manually resolves name to IP, the
actual tls.DialWithDialer call cannot deduce the name
from the target. This changes puts the "lost" name
into tlsConfig to fix certficate name verification.
Tests are also added which fail without and succeed
with this change.

This fixes #225

### DIFF
--- a/prober/tcp_test.go
+++ b/prober/tcp_test.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	pconfig "github.com/prometheus/common/config"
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: prober/tcp_test.go [hunk 1]
```diff
@@ -66,6 +67,113 @@ func TestTCPConnectionFails(t *testing.T) {
...
+	defer cancel()
+
+	// Create test certificates valid for 1 day.
+	certExpiry := time.Now().AddDate(0, 0, 1)
+	testcert_pem, testkey_pem := generateTestCertificate(certExpiry, false)
+
+	// CAFile must be passed via filesystem, use a tempfile.
+	tmpCaFile, err := ioutil.TempFile("", "cafile.pem")
+	if err != nil {
+		panic(fmt.Sprintf("Error creating CA tempfile: %s", err))
```
brian-brazil: You can t.Fatalf here

prober/tcp_test.go
```diff
@@ -66,6 +67,113 @@ func TestTCPConnectionFails(t *testing.T) {
...
+	// Test name-verification with name from target.
+	target := net.JoinHostPort("localhost", listenPort)
+	if !ProbeTCP(testCTX, target, module, registry, log.NewNopLogger()) {
+		t.Fatalf("TCP module failed, expected success.")
+	}
+	<-ch
+
+	registry = prometheus.NewRegistry()
+	go serverFunc()
+	// Test name-verification against name from tls_config
```
brian-brazil: Comments should end in a full stop.

prober/tcp_test.go
thz: I decided for panic() because it is not related to the specific test but indicates some lower-level problem. Same panic() went through your last review in `TestTCPConnectionQueryResponseStartTLS`.
For consistency we should either change both or none. I have no strong opinion here and will change both places to t.Fatalf.

thz:
@brian-brazil done.

brian-brazil:
Thanks!

PR #240, opened 2017-10-02
Advanced ICMP options

Gouthamve:
@RichiH wanted to add the payload size and DF-bit to be configurable in the ICMP prober. While I had fun implementing this, I would like to know if this is okay before I put in more effort.

cc/ @brian-brazil @SuperQ

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/blackbox_exporter/240)

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -168,6 +168,12 @@ validate_additional_rrs:
 # The preferred IP protocol of the ICMP probe (ip4, ip6).
 [ preferred_ip_protocol: <string> | default = "ip6" ]
 
+# Set the DF-bit in the IP-header. Only works with ip4 and on *nix systems.
+[ dont_fragment: <boolean> | default = false ]
+
+# The size of the payload.
+[ payload_size: <int> ]
+
 ```
 
 ### <tls_config>
--- a/config/config.go
+++ b/config/config.go
@@ -96,7 +96,7 @@ type TCPProbe struct {
 
 type ICMPProbe struct {
 	PreferredIPProtocol string `yaml:"preferred_ip_protocol,omitempty"` // Defaults to "ip6".
-
+	Payload             int    `yaml:"payload,omitempty"`
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline"`
 }
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
@@ -96,7 +97,8 @@ type TCPProbe struct {
 
 type ICMPProbe struct {
 	PreferredIPProtocol string `yaml:"preferred_ip_protocol,omitempty"` // Defaults to "ip6".
-	Payload             int    `yaml:"payload,omitempty"`
+	PayloadSize         int    `yaml:"payload_size,omitempty"`
+	DontFragment        bool   `yaml:"dont_fragment,omitempty"`
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline"`
 }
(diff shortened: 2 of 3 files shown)


config/config.go
```diff
@@ -96,7 +96,8 @@ type TCPProbe struct {
 
 type ICMPProbe struct {
 	PreferredIPProtocol string `yaml:"preferred_ip_protocol,omitempty"` // Defaults to "ip6".
-
+	Payload             int    `yaml:"payload,omitempty"`
```
brian-brazil: PayloadSize would make more sense

config/config.go
```diff
@@ -96,7 +96,8 @@ type TCPProbe struct {
...
+	DontFragment        bool   `yaml:"dontfragment,omitempty"`
```
brian-brazil: dont_fragment

prober/icmp.go
```diff
@@ -58,27 +54,27 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
 	defer socket.Close()
 
+	var data []byte
+	payload := module.ICMP.Payload
+	if payload != 0 {
+		data = make([]byte, payload)
+		copy(data, "Prometheus Blackbox Exporter")
+	} else {
+		data = []byte("Prometheus Blackbox Exporter")
+		payload = 1500 // The usual MTU size.
```
brian-brazil: Where is this used?

prober/icmp.go
```diff
@@ -136,3 +132,96 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 		}
 	}
 }
+
+func getICMPSocket(
```
brian-brazil: I don't think you're buying much by breaking this out, and it also makes it more difficult to review what has changed in the code

prober/icmp.go
```diff
@@ -136,3 +132,96 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
+
+	return
+}
+
+type dfConn struct {
+	c *ipv4.RawConn
+}
+
+func (c *dfConn) ReadFrom(b []byte) (int, net.Addr, error) {
+	h, p, _, err := c.c.ReadFrom(b)
```
brian-brazil: This won't work on Windows, which is a supported platform.

SuperQ:
Feature idea seems good to me.

prober/icmp.go
Gouthamve: How do you propose I move about this? Different files for nix and windows? Or error when reading config?

Gouthamve:
Updated to incorporate feedback and to error if `DontFragment` is set on windows. Don't have a windows machine to test though.

prober/icmp.go
brian-brazil: First see if this can be done with Windows, we should be aiming for feature parity

CONFIGURATION.md
```diff
@@ -168,6 +168,12 @@ validate_additional_rrs:
 # The preferred IP protocol of the ICMP probe (ip4, ip6).
 [ preferred_ip_protocol: <string> | default = "ip6" ]
 
+# Set the DF-Bit in the IP-header. Only works with ip4 and on *nix systems.
```
brian-brazil: bit

prober/icmp.go
```diff
@@ -62,23 +62,44 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
+
 	defer socket.Close()
 
 	var data []byte
-	if payload != 0 {
-		data = make([]byte, payload)
-		copy(data[:], "Prometheus Blackbox Exporter")
+	if module.ICMP.PayloadSize != 0 {
+		data = make([]byte, module.ICMP.PayloadSize)
+		copy(data, "Prometheus Blackbox Exporter")
```
brian-brazil: What if the payload size is smaller than this string?

Gouthamve: https://play.golang.org/p/qrDFrlW6He Then only the payload size will be sen't with truncated string.

prober/icmp.go
```diff
@@ -62,23 +62,52 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
+			socket = &dfConn{c: rc}
+		}
 	}
+
 	defer socket.Close()
 
+	var data []byte
+	if module.ICMP.PayloadSize != 0 {
+		data = make([]byte, module.ICMP.PayloadSize)
+		copy(data, "Prometheus Blackbox Exporter")
```
RichiH: Nitpick, but it might make sense to include the exporter version in here.

Issue #184, opened 2017-07-07
Implement UDP 


joker234:
According to [CONFIGURATION.md](https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md), there is currently no possibility to check UDP Ports. Sure, this Problem is a little bit harder to manage than TCP, because you have no defined followable stream. My use case is to check if Kerberos is running resp. available on UDP Port 88.

I would be happy if someone could implement this.

brian-brazil:
I think we could do a simple single query/response.

candlerb:
I think a proper Kerberos check involves making a proper Kerberos call.  Therefore you will probably be better off with a separate script which requests and validates a Kerberos ticket.  You can integrate this using one of the following:

* [nrpe_exporter](https://github.com/RobustPerception/nrpe_exporter)
* [exporter_exporter](https://github.com/QubitProducts/exporter_exporter)
* Run the script from cron, and write the results into a file for `node_exporter`'s [textfile-collector](https://github.com/prometheus/node_exporter#textfile-collector) to pick up

[This script](https://gist.github.com/candlerb/c0360b111e4b51923c04e2718676c295) is a quick hack to check Active Directory - both Kerberos and LDAP (the LDAP query is authenticated using the Kerberos ticket).  I currently use it with Nagios, but nrpe_exporter should be fine with it.  It could probably be better written to avoid temporary files.

As regards the other things you describe: again, to check dhcpd you should send a correctly-formatted dhcp request and validate the response.  DNS queries are already supported by blackbox_exporter.

jirib:
A openvpn prober would be great, see https://github.com/liquidat/nagios-icinga-openvpn/blob/master/bin/check_openvpn

avanichy25:
@brian-brazil  i am trying to configure the UDP probe any update on the ticket when this feature will be available by blackboxexporter?

candlerb:
@avanichy25, what is your use case?

Almost no UDP protocols are text-based. Therefore, generating a text UDP packet and doing a regex check on the response (like you can with say SMTP) wouldn't be useful.  Sending a canned binary UDP packet isn't very useful either, e.g. for any protocol which encodes the current time.

So I can't see a useful way to provide a "generic UDP check".  You need a specific check for the protocol you're interested in - e.g. RADIUS, DHCP, OpenVPN etc - just like the built-in DNS check.

hslabbert:
[...]

PR #242, opened 2017-10-05
Add UDP prober.

KalivarapuReshma:
For issue #184.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -117,6 +117,23 @@ tls_config:
   [ <tls_config> ]
 
 ```
+### <udp_probe>
+
+```yml
+
+# The preferred IP protocol of the TCP probe (ip4, ip6).
+[ preferred_ip_protocol: <string> | default = "ip6" ]
+
+# The query sent in the TCP probe and the expected associated response.
+# starttls upgrades TCP connection to TLS.
+query_response:
+  [ - [ [ expect: <string> ],
+        [ send: <string> ],
+        [ starttls: <boolean | default = false> ]
+      ], ...
+  ]
+
+```
 
 ### <dns_probe>
 
--- a/prober/udp.go
+++ b/prober/udp.go
@@ -1,0 +1,119 @@
+// Copyright 2016 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package prober
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"net"
+	"regexp"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/blackbox_exporter/config"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func dialUDP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (net.Conn, error) {
+	var dialProtocol, dialTarget string
+	dialer := &net.Dialer{}
+	targetAddress, port, err := net.SplitHostPort(target)
+	if err != nil {
+		level.Error(logger).Log("msg", "Error splitting target address and port", "err", err)
+		return nil, err
+	}
+
+	ip, err := chooseProtocol(module.UDP.PreferredIPProtocol, targetAddress, registry, logger)
+	if err != nil {
+		level.Error(logger).Log("msg", "Error resolving address", "err", err)
+		return nil, err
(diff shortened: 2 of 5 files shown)


CONFIGURATION.md
```diff
@@ -117,6 +117,23 @@ tls_config:
...
 
 ```
+### <udp_probe>
+
+```yml
+
+# The preferred IP protocol of the TCP probe (ip4, ip6).
+[ preferred_ip_protocol: <string> | default = "ip6" ]
+
+# The query sent in the TCP probe and the expected associated response.
```
thz: comments are talking about TCP.

CONFIGURATION.md
```diff
@@ -117,6 +117,23 @@ tls_config:
...
+# starttls upgrades TCP connection to TLS.
+query_response:
+  [ - [ [ expect: <string> ],
+        [ send: <string> ],
+        [ starttls: <boolean | default = false> ]
```
thz: starttls is not implemented (as advertised)

prober/udp.go
```diff
@@ -0,0 +1,119 @@
...
+	registry.MustRegister(probeFailedDueToRegex)
+	deadline, _ := ctx.Deadline()
+
+	conn, err := dialUDP(ctx, target, module, registry, logger)
+	if err != nil {
+		level.Error(logger).Log("msg", "Error dialing UDP", "err", err)
+		return false
+	}
+	defer conn.Close()
+	level.Info(logger).Log("msg", "Successfully dialed")
```
brian-brazil: Does this mean much with UDP to an IP address? I can see those not familiar with how socket APIs work for UDP getting confused by this log line and thinking it means the packet was sent.

prober/udp.go
```diff
@@ -0,0 +1,119 @@
...
+		level.Info(logger).Log("msg", "Processing query response entry", "entry_number", i)
+		send := qr.Send
+		if qr.Expect != "" {
+			re, err := regexp.Compile(qr.Expect)
+			if err != nil {
+				level.Error(logger).Log("msg", "Could not compile into regular expression", "regexp", qr.Expect, "err", err)
+				return false
+			}
+			var match []int
+			// Read lines until one of them matches the configured regexp.
```
brian-brazil: I'm not sure this makes sense for UDP. How many text-based UDP protocols are there?

prober/udp.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Set a deadline to prevent the following code from blocking forever.
+	// If a deadline cannot be set, better fail the probe by returning an error
+	// now rather than blocking forever.
+	if err := conn.SetDeadline(deadline); err != nil {
+		level.Error(logger).Log("msg", "Error setting deadline", "err", err)
+		return false
+	}
+
+	scanner := bufio.NewScanner(conn)
```
brian-brazil: What does this do for UDP?

prober/udp.go
erichorst: I have one I'm interested in using blackbox_exporter for. It makes as much sense for UDP as it does for TCP I think.

erichorst:
I have an interest in this pull request and will test the udp functionality sometime soon and offer some supporting feedback.

prober/udp.go
brian-brazil: I suspect that's a very small proportion of UDP use cases

ac427:
not related to this PR. I am going to ask in some go stack overflow. I found this issue while working on similar PR. The connection passes for all UDP ports ? 

```
$cat foo.go 
package main

import "net"
import "fmt"

func main() {

	conn, err := net.Dial("udp", "0.north-america.pool.ntp.org:80")
	if err != nil {
		fmt.Printf("Fail")
	}
	if conn != nil {
		fmt.Printf("Pass")
	}
}
$go run foo.go 
Pass$
$sudo nmap -sT -sU 0.north-america.pool.ntp.org
[sudo] password for ac: 

Starting Nmap 6.40 ( http://nmap.org ) at 2019-03-08 15:51 EST
Nmap scan report for 0.north-america.pool.ntp.org (216.229.4.69)
Host is up (0.038s latency).
Other addresses for 0.north-america.pool.ntp.org (not scanned): 45.56.118.161 45.33.48.4 45.79.1.70
rDNS record for 216.229.4.69: triton.ellipse.net
Not shown: 1000 filtered ports, 999 open|filtered ports
PORT    STATE SERVICE
123/udp open  ntp
```

brian-brazil:
You didn't send any packets.

ac427:
There is no open port 80 on 0.north-america.pool.ntp.org to send. Isn't it supposed to timeout with error when failed? or at least hang ( that is what it does for tcp ) 

```
$cat foo.go 
package main

import "net"
import "fmt"

func main() {

	conn, err := net.Dial("tcp", "google.com:123")
	if err != nil {
		fmt.Printf("Fail")
	}
	if conn != nil {
		fmt.Printf("Pass")
	}
}
$go run foo.go 
^Csignal: interrupt
```

thz:
@ac427 that is exactly the difference to UDP. You can send a packet and that is it. There is no connection to be established like in TCP. You could wait for a reply packet of course. But then you should start to define the expectation.

ac427:
@thz  , ACK :) .  I feel dump now. Long day!

brian-brazil:
This has gone stale, and it remains unclear how we'd make this work in principle due to the connectionless and usually binary nature of UDP.

PR #251, opened 2017-10-26
Add json path regex support

tanner-bruce:
Allows creating an HTTP prober that can select a JSON key via a selector, and match the value against regexs

```
modules:
  jsonmatch:
    prober: http
    http:
      json_match:
      - key: a.0.b
        fail_if_not_matches_regexp:
        - abc
        fail_if_matches_regexp:
        - def
      preferred_ip_protocol: ip4
```

and have your target be a JSON doc like the below, allowing for a nice UX when testing JSON endpoints.

```
{
  "a": [{
    "b":"abc"
  },{
    "foo":"bar"
  }]
}
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,15 @@ type Module struct {
 	XXX map[string]interface{} `yaml:",inline"`
 }
 
+type JSONMatch struct {
+	JsonPath               string   `yaml:"key"`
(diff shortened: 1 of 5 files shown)


brian-brazil:
The blackbox exporter is intended to be fairly dumb, and agnostic to exactly what it is probing. If you need to add a parser for specific output formats that indicates that you need to write a custom exporter and/or move this to whitebox monitoring.

PR #252, opened 2017-10-26
Overwrite Config per Module via URL Query Parameters

cimnine:
This change allows Prometheus to dynamically adjust the config of a module with URL query parameters. The adjusted config is only valid for the lifecycle of just the one call.

Our use-case for this is the ability to configure the regexes via labels in Kubernetes and, leveraging some relabelling magic, passing those from the Kubernetes service discovery and to the blackbox exporter.

Examples:
* [...&HTTP_FailIfMatchesRegexp=Prometheus](http://localhost:9115/probe?target=prometheus.io&module=http_2xx&HTTP_FailIfMatchesRegexp=Prometheus) → `probe_failed_due_to_regex 1`
* [...&HTTP_FailIfNotMatchesRegexp=Prometheus](http://localhost:9115/probe?target=prometheus.io&module=http_2xx&HTTP_FailIfNotMatchesRegexp=Prometheus) → `probe_failed_due_to_regex 0`
* [...&HTTP_FailIfMatchesRegexp=Schnitzel](http://localhost:9115/probe?target=prometheus.io&module=http_2xx&HTTP_FailIfMatchesRegexp=Schnitzel) → `probe_failed_due_to_regex 0`
* [...&HTTP_FailIfNotMatchesRegexp=Schnitzel](http://localhost:9115/probe?target=prometheus.io&module=http_2xx&HTTP_FailIfNotMatchesRegexp=Schnitzel) → `probe_failed_due_to_regex 1`

We just started to use this patch in our internal blackbox exporter package and we can't wait to hear the feedback of the community on this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: go
 
 go:
 - 1.8.x
+- 1.9.x
 - master
 
 go_import_path: github.com/prometheus/blackbox_exporter
--- a/README.md
+++ b/README.md
@@ -69,6 +69,31 @@ scrape_configs:
         replacement: 127.0.0.1:9115  # Blackbox exporter.
 ```
 
+### Overwrite Blackbox Exporter Configuration From Prometheus
+
+It's possible to overwrite each configuration that is made in the config file from Prometheus.
+
+Example config that overwrites the regex to check:
+```yml
+scrape_configs:
+  - job_name: 'blackbox'
+    metrics_path: /probe
+    params:
+      module: [http_2xx]  # Look for a HTTP 200 response.
+    static_configs:
+      - targets:
+        - https://prometheus.io
+    relabel_configs:
+      - source_labels: [__address__]
+        target_label: __param_target
+      - source_labels: [__param_target]
+        target_label: instance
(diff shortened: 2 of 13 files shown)


brian-brazil:
Thanks for your PR, this is explicitly not a feature that the blackbox exporter has. In Prometheus all configuration comes from config files for management/security reasons, and we don't want to make the blackbox exporter into an open proxy. If you need this, you should auto-generate the blackbox exporter config file.

cimnine:
Can you explain further on the security implication you claim? IMO this change does not make the blackbox-exporter more "unsecure". The blackbox exporter receives the targets to probe via URL query parameter anyway, which is the only place it's told where to connect to.

cimnine:
And what if I would add a parameter to the config file, something like `allow_overwrite_via_url`?

brian-brazil:
This would allow changing of URL parameters, http bodies, headers, tcp messages and reading of arbitrary files via bearer_token_file.

In Prometheus we try to have just one way to do things and let users build upon it, which usually means as a config file that you're free to integrate with your configuration management systems as you wish.

Issue #253, opened 2017-10-30
probe_dns_lookup_time_seconds wrong by orders of magnitude


PadaKwaak:
### Host operating system:
I am using prom/blackbox-exporter docker image on a CentOS 7.3 host:
`uname -a`: `Linux mydockerhost.local 3.10.0-514.26.2.el7.x86_64 #1 SMP Tue Jul 4 15:04:05 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
`cat /etc/centos-release`: `CentOS Linux release 7.3.1611 (Core)`
`docker --version`: `Docker version 17.07.0-ce, build 8784753`

### blackbox_exporter version:
I am using prom/blackbox-exporter:v0.10.0 docker image:
```
blackbox_exporter, version 0.10.0 (branch: HEAD, revision: 75681e3c4f051e24269a132e8ec19517118e1586)
  build user:       root@0334aae810a6
  build date:       20171009-14:25:49
  go version:       go1.9.1
```
The behaviour was the same with v0.7.0, v0.8.1, v0.9.0 & v0.9.1 docker images.

### What is the blackbox.yml module config.
```
modules:
  tcp_ipv4:
    prober: tcp
    tcp:
      preferred_ip_protocol: ip4
    timeout: 5s
```

### What is the prometheus.yml scrape config.
N/A, since I'm using `curl` from another host.

### What logging output did you get from adding `&debug=true` to the probe URL?
```
[root@chris ~]#  curl 'http://172.30.18.21:9115/probe?target=www.google.co.za:443&module=tcp_ipv4&debug=true'
Logs for the probe:
ts=2017-10-30T10:06:40.969596597Z caller=main.go:116 module=tcp_ipv4 target=www.google.co.za:443 level=info msg="Beginning probe" probe=tcp timeout_seconds=4.5
ts=2017-10-30T10:06:40.969686476Z caller=utils.go:43 module=tcp_ipv4 target=www.google.co.za:443 level=info msg="Resolving target address" preferred_ip_protocol=ip4
ts=2017-10-30T10:06:41.188253005Z caller=utils.go:63 module=tcp_ipv4 target=www.google.co.za:443 level=info msg="Resolved target address" ip=216.58.223.35
ts=2017-10-30T10:06:41.188298125Z caller=tcp.go:55 module=tcp_ipv4 target=www.google.co.za:443 level=info msg="Dialing TCP without TLS"
[...]

PR #254, opened 2017-10-30
Wrap DNS lookup metric time defer

SuperQ:
Wrap the DNS lookup time `defer` in an anonymous function to avoid arguments being evaluated early.

Closes: https://github.com/prometheus/blackbox_exporter/issues/253

### DIFF
--- a/prober/utils.go
+++ b/prober/utils.go
@@ -43,7 +43,9 @@ func chooseProtocol(preferredIPProtocol string, target string, registry *prometh
 	level.Info(logger).Log("msg", "Resolving target address", "preferred_ip_protocol", preferredIPProtocol)
 	resolveStart := time.Now()
 
-	defer probeDNSLookupTimeSeconds.Add(time.Since(resolveStart).Seconds())
+	defer func() {
+		probeDNSLookupTimeSeconds.Add(time.Since(resolveStart).Seconds())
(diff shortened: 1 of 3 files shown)


prober/utils.go
```diff
@@ -43,8 +43,6 @@ func chooseProtocol(preferredIPProtocol string, target string, registry *prometh
 	level.Info(logger).Log("msg", "Resolving target address", "preferred_ip_protocol", preferredIPProtocol)
 	resolveStart := time.Now()
 
-	defer probeDNSLookupTimeSeconds.Add(time.Since(resolveStart).Seconds())
```
brian-brazil: What you want to do is make this an anonymous function, so we still get the right value if the below return is hit.

SuperQ: There only seems to be one valid return, the other is an error return.  Is it valid to return a DNS lookup time metric if the lookup fails?

brian-brazil: That's what we do elsewhere, the dns lookup still took time.

SuperQ: Ok, I will implement this tomorrow.

SuperQ:
Updated with anon func, tested, works as expected.

brian-brazil:
Unittests are failing on Travis.

PR #186, opened 2017-07-10
support redirects that require cookies

dabear:
We have a case where we're running blackbox_exporter on a linux host, and we want to monitor a Windows machine running IIS web server with the DotNetNuke (DNN) framework.

If you hit the website root path for the first time, you will get a 302 Moved Error Message together with a cookie and a redirect back to it self. If you follow that redirect with the cookie, you will get a 200 OK.

Blackbox_exporter would previously not allow cookies, and therefore DNN would always return 302. With this fix, blackbox_exporter will follow the redirect with a cookiejar and the DNN site would return a 200 OK message.

This can also be seen with curl by toggling the use of cookies or not:
```
bjorninge@bjorninge-lenovo:~/code/blackbox_exporter$ curl  https://mydnnsite.example.com/
<html><head><title>Object moved</title></head><body>
<h2>Object moved to <a href="https://mydnnsite.example.com//">here</a>.</h2>
</body></html>

bjorninge@bjorninge-lenovo:~/code/blackbox_exporter$ curl -I https://mydnnsite.example.com/
HTTP/1.1 302 Found
Content-Length: 153
Location: https://mydnnsite.example.com/
X-Redirect-Reason: Requested SplashPage Requested
Set-Cookie: dnn_IsMobile=False; path=/; HttpOnly
Set-Cookie: SplashPageView=true; path=/; HttpOnly
Date: Mon, 10 Jul 2017 13:24:18 GMT


```
By specifying cookies you will get a 200 OK response

```
bjorninge@bjorninge-lenovo:~/code/blackbox_exporter$ curl -I -L  -c cookies.txt  https://mydnnsite.example.com/
HTTP/1.1 302 Found
Content-Length: 153
Location: https://mydnnsite.example.com/
X-Redirect-Reason: Requested SplashPage Requested
Set-Cookie: dnn_IsMobile=False; path=/; HttpOnly
Set-Cookie: SplashPageView=true; path=/; HttpOnly
Date: Mon, 10 Jul 2017 13:25:30 GMT

HTTP/1.1 200 OK
Cache-Control: no-cache
Pragma: no-cache
Content-Length: 9326
Content-Type: text/html; charset=utf-8
Expires: -1
Set-Cookie: dnn_IsMobile=False; path=/; HttpOnly
[...]

PR #257, opened 2017-11-07
Support redirects that require cookies plus unit test

Conorbro:
Re-added code introduced in #186  as well as the missing unit test.


prober/http.go
```diff
@@ -21,13 +21,16 @@ import (
...
+	"net/http/cookiejar"
 	"net/http/httptrace"
 	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
+	"golang.org/x/net/publicsuffix"
+
```
brian-brazil: No need for this blank line, we want 3rd party libraries together

prober/http.go
```diff
@@ -221,6 +224,14 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		return false
 	}
 
+	jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
+	if err != nil {
+		level.Error(logger).Log("msg", "Error generating cookiejar", "err", err)
+		return false
+	}
+	client.Jar = jar
+	client.Timeout = module.Timeout
```
brian-brazil: Where is this change coming from?

Conorbro: Whoops, I mistakenly added that line about timeouts!

PR #262, opened 2017-11-16
allow specification of source address for probes (icmp, tcp, dns)

thz:
For IPv4 raw socket is used and the source IP address is set in the IP header.
For IPv6 a control message (ancillary data / RFC3542) is passed for specifying
the address.


thz:
Addresses the ICMP part of #3.

thz:
Hmm. Check failure seems to be unrelated: `*** Test killed with quit: ran too long (10m0s).` while TCP testing.

CONFIGURATION.md
```diff
@@ -174,6 +174,9 @@ validate_additional_rrs:
 # The size of the payload.
 [ payload_size: <int> ]
 
+# The source IP address.
+[ source_address: <string> ]
```
brian-brazil: Would it be possible to support the interface name too? If there's DHCP involved you may not know the IP.

config/config.go
```diff
@@ -99,6 +99,7 @@ type ICMPProbe struct {
 	PreferredIPProtocol string `yaml:"preferred_ip_protocol,omitempty"` // Defaults to "ip6".
 	PayloadSize         int    `yaml:"payload_size,omitempty"`
 	DontFragment        bool   `yaml:"dont_fragment,omitempty"`
+	SourceAddress       string `yaml:"source_address,omitempty"`
```
brian-brazil: This should apply to all probes

CONFIGURATION.md
dswarbrick: Should be possible. Resolve the interface name to interface index first from `net.InterfaceByName()`, then set `IfIndex` in the `ipv6.ControlMessage` struct.

thz: @brian-brazil how about a source_iface attribute in another PR?

config/config.go
thz: @brian-brazil Okay... Added similar code for tcp, http, dns and moved the config to the module.

thz:
Now also addresses the TCP part of #3.

example.yml
```diff
@@ -82,6 +82,7 @@ modules:
         - expect: "^:[^ ]+ 001"
   icmp_example:
     prober: icmp
+    source_address: "127.0.0.1"
```
brian-brazil: it's odd that this is in the probe but preferred_ip_protocol is in the module

prober/http.go
```diff
@@ -217,7 +217,20 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
 		httpClientConfig.TLSConfig.ServerName = targetHost
 	}
-	client, err := pconfig.NewHTTPClientFromConfig(&httpClientConfig)
+
+	// Target protocol is used as hint for local address protocol.
+	sourceNetwork := "ip4"
+	if ip.IP.To4() == nil {
+		sourceNetwork = "ip6"
+	}
+	// Resolve local bind address
```
brian-brazil: Comments should end with a full stop

prober/http.go
```diff
@@ -217,7 +217,20 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+	saddr, err := net.ResolveIPAddr(sourceNetwork, module.SourceAddress)
```
brian-brazil: What if there is no source address configured?

prober/http.go
```diff
@@ -403,3 +416,52 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+// NewHTTPClientWithLocalAddr wraps NewHTTPClientFromConfig from
+// github.com/prometheus/common/config/http_config.go and
+// adds the ability to specify a local address to bind the client.
+// The difference of this wrapper to the prometheus config
+// version is the specification of a DialContext for the
+// http.RoundTripper. This is currently not possible at a later
+// stage because of many possible RoundTripper wrappings.
+// Adding this to the prometheus config is an alternative
+// to this wrapper.
+func NewHTTPClientWithLocalAddr(cfg *pconfig.HTTPClientConfig, laddr *net.TCPAddr) (*http.Client, error) {
```
brian-brazil: This is duplicating code with common, it'd be better to extend it. Though that code is just about to get redone.

prober/tcp.go
```diff
@@ -46,9 +46,22 @@ func dialTCP(ctx context.Context, target string, module config.Module, registry
...
 	if ip.IP.To4() == nil {
 		dialProtocol = "tcp6"
+		sourceNetwork = "ip6"
 	} else {
 		dialProtocol = "tcp4"
+		sourceNetwork = "ip4"
 	}
+
+	if len(module.SourceAddress) > 0 {
+		saddr, err := net.ResolveIPAddr(sourceNetwork, module.SourceAddress)
```
brian-brazil: Why do you resolve this here and not elsewhere? I don't think there's any need to resolve it.

example.yml
thz: True. But your "This should apply to all probes" made me agree and also implement source-address specification for the other probers. Is it possible (from a configuration compatibility perspective) to pull preferred_ip_protocol up to this level as well?

prober/tcp.go
thz: When module.SourceAddress is an IP-Address already no resolving will happen. In that case net.ResolveIPAddr will just convert to an net.IPAddr. I did not want to prevent names being used as module.SourceAddress so resolving would be necessary.
As to "why here not elsewhere". dialTcp() is the place where the target is resolved as well (chooseProtocol) and resolving in the dial function seems to be aligned with golang dialers in general. Which place do you think would be better suited for resolving?

prober/http.go
thz: Condition is indeed missing here. Will update soon.

prober/http.go
thz: So how should we proceed? Skipping source-address config for http prober and contribute to that common code redoings? Http prober can be made source-address-capable once that work is done.

example.yml
brian-brazil: That'd be a breaking change, so best to keep it consistent.

prober/http.go
brian-brazil: Waiting for that work to complete is probably best.

prober/tcp.go
brian-brazil: I'm saying that in some probes you support dns names and in others you don't. I'm suggesting not supporting it anywhere.

thz: Ahh! Actually resolving is in for DNS, TCP and HTTP. Only ICMP (initial PR) is using net.ParseIp.
The resolving also solves the question of using IPv4 or IPv6 which also depends on the target (You need a local v6 address for a v6 target but a v4 address for a v4 taget).
A DNS name could provide v4 and v6 address. Going with IP-addresses only in configuration would require
us to provide both (v4/v6). That is why I concluded that using a resolve-step to be a good idea.
When only an IP address is specified in config there is no resolving happening but only IP parsing.

example.yml
thz: So you propose to move it back into the prober-specific configuration?

brian-brazil: Yes

prober/tcp.go
brian-brazil: Relying on DNS to handle v4 vs v6 is a bit odd to me. I'd expect the user to provide only the address type that they need. Users requiring this are going to have a good idea of how networking works.

thz:
@brian-brazil please have a look. 
Source-IP-address resolving is removed. HTTP probe changes reverted (delayed for common code restructure and support).
I have kept the commits for review, and would squash after feedback.

prober/dns.go
```diff
@@ -137,6 +137,22 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 
 	client := new(dns.Client)
 	client.Net = dialProtocol
+
+	// Use configured SourceIPAddress.
+	if len(module.DNS.SourceIPAddress) > 0 {
+		if srcip := net.ParseIP(module.DNS.SourceIPAddress); srcip == nil {
+			level.Warn(logger).Log("msg", "Error parsing source ip address", "srcip", module.DNS.SourceIPAddress)
```
brian-brazil: This should fail the probe

prober/dns.go
```diff
@@ -137,6 +137,22 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 
 	client := new(dns.Client)
 	client.Net = dialProtocol
+
+	// Use configured SourceIPAddress.
+	if len(module.DNS.SourceIPAddress) > 0 {
+		if srcip := net.ParseIP(module.DNS.SourceIPAddress); srcip == nil {
```
brian-brazil: srcIP would be clearer

prober/icmp.go
```diff
@@ -122,9 +127,22 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 		return
 	}
 	level.Info(logger).Log("msg", "Writing out packet")
-	if _, err = socket.WriteTo(wb, ip); err != nil {
-		level.Warn(logger).Log("msg", "Error writing to socket", "err", err)
-		return
+	if socket6 != nil && len(module.ICMP.SourceIPAddress) > 0 {
+		// Set source address for IPv6.
```
brian-brazil: Why are you handling this in a different place than for v4?

thz: Checking for len() > 0 and parse errors is now in common code. 
The actual control message is v6 specific.

prober/dns.go
thz: Changed to Error and failure in all probes.

prober/dns.go
thz: Ack.

thz:
@brian-brazil, last review feedback is addressed. Commits are squashed. Please have a look.

prober/icmp.go
```diff
@@ -58,40 +59,44 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
-		} else {
-			s, err := net.ListenPacket("ip4:icmp", "0.0.0.0")
-			if err != nil {
-				level.Error(logger).Log("msg", "Error listening to socket", "err", err)
-				return
-			}
+		s, err := net.ListenPacket("ip4:icmp", "0.0.0.0")
+		if err != nil {
+			level.Error(logger).Log("msg", "Error listening to socket:", "err", err)
+			return
```
brian-brazil: Explicitly returning false would be clearer.

thz: I agree. That is not my code. I just moved it. I am fine with doing a cleanup commit but would separate it from this PR.

prober/icmp.go
```diff
@@ -58,40 +59,44 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 		return false
 	}
 
+	var srcIP net.IP
+	if len(module.ICMP.SourceIPAddress) > 0 {
+		if srcIP = net.ParseIP(module.ICMP.SourceIPAddress); srcIP == nil {
+			level.Error(logger).Log("msg", "Error parsing source ip address", "srcIP", module.ICMP.SourceIPAddress)
+			return false
+		}
+		level.Info(logger).Log("msg", "Using source address:", "srcIP", srcIP)
```
brian-brazil: There shouldn't be colons in the msg

thz:
@brian-brazil `:` removed.

brian-brazil:
Are you sure you pushed your change? I'm still seeing 2 of those.

thz:
@brian-brazil removed two more `:` and pushed.

brian-brazil:
Thanks!

bobrik:
I opened #765 for `http` probe support.

PR #263, opened 2017-11-17
Addresses a bug where the Content-Length header is empty

bigkraig:
We noticed that when testing some of our applications that are sensitive to the content-length header the test was failing. This duplicates some of the logic from http.NewRequest with the exception of the GetBody() func, which didn't appear to be necessary.

https://golang.org/src/net/http/request.go?s=25171:25240#L755

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -273,7 +273,9 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 
 	// If a body is configured, add it to the request.
 	if httpConfig.Body != "" {
-		request.Body = ioutil.NopCloser(strings.NewReader(httpConfig.Body))
+		body := strings.NewReader(httpConfig.Body)
+		request.Body = ioutil.NopCloser(body)
+		request.ContentLength = int64(body.Len())
 	}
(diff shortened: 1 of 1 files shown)


brian-brazil:
What is the bug exactly? This sounds like an issue with your applications not correctly implementing HTTP.

bigkraig:
@brian-brazil most of my information is coming from https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4.

As it is, when using an HTTP POST the only header provided is the `Host` header. I can configure `Content-Type` headers, and some others, but I can not add `Content-Length` or `Transfer-Encoding`. The test is running through an Akamai reverse proxy and it's throwing back a 411 error.

```For compatibility with HTTP/1.0 applications, HTTP/1.1 requests containing a message-body MUST include a valid Content-Length header field unless the server is known to be HTTP/1.1 compliant. If a request contains a message-body and a Content-Length is not given, the server SHOULD respond with 400 (bad request) if it cannot determine the length of the message, or with 411 (length required) if it wishes to insist on receiving a valid Content-Length.```

megaqube:
Would be so cool if we can get this. The content-length headers cannot be set via the config.

brian-brazil:
I think it'd be better to change how we call NewRequest, rather than patching this later on.

PR #265, opened 2017-11-20
Setting httpConfig.Body at NewRequest to get Content-Length set impli…

megaqube:
Setting the post body in the NewRequest so the Content-Length will be available during the request. Alternative approach to solving: https://github.com/prometheus/blackbox_exporter/pull/263

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -256,8 +256,9 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		targetURL.Host = net.JoinHostPort(ip.String(), targetPort)
 	}
 
-	var body io.Reader = nil
-	//If a body is configured, add it to the request.
+	var body io.Reader
+
+	// If a body is configured, add it to the request.
 	if httpConfig.Body != "" {
 		body = strings.NewReader(httpConfig.Body)
 	}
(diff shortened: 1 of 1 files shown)


prober/http.go
```diff
@@ -255,7 +255,14 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	} else {
 		targetURL.Host = net.JoinHostPort(ip.String(), targetPort)
 	}
-	request, err := http.NewRequest(httpConfig.Method, targetURL.String(), nil)
+
+	var body io.Reader = nil
+	//If a body is configured, add it to the request.
```
brian-brazil: Space after //

prober/http.go
```diff
@@ -255,7 +255,14 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	} else {
 		targetURL.Host = net.JoinHostPort(ip.String(), targetPort)
 	}
-	request, err := http.NewRequest(httpConfig.Method, targetURL.String(), nil)
+
+	var body io.Reader = nil
```
brian-brazil: The zero object is the default, no need to specify the nil here

megaqube:
Updated based on feedback. Thanks!

brian-brazil:
Thanks!

PR #266, opened 2017-11-21
Use net/http, time package constants

dswarbrick:
Note that one out of the three modified getEarliestCertExpiry() lines was doing integer division anyway, and was thus futile. The other two were at least converting the UnixNano() to float64 before dividing it by 1e9, but I'm not sure if anyone _really_ cares about nanoseconds on a TLS cert expiry. However, if they do, I'd then still recommend using a `float64(time.Second)` constant for pedantry's sake.


brian-brazil:
Thanks!

PR #276, opened 2017-12-18
Mention the Go version in the README

arianvp:
Ubuntu 16.04 ships with go `1.6.2` which does not have the `"context"` package

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,10 @@ HTTP, HTTPS, DNS, TCP and ICMP.
 ## Building and running
 
 ### Local Build
-First of all, you need at least Go version `1.7`
+Building from source
+
+To build Prometheus from the source code yourself you need to have a working
+Go environment with [version 1.9 or greater installed](http://golang.org/doc/install).
 
     make
     ./blackbox_exporter <flags>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,7 @@ HTTP, HTTPS, DNS, TCP and ICMP.
 ## Building and running
 
 ### Local Build
+First of all, you need at least Go version `1.7`
```
brian-brazil: I'd suggest using the same text in Prometheus itself's readme

README.md
```diff
@@ -10,7 +10,10 @@ HTTP, HTTPS, DNS, TCP and ICMP.
 ## Building and running
 
 ### Local Build
-First of all, you need at least Go version `1.7`
+Building from source
+
+To build Prometheus from the source code yourself you need to have a working
```
brian-brazil: s/Prometheus/the blackbox exporter/

sofianinho:
Hi @brian-brazil would something like my recent PR be useful for this case ?

brian-brazil:
This is stale, and we have published builds.

PR #277, opened 2017-12-20
Update CONFIGURATION.md

ntdt:
Should read *at least* not *at most*


brian-brazil:
You are meant to specify exactly one, so neither are quite correct. I'd say one of these must be specified.

brian-brazil:
Each module can only do one check, but each blackbox exporter can have multiple modules configured.

Issue #278, opened 2017-12-22
HTTP client configuration isn't validated as expected


simonpasquier:
### Host operating system: output of `uname -a`

Ubuntu 16.04

### blackbox_exporter version: output of `blackbox_exporter -version`

blackbox_exporter, version 0.11.0 (branch: master, revision: 200181ba84e595570eeed8cb1848491d3d4f85f0)
  build user:       pasquier-s@simon-trusty
  build date:       20171219-09:41:05
  go version:       go1.9.2

### What is the blackbox.yml module config.

```
modules:
  http_post_2xx:
    prober: http
    timeout: 5s
    http:
      method: POST
      bearer_token: foo
      bearer_token_file: foo
      basic_auth:
        username: "username"
        password: "mysecret"
```

### What is the prometheus.yml scrape config.

N/A

### What logging output did you get from adding `&debug=true` to the probe URL?

N/A

### What did you do that produced an error?

N/A

### What did you expect to see?

I expected that the blackbox exporter would reject the configuration because the module contains both `bearer_token` and `bearer_token_file`.

### What did you see instead?

The blackbox exporter didn't complain.

I've reproduced the problem with a unit test (see https://github.com/simonpasquier/blackbox_exporter/commit/17fb97295e808aa46d95b870b9c45c9b05ca08c1). The fix in the code of the blackbox exporter itself is fairly contained (see https://github.com/simonpasquier/blackbox_exporter/commit/1e9bed841ebc81ad20766be17a39dc80d2d55d5b) but it has implications on the prometheus/common/config package too (see https://github.com/simonpasquier/blackbox_exporter/commit/1e9bed841ebc81ad20766be17a39dc80d2d55d5b#diff-fbe65e96815b1559a386cd8521b5c79e) which are less trivial.

brian-brazil:
This should be fixed up in common, and then we can vendor it in.

simonpasquier:
Yep that was my understanding. So is it ok from your POV to remove the `checkOverflow(XXX)` calls from common? My gut feeling said yes since `HTTPClientConfig` should always be nested in another struct but I wanted a confirmation :-)

brian-brazil:
They should stay there, otherwise we won't detect fields that shouldn't be there (usually typos).

simonpasquier:
[...]

PR #284, opened 2018-01-12
Fix missing validation of HTTP client configuration

simonpasquier:
Fix for #278

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -166,6 +166,9 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := unmarshal((*plain)(s)); err != nil {
 		return err
 	}
+	if err := s.HTTPClientConfig.Validate(); err != nil {
+		return err
+	}
 	if err := checkOverflow(s.XXX, "http probe"); err != nil {
 		return err
 	}
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -31,12 +31,20 @@ func TestLoadBadConfigs(t *testing.T) {
 			ExpectedError: "Error parsing config file: unknown fields in dns probe: invalid_extra_field",
 		},
 		{
+			ConfigFile:    "testdata/blackbox-bad2.yml",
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -59,18 +59,21 @@ type Module struct {
...
+	FailIfSSL              bool              `yaml:"fail_if_ssl,omitempty"`
+	FailIfNotSSL           bool              `yaml:"fail_if_not_ssl,omitempty"`
+	Method                 string            `yaml:"method,omitempty"`
+	Headers                map[string]string `yaml:"headers,omitempty"`
+	FailIfMatchesRegexp    []string          `yaml:"fail_if_matches_regexp,omitempty"`
+	FailIfNotMatchesRegexp []string          `yaml:"fail_if_not_matches_regexp,omitempty"`
+	Body                   string            `yaml:"body,omitempty"`
+
+	// This can't be an embedded struct because the parser doesn't support it.
+	// See https://github.com/go-yaml/yaml/issues/125
```
brian-brazil: I don't see how that bug means you have to change this, the opposite in fact.

config/config.go
```diff
@@ -166,6 +169,13 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := unmarshal((*plain)(s)); err != nil {
 		return err
 	}
+	// The UnmarsharlYAML method of HTTPClientConfig isn't called explicitly
+	// because it isn't a pointer. and it can't be made a pointer otherwise the
```
brian-brazil: It is a pointer.

brian-brazil: You should clarify what "it" is

simonpasquier: What about `s.HTTPClientconfig isn't a pointer so its its UnmarsharlYAML method isn't called by the parser. And it can't be made a pointer otherwise ...`?

brian-brazil: It can be made a pointer, so I'd remove that bit.

simonpasquier: Yeah even simpler 🙂

brian-brazil:
Thanks!

PR #287, opened 2018-01-20
Fix HTTP content length metric name

juliusv:
This got broken as part of introducing the Prometheus client library for
metrics exposure in 7abdf0570d44cc8dbd258987e9d3353af1d6126e.


Conorbro:
Looks good to me. The failing tests look unrelated as far as I can tell.

PR #289, opened 2018-01-26
Add example with custom CA

mprasil:
Just a small addition to example configuration. I had to find that one searching issues.


brian-brazil:
Thanks!

PR #299, opened 2018-02-28
docs: trim down relabel_configs in readme

adamdecaf:
I only needed one of the rewrites and adding the others actually
caused probes to fail.

They're working with only this one `relabel_config`. I'm on `v0.11.0`.

screen shot 2018-02-28 at 9 56 58 am
screen shot 2018-02-28 at 9 57 07 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Additionally, an [example configuration](example.yml) is also available.
 HTTP, HTTPS (via the `http` prober), DNS, TCP socket and ICMP (see permissions section) are currently supported.
 Additional modules can be defined to meet your needs.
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
This relabelling is not correct, you need all 3.

adamdecaf:
I wasn't getting any `probe_*` metrics with all three, which I assume was from `replacement: 127.0.0.1:9115`. How am I to know that address if prometheus is in a k8s cluster?

brian-brazil:
That's a usage question, and you need to arrange for there to be a blackbox exporter you know the dns name or IP of.

It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

PR #308, opened 2018-04-02
Make the history limit a configurable flag

sgmitchell:
This change replaces the hardcoded limit on the number of elements stored in the log history with a configurable value that can be set as a runtime argument. The the old value of 100 elements will be use by default either when the argument is left off or the argument is out of range (less than 1).

The driving use case for this is when there are over 100 endpoints queried, some of them will be removed almost immediately from the logs.

### DIFF
--- a/history.go
+++ b/history.go
@@ -26,9 +26,10 @@ type result struct {
 }
 
 type resultHistory struct {
-	mu      sync.Mutex
-	nextId  int64
-	results []*result
+	mu         sync.Mutex
+	nextId     int64
+	results    []*result
+	maxResults uint
 }
 
 // Add a result to the history.
@@ -46,7 +47,7 @@ func (rh *resultHistory) Add(moduleName, target, debugOutput string, success boo
 	rh.nextId++
 
 	rh.results = append(rh.results, r)
-	if len(rh.results) > 100 {
+	if uint(len(rh.results)) > rh.maxResults {
 		results := make([]*result, len(rh.results)-1)
 		copy(results, rh.results[1:])
 		rh.results = results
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -50,6 +50,7 @@ var (
 	listenAddress = kingpin.Flag("web.listen-address", "The address to listen on for HTTP requests.").Default(":9115").String()
 	timeoutOffset = kingpin.Flag("timeout-offset", "Offset to subtract from timeout in seconds.").Default("0.5").Float64()
 	configCheck   = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
+  historyLimit  = kingpin.Flag("historyLimit", "The maximum amount of items to keep in the history.").Default("100").Int()
```
brian-brazil: Why not uint? This also needs a gofmt.

sgmitchell: I went with int because I thought it looked cleaner to avoid the having a cast on history.go:54 and negative values will be set to 100 by history.go:40. I have no problem changing it though.

Formatting issue was because I tried to use the github merge ui. Will fix with the uint commit.

history.go
```diff
@@ -26,16 +26,21 @@ type result struct {
...
+	results    []*result
+	maxResults uint
 }
 
 // Add a result to the history.
 func (rh *resultHistory) Add(moduleName, target, debugOutput string, success bool) {
 	rh.mu.Lock()
 	defer rh.mu.Unlock()
 
+	if rh.maxResults < 1 {
```
brian-brazil: You don't need this now.

main.go
```diff
@@ -50,6 +50,7 @@ var (
 	listenAddress = kingpin.Flag("web.listen-address", "The address to listen on for HTTP requests.").Default(":9115").String()
 	timeoutOffset = kingpin.Flag("timeout-offset", "Offset to subtract from timeout in seconds.").Default("0.5").Float64()
 	configCheck   = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
+	historyLimit  = kingpin.Flag("historyLimit", "The maximum amount of items to keep in the history.").Default("100").Uint()
```
brian-brazil: history.limit

sgmitchell:
@brian-brazil Any other changes you want to see?

brian-brazil:
Thanks!

PR #309, opened 2018-04-09
Update docker image path

Kosmas:


brian-brazil:
Why do you want to change this?

Kosmas:
Unable to find image 'blackbox_exporter:latest' locally
docker: Error response from daemon: pull access denied for blackbox_exporter, repository does not exist or may require 'docker login'.

but after the change:

Unable to find image 'quay.io/prometheus/blackbox-exporter:latest' locally
latest: Pulling from prometheus/blackbox-exporter

discordianfish:
I think the issue here is that the README only talks about building your own image, not how to use the official image on quay.io.

PR #310, opened 2018-04-12
Implement strict yaml unmarshalling.

manosf:
As a followup from prometheus/prometheus#4033 and prometheus/common/pull/127, I added strict yaml unmarshalling by updating the yaml vendor package and removing unnecessary checks and data structures.
I avoided code refactoring on the yaml Unmarshaler implementations inside `config/config.go` for this PR, in order to keep it simple. Please let me know if I should change it.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"runtime"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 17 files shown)


brian-brazil:
That's fine, could you fix the tests?

manosf:
Yes, this is what I'm trying to do right now, even though I have found a really weird problem with the tests. I'll update everything as soon as it's ready, and probably I'll squash all the commits to satisfy the DCO check.

brian-brazil:
Thanks!

PR #321, opened 2018-05-17
Update dependencies

discordianfish:
I've updated the dependencies which required changing NewHTTPClientFromConfig to NewClientFromConfig which introduced a new arg for go-conntrack. Not sure what to put there but it doesn't seem this gets exposed..? Anyway, let me know if you want me to use something else.


brian-brazil:
The name of the probe would be a good choice, it's used in some metrics.

The newer version however enables keepaloves for http, which aren't
desirable for the blackbox exporter so that needs solving first.

discordianfish:
@brian-brazil I've changed it to use module.Prober but I guess you mean we should put the module name in there (e.g the key in the config for the prober). Should I change ProbeFn to include the name?

brian-brazil:
The prober is fine, which will always be http.

These metrics are largely to help pinpoint fd leaks in service discovery code, so not too important for this exporter which only uses http in one way.

brian-brazil:
This is obsolete.

PR #329, opened 2018-06-01
For ICMP, only use RawConn for the Don't Fagment case

avonwyss:
Changed code to unify the IPv6 and IPv4 ICMP usage, and also how the source IP is used (listen on the source IP instead of defining it on the packet). Tested on Windows for both IPv4 and IPv6, with and without explicit source IP.

Not tested DF flag and *nix in general.

### DIFF
--- a/prober/icmp.go
+++ b/prober/icmp.go
@@ -46,7 +46,6 @@ func getICMPSequence() uint16 {
 func ProbeICMP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var (
 		socket      net.PacketConn
-		socket6     *ipv6.PacketConn
 		requestType icmp.Type
 		replyType   icmp.Type
 	)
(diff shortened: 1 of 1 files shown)


SuperQ:
Nice, we should probably mention in the docs that DF won't work on Windows due to RawConn.

We may possibly want to split this out into separate package files like we do in the `node_exporter`, to avoid building RawConn at all into Windows.  What do you think?

avonwyss:
The docs already specify that: https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md
# Set the DF-bit in the IP-header. Only works with ip4 and on *nix systems.

Not sure about a separate package, users will just get non-successful probes on Windows if they try to use the DF, that's all. Doesn't seem to be too much of an issue IMHO.

SuperQ:
Thanks, that seems like a reasonable enough mitigation.  Thanks.

brian-brazil:
Agreed. The long weekend is just starting here, so I'll review this next week.

brian-brazil:
Thanks!

PR #332, opened 2018-06-11
HTTP response header regex matching

mitchellrj:
Also enables validating if response headers exist or not.

Defaulting to require all headers to match/not match is something I'm not sure about, but seems like the least worst default. Having multiple headers with the same name is a fairly uncommon scenario.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -70,6 +70,7 @@ The other placeholders are specified separately.
   fail_if_header_matches_regexp:
     [ - header: <header-name>
         pattern: <regex>
+        # Also fail if the header is missing from the response.
         required: <bool>
       ... ]
 
@@ -77,6 +78,7 @@ The other placeholders are specified separately.
   fail_if_header_not_matches_regexp:
     [ - header: <header-name>
         pattern: <regex>
+        # Also fail if the header is missing from the response.
         required: <bool>
       ... ]
 
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -66,6 +66,20 @@ The other placeholders are specified separately.
   fail_if_not_matches_regexp:
     [ - <regex>, ... ]
 
+  # Probe fails if response header matches regex.
+  fail_if_header_matches_regexp:
+    [ - header: <header-name>
+        pattern: <regex>
+        required: <bool>
+      ... ]
+
+  # Probe fails if response does not header matches regex.
+  fail_if_header_not_matches_regexp:
+    [ - header: <header-name>
+        pattern: <regex>
+        required: <bool>
+      ... ]
+
   # Configuration for TLS protocol of HTTP probe.
   tls_config:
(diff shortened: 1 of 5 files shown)


mitchellrj:
Any thoughts on this, @SuperQ / @brian-brazil? :)

config/config.go
```diff
@@ -50,20 +50,28 @@ type Module struct {
 	DNS     DNSProbe      `yaml:"dns,omitempty"`
 }
 
+type HTTPHeaderMatch struct {
+	Header   string `yaml:"header,omitempty"`
+	Pattern  string `yaml:"pattern,omitempty"`
+	Required bool   `yaml:"pattern,omitempty"`
```
brian-brazil: pattern?

CONFIGURATION.md
```diff
@@ -66,6 +66,22 @@ The other placeholders are specified separately.
...
+        pattern: <regex>
+        # Also fail if the header is missing from the response.
+        required: <bool>
+      ... ]
+
+  # Probe fails if response does not header matches regex.
+  fail_if_header_not_matches_regexp:
+    [ - header: <header-name>
+        pattern: <regex>
+        # Also fail if the header is missing from the response.
```
brian-brazil: How would you assert that a header must be missing?

CONFIGURATION.md
```diff
@@ -66,6 +66,22 @@ The other placeholders are specified separately.
   fail_if_not_matches_regexp:
```
brian-brazil: This needs clarification that it's for the body

prober/http.go
```diff
@@ -169,9 +209,14 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Help: "Returns the version of HTTP of the probe response",
 		})
 
-		probeFailedDueToRegex = prometheus.NewGauge(prometheus.GaugeOpts{
+		probeFailedDueToHeaderRegex = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_failed_due_to_header_regex",
```
brian-brazil: I think one metric for this is fine, the debug logs will always have the full details

prober/http.go
```diff
@@ -38,13 +38,53 @@ import (
...
+				return false
+			}
+			for _, value := range headerValues {
+				if re.MatchString(value) {
+					level.Error(logger).Log("msg", "Header matched regular expression", "regexp", key, value, matcher.Pattern)
+					return false
+				}
+			}
+		} else if matcher.Required {
+			return false
```
brian-brazil: Log when you return false so the user can know why it failed

mitchellrj:
I've been away on holiday, I shall pick it up again this week, thanks.

brian-brazil:
Did you make any progress since your holiday?

brian-brazil:
Obsoleted by #419

Issue #303, opened 2018-03-08
Hosts gets converted to IP and makes request fail


elbryan:
The request the blackbox exporter is making converts the hostname to IP in the request whilst I would like it to keep the hostname in the request itself. This comes to the fact that there is a proxy validating the request that needs the hostname to properly process it. In my specific case, the request times out.

Is it possible to disable resolving the DNS and to make the request "vanilla"?

### Host operating system: output of `uname -a`

Linux 2c16de9e1e0b 4.4.0-116-generic #140~14.04.1-Ubuntu SMP Fri Feb 16 09:25:20 UTC 2018 x86_64 GNU/Linux

### blackbox_exporter version: output of `blackbox_exporter -version`
0.12.0

### What did you expect to see?

`ts=2018-03-08T14:05:48.014991379Z caller=http.go:282 module=mymodule target=https://hostname/endpoint level=info msg="Making HTTP request" url=https://hostname/endpoint host=hostname
`
### What did you see instead?

`ts=2018-03-08T14:05:48.014991379Z caller=http.go:282 module=mymodule target=https://hostname/endpoint level=info msg="Making HTTP request" url=https://[IPADDRESS]/endpoint host=hostname`

brian-brazil:
No, this is not possible. The blackbox exporter has to do resolution in order to implement v4/v6 selection logic. I'd suggest probing the backend directly.

brian-brazil:
That works off the Host header, not the HTTP request line. That is handled correctly. Proxies where you're depending on the proxy for resolution will not work, we only have that parameter as it happens to come from the http library we use.

vincepii:
So, when using the `blackbox_exporter` against any application that is behind an HTTP load-balancer, the user needs to specify the `Host` header or the request will just stop at the load balancer.

It's not extremely user-friendly as normally one would expect the `Host` header to be set based on the hostname in the URI.

I mean, normally people don't send requests with the `Host` header explicitly set:

```
curl -H "Host: vhost.host.domain.com" http://1.1.1.1
```

But just:

```
curl http://vhost.host.domain.com
```

and then find the `Host` header in the HTTP set to `vhost.host.domain.com`.

I think that, if you want to keep this behavior of translating hostname to IP in the URIs, a note in the documentation should at least be written, as many people will try to probe something behind a load balancer and fail, if they don't know that for that probe configuration the `Host` header should be explicitly set.

vincepii:
[...]

Issue #264, opened 2017-11-18
Difficulty of probing Tor hidden services


ageis:
Debian stretch/9.x
all versions of blackbox_exporter

This is sort of partly a bug and partly a feature request. I wanted to monitor some Tor hidden services using Prometheus and I eventually got it to work, but it required a lot of effort. I discovered that these tools are pretty ill-suited to that use case. Unfortunately I can't tell you exactly what the magic combination was since there's so many factors involved, but I'd like to document my experience and touch upon the difficulties encountered. I'll provide configuration files at the bottom of this post.

My setup is thus: I run Tor with the DNSPort (5353), SOCKSPort (9050) and TransPort (9040) enabled. I configure it to listen on and accept requests from localhost, as well as resolve all .onion addresses to a virtual/fake IPv4 address (AutomapHostsOnResolve+VirtualAddrNetworkIPv4). For some help with caching, I run dnsmasq and put it on top of that, and I then direct _all_ name resolution on my machine through Tor via resolv.conf and/or iptables. 

We also need to have a transparent HTTP proxy in the picture, since Tor doesn't provide one. Alas, it only provides support for SOCKS, transparent TCP proxying, and DNS. SOCKS is not supported by blackbox_exporter, so I guess that would be one of the first problems I'd highlight. But anyway, to solve that I run Privoxy, which listens on port 8118 and sends HTTP requests through Tor's SOCKS port. I set `proxy_url` in the params of my scrape configs in prometheus.yml and include that in blackbox.yml as well, as well as setting the HTTP_PROXY environment variable.

Right off the bat, I had issues with Go's built-in DNS resolver. We need to use CGO in order for this to work! Apparently you can force CGO by setting the RES_OPTIONS environment variable, but I'm not sure on that. So I rebuilt Go from source with the netcgo build tag, plus built the blackbox_exporter binary with `GODEBUG="netdns=cgo+1" GOFLAGS="-tags=netcgo" CGO_ENABLED=1`.  

[...]

PR #338, opened 2018-06-29
http probe: skip DNS lookup if using proxy

veksh:
This PR implements simple work-around to skip DNS lookup for target if `proxy_url` is defined in module config. It seems to be popular request for monitoring availability of external hosts if only proxied access to Internet is available (see #303 and #264).


brian-brazil:
This would break the functionality of choosing v4 vs v6. This probe is not intended to be used with a proxy, and changing this will likely break other uses cases.

veksh:
"v4 vs v6" is fixable by probing proxy address instead of target, I could implement that in PR.

As for using probe with proxy: well, with `proxy_url` parameter (and fixes from this PR) it works pretty well for probing availability and latency of external dependencies. Putting probe to DMZ for direct access will measure different metrics, less relevant to internal application. What are other use cases that will be broken by fixing proxy functionality? Maybe they are fixable too?

brian-brazil:
It's a general complexity concern, you're second guessing the user by doing magic under the covers. That's bound to cause confusion at some point.

brian-brazil:
As I've indicated elsewhere, the proxy_url is only there because the library we use for http configuration includes it. Even with a proxy, we still should be sending the same HTTP request to it so that we know what IP we're talking to - which is important for debugging failed probes.

The purpose of the http probe is to test real http targets, not to test proxies.

veksh:
> ... the proxy_url is only there because the library we use for http configuration includes it.

Ok, and it is a great feature to have, thanks to that http library :)

> Even with a proxy, we still should be sending the same HTTP request to it so that we know what IP we're talking to - which is important for debugging failed probes.

If "it" means "proxy" in this context, we could query it with HTTP to get back (local) v4 vs v6 auto-detection and (local) DNS timing (and mb document that in blackbox_exporter README).

> The purpose of the http probe is to test real http targets, not to test proxies.

Goal is not to test proxy, but to check accessibility of external dependency. Like, our internal app depends on external API (like Google Maps for some geo-spatial calculations), so latency and alerting must be correlated to (proxied) http metrics for external host. Current blackbox_exporter could do that just fine, if only not for DNS lookup requirements. Actually I could proxy limited DNS access to internal host (and for plain http just adding "tgt_hostname" to /etc/hosts seem to do the trick), but it makes little sense outside of making blackbox_exporter happy.

brian-brazil:
> we could query it with HTTP to get back (local) v4 vs v6 auto-detection and (local) DNS timing (and mb document that in blackbox_exporter README).

I don't see how that would work, and that'd be getting a bit involved.

> Actually I could proxy limited DNS access to internal host (and for plain http just adding "tgt_hostname" to /etc/hosts seem to do the trick), but it makes little sense outside of making blackbox_exporter happy.

It sounds like you have a non-trivial setup that includes some form of split-DNS setup, I don't think that's something we can sanely support given all the other features we currently do support I'm afraid.

veksh:
> I don't see how that would work, and that'd be getting a bit involved.

Maybe just resoling proxy address instead of target [would do](https://github.com/veksh/blackbox_exporter/commit/11d5fe9f4300ede0155b0eb1cb731aab62a72575#diff-29df7d46fe6b972c1946a34608c58ab3R226)?

My setup is simple: application could access Internet only via proxy, DNS is local-only so external resources are accessible, but not directly resolvable. Testing proxy itself is easy (just http connect to it), checking external resources must follow the same path that application uses to access them.

brian-brazil:
Per previous discussions looking at a way to test proxy servers is the current plan. It's not determined yet how that would look config wise, as it could be added onto the http probe or be a new probe type.

veksh:
Testing proxy itself is easy: in case of [squid](http://www.squid-cache.org/) it could be [implemented](https://wiki.squid-cache.org/Features/CacheManager#default) with either http_probe or just connecting to proxy port (3128 by default) and issuing "GET cache_object://mycache.example.com/info HTTP/1.0". Checking remote host accessibility via proxy is another matter.

brian-brazil:
Think more how the dns probe is to test dns servers rather than dns resolution generally.

alexeiaguiar:
Our security department is implemented a policy that all services in production must pass through an outbound proxy to reach external services. We are the first team to implement that. We configured tens of applications to use the proxy and all them worked, including many that were developed in Go. blackbox_exporter is the only one that doesn't word because it sends the ip address instead of sending the hostname like all other applications do. I did a fix that is pretty similar to this one and I was about to send it when I found this PR.

dabear:
@alexeiaguiar agreed, we do have the same problem at my workspace currently. I've added a comment to #303 but that issue seems to be closed. Makes it kinda hard to monitor sites that changes ips frequently, when you have to pass through a proxy for outgoing connections to the internet.

dabear:
@brian-brazil @veksh Any chance to get this PR built against latest master and mainlined?

DeviaVir:
@dabear feel free to grab https://github.com/deviavir/blackbox_exporter, in a Dockerfile:
```
FROM golang:alpine

RUN apk add --no-cache git
RUN go get github.com/DeviaVir/blackbox_exporter \
    && GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build github.com/DeviaVir/blackbox_exporter

FROM alpine:latest
COPY --from=0 /go/bin/blackbox_exporter /bin/blackbox_exporter

EXPOSE      9115
ENTRYPOINT  [ "/bin/blackbox_exporter" ]
CMD         [ "--config.file=/etc/blackbox_exporter/config.yml" ]
```

SuperQ:
@DeviaVir Would you be able to cherrypick and publish a new PR to replace this one?

DeviaVir:
@SuperQ I think @brian-brazil was clear in that the PR in its current form will not be merged. Feel free to use my fork.

raypettersen:
Voicing additional concern, blackbox must be able to function through a proxy, so we can safe-guard the applications and detect anomalies in proxy-only environments.

veksh:
I've (force-)rebased this PR against current master.

brian-brazil:
As indicated above, this change will not be accepted. However there is the path of making the http probe suitable for probing proxies, which would also cover your use case.

veksh:
I've also updated [binary](https://github.com/veksh/blackbox_exporter/releases/tag/v0.12.0-proxy)

veksh:
@brian-brazil once again: this PR is not for probing proxies, but for probing http servers **via proxy**.

brian-brazil:
I am aware, and that is not in line with the rest of the exporter.

DeviaVir:
@brian-brazil what is the recommended way of healthchecking tor addresses using the exporter?

dnascimento:
Same use case here, proxy is the only way to reach outside. 
How can we make changes so that this does not break ipv4/ipv6 but supports via proxy?

brian-brazil:
A thought: Why not put the blackbox exporter on the other side of your proxy?

okushchenko:
> A thought: Why not put the blackbox exporter on the other side of your proxy?

@brian-brazil this is not possible in some environment that mandate usage of HTTP proxies for all outgoing traffic as an InfoSec practice.

brian-brazil:
You'd still be going through the proxy, the blackbox exporter would live on the other side of it.

raypettersen:
This is becoming a bit silly. It's clear that Brian doesn't want this feature at all in the exporter, and we're forced to use forked versions for getting this basic feature as PRs are just left to rot. If it was a wanted feature then I would expect a more constructive tone than this.

brian-brazil:
Constructive in this context means helping solve the problem where possible and not leading you on, on which front I've made several suggestions and made clear what won't be accepted. Constructive does not mean agreeing with everything someone else says.

As this is going nowhere, I'm going to close and lock this. You're always free to discuss this on prometheus-developers.

PR #353, opened 2018-09-04
[ PoC - IDEA ] calculate download time instead of simple reachability

maurorappa:
I wanted to monitor the download time from our internal repositories (Artifactory) using blackbox instead of creating a new tool/script.
I checked the code and I saw the URL after being contacted gets silently discarded by a goroutine, by simply removing it, I was able to monitor the time needed to download a binary from the internal repo.
I used an url like http://artifactory.local/files/Debian7.iso and I can monitor variation of its download time.
I think this could be useful to everybody and it could be refined / transformed in a configuration at a target level. 
I am aware the removed goroutine was used to reduce the load of the task, but it could be easily parametrized to provide this new functionality.

Let me know your thoughts.

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -296,10 +296,8 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if err != nil && resp == nil {
 		level.Error(logger).Log("msg", "Error for HTTP request", "err", err)
 	} else {
-		
-			io.Copy(ioutil.Discard, resp.Body)
-			resp.Body.Close()
-		
+		io.Copy(ioutil.Discard, resp.Body)
+		resp.Body.Close()
 		level.Info(logger).Log("msg", "Received HTTP response", "status_code", resp.StatusCode)
 		if len(httpConfig.ValidStatusCodes) != 0 {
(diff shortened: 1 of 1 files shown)


brian-brazil:
I believe this would break the regular expression matching, which we need to keep working.

maurorappa:
sorry, I don't get it, can you articulate a bit more?

brian-brazil:
If you read in and throw away the body then it's not available for regex matching later one.

SuperQ:
The blackbox exporter includes timing information in the probe results data. See the metric `probe_http_duration_seconds{phase="transfer"}`.

If you're looking to test download time, but it's slow enough that it could take longer than a couple minutes, you could use `curl` to execute the probes from cron, and push the data to the pushgateway or node_exporter's file collector. This avoids having to worry about Prometheus server scrape interval/timeouts.

maurorappa:
In my tests the transfer phase  is always roughtly the same time regardless the size of the URL object, this is why I wrote this.
Using Curl is of course a working solution but it's another moving part.

brian-brazil:
I believe this is fixed in 0.13.0

Issue #356, opened 2018-09-08
Introduce option to set TOS field in ICMP and other probes


s1sfa:
I use the TOS field for PBR routing to monitor ping latency through different paths on my network. 

Currently I use smokeping for this and it would be interesting to replace with this.
Probe setup Smokeping config:
```
+ FPing
binary = /usr/bin/fping

++ FPingPath1
tos = 0x30

++ FPingPath2
tos = 0x20

++ FPingPath3
tos = 0x28
```

Or using the ping -Q flag `ping 8.8.8.8 -Q 32`

I've seen use cases in other tools where others like to set TOS for other tests not only for PBR but maybe want to set a certain queuing priority on their network for testing traffic which is why I think this feature can apply to all modules.

brian-brazil:
Does Go offer a way to set this on things like http requests?

fredericve:
I am not at all acquainted with Go, but it looks like it should be possible with Go 1.11 according to [this post on SO](https://stackoverflow.com/questions/40544096/how-to-set-socket-option-ip-tos-for-http-client-in-go-language)

electron0zero:
~Hi 👋🏼, coming from the cleanup 🧹~

~This PR is quite old and there's https://github.com/prometheus/blackbox_exporter/pull/613 looks to be superset of this PR.~

~I am going to close this, please feel free to reopen if you want to update it, and discuss it further.~

electron0zero:
oops, meant to close the [PR](https://github.com/prometheus/blackbox_exporter/pull/358). reopening this issue.

PR #358, opened 2018-09-14
allow setting the TOS bit for icmp probes

fredericve:
Partly implements #356 as it allows to set outgoing TOS bits for ICMP probes.

This is my first encounter with Go, so I may be missing a million things here. Please bear with me :-)

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -137,6 +137,9 @@ query_response:
 tls_config:
   [ <tls_config> ]
 
+# The TOS bits (Traffic Class for ipv6) to set on the outgoing TCP packets
+[ tos: <int> ]
+
 ```
 
 ### <dns_probe>
@@ -184,6 +187,9 @@ validate_additional_rrs:
   fail_if_not_matches_regexp:
     [ - <regex>, ... ]
 
+# The TOS bits (Traffic Class for ipv6) to set on the outgoing DNS requests
+[ tos: <int> ]
+
 ```
 
 ### <icmp_probe>
@@ -205,7 +211,6 @@ validate_additional_rrs:
 
 # The TOS bits (Traffic Class for ipv6) to set on the outgoing ICMP packets
 [ tos: <int> ]
-
 
 ```
 
--- a/config/config.go
+++ b/config/config.go
@@ -157,6 +157,7 @@ type TCPProbe struct {
 	QueryResponse      []QueryResponse  `yaml:"query_response,omitempty"`
 	TLS                bool             `yaml:"tls,omitempty"`
 	TLSConfig          config.TLSConfig `yaml:"tls_config,omitempty"`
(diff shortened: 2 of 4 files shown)


fredericve:
Note that I am unable to test this for ipv6 since I have no environment for it.

brian-brazil:
Can you add this for all the probes?

Does this require privileges?

fredericve:
@brian-brazil 
I updated the PR for the dns and tcp probes. I can confirm this works with go1.11 but the build fails since it is building on golang-builder:1.10-base and there is no 1.11-base to switch to.

For the http probe, I attempted to add it as well, but I'm not familiar with the http RoundTrippers (or go in general) to get this done. Can someone else help on this or give me a hint on how it should be done?

Concerning privileges, no special privileges are required as far as I can tell (apart from the privileges already required for icmp probes). I'm testing with the blackbox exporter running as its own, unprivileged user.

SuperQ:
We'll be updating our build pipelines to go1.11 soon, we're just waiting on upstream support in our tools repo.

brian-brazil:
> Can someone else help on this or give me a hint on how it should be done?

Looks like DialContext is the current way to do that.

fredericve:
@brian-brazil Thanks for the suggestion, but I really don't know what that means. I don't know how to set or override the `DialContext` of an already existing `Transport`

brian-brazil:
DialContext is a public field of the Transport struct, so you should be able to assign to it.

aomi:
I was wondering if there was any development on this recently as I was also looking to attempt to implement this feature as well.

fredericve:
@aomi 

Basically all probes were implemented in this PR, except for the http probe. I don't have the experience with GO and the http roundtrippers to get this implemented. If you do, any help would be greatly appreciated!

fredericve:
@brian-brazil or @SuperQ any chance this can get merged without http TOS support?

SuperQ:
@fredericve It looks like this needs to be rebased, and the tests are broken.

fredericve:
@SuperQ I rebased the PR.

The build fails on windows only. SetsockoptInt on windows wants a Handle object. On other OS an int. Any way I can fix that? I tried something like this:
```
                    if runtime.GOOS == "windows" {
                        handle := syscall.Handle(fd)
                    }
                    err := syscall.SetsockoptInt(handle, syscall.IPPROTO_IP, syscall.IP_TOS, module.DNS.TOS)
```
but that does not compile on linux:
```
prober/dns.go:160:17: undefined: syscall.Handle
```

brian-brazil:
You'll need to use different files with build tags to cover that. Check out
the node exporter for examples.

electron0zero:

This PR is quite old and there's https://github.com/prometheus/blackbox_exporter/pull/613 looks to be superset of this PR.

I am going to close this, please feel free to reopen if you want to update it, and discuss it further.

PR #363, opened 2018-09-25
Account response read time for http prober

bobrik:
Withot this change `transfer` time is near zero, because `trace.end` time is clocked when response body is available, not when it's read.

cc @discordianfish

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -104,7 +104,6 @@ func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 	t.current = trace
 	t.traces = append(t.traces, trace)
-	defer func() { trace.end = time.Now() }()
 	return t.Transport.RoundTrip(req)
 }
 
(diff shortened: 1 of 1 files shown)


prober/http.go
```diff
@@ -296,10 +295,8 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if err != nil && resp == nil {
 		level.Error(logger).Log("msg", "Error for HTTP request", "err", err)
 	} else {
-		defer func() {
-			io.Copy(ioutil.Discard, resp.Body)
-			resp.Body.Close()
-		}()
+		requestErrored := err != nil
```
brian-brazil: It'd be usual to put parens around the RHS here

prober/http.go
```diff
@@ -327,6 +324,19 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+			_, err = io.Copy(ioutil.Discard, resp.Body)
+			if err != nil {
+				level.Info(logger).Log("msg", "Failed to read HTTP response body", "err", err)
+				success = false
+			}
+
+			resp.Body.Close()
+		}
+
+		// At this point body is fully read and we can write end time
```
brian-brazil: Full stop

prober/http.go
```diff
@@ -327,6 +324,19 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+		tt.current.end = time.Now()
```
brian-brazil: This is only setting the end time for the last request

bobrik: Considering that only the last request matters for `transfer` times, should I prefix this:

```go
durationGaugeVec.WithLabelValues("transfer").Add(trace.end.Sub(trace.responseStart).Seconds())
```

with this:

```go
if trace.end.IsZero() {
	continue
}
```

or do you have other suggestions?

brian-brazil: That sounds reasonable to me, sound okay to you @discordianfish?

bobrik: I've pushed the change speculatively.

discordianfish: Sounds reasonable

brian-brazil:
Thanks!

PR #372, opened 2018-10-15
Update vendor github.com/prometheus/client_golang/...@v0.9.0

SuperQ:
Signed-off-by: Ben Kochie <superq@gmail.com>

PR #373, opened 2018-10-15
WIP: Update vendor github.com/prometheus/common/...

SuperQ:
Follow-up to https://github.com/prometheus/blackbox_exporter/pull/372


beorn7:
Oooh, something got removed from common/config...

SuperQ:
@beorn7 Yea, this one was way out of date.  Looks like the use was changed slightly. Pushed a fix.

beorn7:
Is that the story where somebody decided keepalive is always good?

brian-brazil:
No, different thing. Someone was doing some other work which removed the option to disable it, and I hadn't gotten around to fixing it all yet.

beorn7:
Alright. I'd say for now it's fine to not update the prometheus/common vendoring as the metric response size of the blackbox exporter is usually small.

beorn7:
Hmm, following https://github.com/prometheus/client_golang/issues/366 , it seems my assumption was wrong and a blackbox exporter might run quite a bit of encoding load.

So I guess we should update at least the vendoring of the expfmt package (which could leave out the HTTP config stuff, even if it is ugly to vendor different commits from the same repo) to pull in the text format creation improvement. We should also consider to allow users to disable gzip compression (client_golang allows this via HandlerOpts, but it has to be exposed by a command line flag for the blackbox_exporter).

beorn7:
How about my suggestion above?

brian-brazil:
Partial vendoring sounds okay to me.

beorn7:
I guess with Go modules (as they are now used in blackbox_exporter), the partial vendoring doesn't work anymore. I guess it makes most sense to close this PR and do a normal Go modules update once `common` is fixed.

PR #382, opened 2018-11-14
Add force preferred ip protocol

laghoule:
In some case we may want to force a preferred protocol. 
I'v added an option (false by default) for that purpose:
[ force_preferred_ip_protocol: <boolean | default = false> ]

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -84,9 +84,9 @@ The other placeholders are specified separately.
   # HTTP proxy server to use to connect to the targets.
   [ proxy_url: <string> ]
 
-  # The IP protocol of the HTTP probe (ip4, ip6).
-  [ ip_protocol: <string> | default = "ip6" ]
-  [ preferred_ip_protocol: <boolean | default = false> ]
+  # The preferred IP protocol of the HTTP probe (ip4, ip6).
+  [ preferred_ip_protocol: <string> | default = "ip6" ]
+  [ force_preferred_ip_protocol: <boolean | default = false> ]
 
   # The body of the HTTP request used in probe.
   body: [ <string> ]
@@ -98,9 +98,9 @@ The other placeholders are specified separately.
 
 ```yml
 
-# The IP protocol of the TCP probe (ip4, ip6).
-[ ip_protocol: <string> | default = "ip6" ]
-[ preferred_ip_protocol: <boolean | default = false> ]
+# The preferred IP protocol of the TCP probe (ip4, ip6).
+[ preferred_ip_protocol: <string> | default = "ip6" ]
+[ force_preferred_ip_protocol: <boolean | default = false> ]
 
 # The source IP address.
 [ source_ip_address: <string> ]
@@ -127,9 +127,9 @@ tls_config:
 
 ```yml
 
-# The IP protocol of the DNS probe (ip4, ip6).
-[ ip_protocol: <string> | default = "ip6" ]
-[ preferred_ip_protocol: <boolean | default = false> ]
+# The preferred IP protocol of the DNS probe (ip4, ip6).
+[ preferred_ip_protocol: <string> | default = "ip6" ]
+[ force_preferred_ip_protocol: <boolean | default = false> ]
 
 # The source IP address.
 [ source_ip_address: <string> ]
@@ -174,9 +174,9 @@ validate_additional_rrs:
 
 ```yml
 
-# The IP protocol of the ICMP probe (ip4, ip6).
-[ ip_protocol: <string> | default = "ip6" ]
-[ preferred_ip_protocol: <boolean | default = false> ]
+# The preferred IP protocol of the ICMP probe (ip4, ip6).
+[ preferred_ip_protocol: <string> | default = "ip6" ]
+[ force_preferred_ip_protocol: <boolean | default = false> ]
 
 # The source IP address.
 [ source_ip_address: <string> ]
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -86,6 +86,7 @@ The other placeholders are specified separately.
 
   # The preferred IP protocol of the HTTP probe (ip4, ip6).
   [ preferred_ip_protocol: <string> | default = "ip6" ]
+  [ force_preferred_ip_protocol: <boolean | default = false> ]
 
   # The body of the HTTP request used in probe.
   body: [ <string> ]
@@ -99,6 +100,7 @@ The other placeholders are specified separately.
 
 # The preferred IP protocol of the TCP probe (ip4, ip6).
 [ preferred_ip_protocol: <string> | default = "ip6" ]
+[ force_preferred_ip_protocol: <boolean | default = false> ]
 
 # The source IP address.
 [ source_ip_address: <string> ]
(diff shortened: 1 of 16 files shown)


SuperQ:
What about calling this feature `ip_protocol`? IMO "forced" should be implied, as this is how most utilities do it. (IE `curl -(4|6)`)

SuperQ:
@laghoule Yes, I understand. What I'm suggesting is we simplify the naming so that we have two options. `preferred_ip_protocol`, and `ip_protocol`. In the new option "Forced" is implied, as this is the standard behavior expected by most tools when setting a protocol.

laghoule:
@SuperQ sorry for the misunderstanding. I agree with you. I will make the modification.
Thanks for the feedback.

CONFIGURATION.md
```diff
@@ -84,8 +84,9 @@ The other placeholders are specified separately.
   # HTTP proxy server to use to connect to the targets.
   [ proxy_url: <string> ]
 
-  # The preferred IP protocol of the HTTP probe (ip4, ip6).
-  [ preferred_ip_protocol: <string> | default = "ip6" ]
+  # The IP protocol of the HTTP probe (ip4, ip6).
+  [ ip_protocol: <string> | default = "ip6" ]
+  [ preferred_ip_protocol: <boolean | default = false> ]
```
SuperQ: Typo, should be `preferred_ip_protocol: <boolean> | default = false`

SuperQ:
I would like to maintain the original behavior, which is to "preferred" mode by default.

Also, with the switch to a bool, `preferred` doesn't make sense. what about `fallback_ip_protocol` with a default of `true`?

laghoule:
Yeah make sense. I will do that.

SuperQ:
Oh, I think there was a misunderstanding. I want to keep `ip_protocol`, and drop `preferred_ip_protocol` entirely. This way the exporter will behave more like standard tools, but have fallback by default.

laghoule:
@SuperQ sorry, I was doing the change after work, my brain was not fully operational ;)
Should be good now.

prober/utils.go
```diff
@@ -25,32 +25,37 @@ func chooseProtocol(preferredIPProtocol string, target string, registry *prometh
...
 		probeDNSLookupTimeSeconds.Add(lookupTime)
 	}()
 
-	ip, err = net.ResolveIPAddr(preferredIPProtocol, target)
+	ip, err = net.ResolveIPAddr(IPProtocol, target)
 	if err != nil {
-		level.Warn(logger).Log("msg", "Resolution with preferred IP protocol failed, attempting fallback protocol", "fallback_protocol", fallbackProtocol, "err", err)
-		ip, err = net.ResolveIPAddr(fallbackProtocol, target)
+		if fallbackIPProtocol == false {
+			level.Error(logger).Log("msg", "Resolution with IP protocol failed (fallback IP protocol is FALSE): err", err)
```
brian-brazil: fallback_ip_protocol would be easier for users to find, and the FALSE could be false.

CONFIGURATION.md
```diff
@@ -84,8 +84,9 @@ The other placeholders are specified separately.
   # HTTP proxy server to use to connect to the targets.
   [ proxy_url: <string> ]
 
-  # The preferred IP protocol of the HTTP probe (ip4, ip6).
-  [ preferred_ip_protocol: <string> | default = "ip6" ]
+  # The IP protocol of the HTTP probe (ip4, ip6).
+  [ ip_protocol: <string> | default = "ip6" ]
```
brian-brazil: This will break any users currently setting this field, so I'm not sure about changing its name.

laghoule: Maybe support both, and print a warning about deprecating `preferred_ip_protocol` in the next release?

SuperQ: I wanted to make it obvious that this is a behavior change.

But I like the idea of a deprecation warning.

brian-brazil:
The default behavior is the same though.

SuperQ:
@brian-brazil Yea, works for me. I think adding `preferred_ip_protocol` as a config alias with a deprecation warning is a good option.

SuperQ:
@laghoule, do you want to implement the `preferred_ip_protocol` as an alias for `ip_protocol`, with a warning that it's deprecated? I think that's the only thing we need for merging.

laghoule:
@SuperQ, yes I will do it. I hope to have some free time this week to work on this.

CONFIGURATION.md
```diff
@@ -84,8 +84,10 @@ The other placeholders are specified separately.
   # HTTP proxy server to use to connect to the targets.
   [ proxy_url: <string> ]
 
-  # The preferred IP protocol of the HTTP probe (ip4, ip6).
-  [ preferred_ip_protocol: <string> | default = "ip6" ]
+  # The IP protocol of the HTTP probe (ip4, ip6).
+  [ preferred_ip_protocol: <string> | default = "ip6" *DEPRECATED* ]
```
brian-brazil: I don't see why we're deprecating this just to change its name, this would cause unnecessary work for all our existing users.

CONFIGURATION.md
```diff
@@ -84,8 +84,10 @@ The other placeholders are specified separately.
...
+  [ ip_protocol: <string> | default = "ip6" ]
+  [ fallback_ip_protocol: <boolean | default = true> ]
```
brian-brazil: ip_protocol_fallback would make it clearer that this is a bool, rather than a string.

SuperQ: I never liked the name or behavior of this code. With the new code, it's nolonger exactly "preferred", it's just specifying a specific protocol, and having a fallback or not.

I do not want to go in circles on this review.

CONFIGURATION.md
SuperQ: I never liked the name or behavior of this code. With the new code, it's nolonger exactly "preferred", it's just specifying a specific protocol, and having a fallback or not.

I do not want to go in circles on this review.

brian-brazil: It's still preferred, and we've already broken this config option before. If we're going to break it again I want to to be due a complete change in how it works, not a minor name change.

SuperQ: It is not, the new behavior does not represent preferred anymore. With preferred, fallback was always an option. Now fallback is optional, and this necessitates a name change.

brian-brazil: I don't think that adding a new option necessitates breaking existing users.

CONFIGURATION.md
laghoule: I agree, it's better visually and conceptually.

laghoule:
@brian-brazil @SuperQ Let me know what you want to do about the breaking changes / deprecation issues.

laghoule:
Ok, I will adjust the behavior.

laghoule:
Is it OK to keep IPProtocol var, but use the preferred_ip_protocol yaml config?
```
IPProtocol             string                  `yaml:"preferred_ip_protocol,omitempty"`
```

brian-brazil:
That's fine.

brian-brazil:
Thanks!

brian-brazil:
I was just testing this for release, and the default isn't true as the docs say.

laghoule:
I will take a look at it as soon as possible.

Daniel15:
Is there somewhere I can download an alpha build containing this change, or would I need to build blackbox_exporter myself?

laghoule:
I will have the merge request this week. I will post it here.

Pascal Gauthier

brian-brazil:
Great!

laghoule:
This MR should fix the issue: https://github.com/prometheus/blackbox_exporter/pull/436
Give it a try / review / comments.

PR #383, opened 2018-11-16
adding probe error log support 

QingsongYao:
this PR address issue #350 due to we need to keep failed probe history longer for people to investigate issues. debug logging can not address this issue due to people should not need to look at log to troubleshot probe issue. when probing large number of sites, the failed log entry will be override quickly by succeed logs, so we create a separate entry for this.

I have no tested the change yet due to don't know how to setup build and test on local env

### DIFF
--- a/main.go
+++ b/main.go
@@ -51,7 +51,6 @@ var (
 	timeoutOffset = kingpin.Flag("timeout-offset", "Offset to subtract from timeout in seconds.").Default("0.5").Float64()
 	configCheck   = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
 	historyLimit  = kingpin.Flag("history.limit", "The maximum amount of items to keep in the history.").Default("100").Uint()
-	errorHistoryLimit  = kingpin.Flag("error.history.limit", "The maximum amount of items to keep in the probe error history.").Default("300").Uint()
 
 	Probers = map[string]prober.ProbeFn{
 		"http": prober.ProbeHTTP,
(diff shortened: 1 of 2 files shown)


brian-brazil:
I think this would be better combined with the existing list, and reusing the existing flag.

QingsongYao:
It make sense that existing list track succeed requests,  new failed requested list tracked failed requests.  I will reusing the existing flag

brian-brazil:
The existing list tracks all requests.

brian-brazil:
You can have two lists, and merge them.

QingsongYao:
Thank you for the suggestion. I made change to this PR to use two lists and merged them when display to UI.

main.go
```diff
@@ -285,14 +290,18 @@ func main() {
     <h2>Recent Probes</h2>
     <table border='1'><tr><th>Module</th><th>Target</th><th>Result</th><th>Debug</th>`))
 
-		results := rh.List()
+		results := eh.List()
```
brian-brazil: You're not merging the lists, it should be ordered by time.

main.go
```diff
@@ -308,6 +317,9 @@ func main() {
 			return
 		}
 		result := rh.Get(id)
+		if r.URL.Query().Get("error") == "true" {
```
brian-brazil: This should work transparently

main.go
QingsongYao: result struct has no datetime field, so it will be hard to merge and sort by time.

brian-brazil: You could use the ids

QingsongYao: The ids in two list might conflicts since both set and increase independently .

brian-brazil: A common counter will handle that

SuperQ:
It would be nice to get this done, being able to log all errors rather than just all results would be pretty useful.

brian-brazil:
Obsoleted by #517

Issue #391, opened 2018-12-14
Feature request: ability to specify HTTP probe POST body content as on-disk file


emmanuel:
Blackbox exporter config allows specifying POST body for HTTP probe as an in-line string in the config file. I have a circumstance where I want to probe with some large body sizes, and there is a significant memory impact of treating those bodies as inline strings.

I propose adding a `BodyPath` parameter to the HTTP prober, that will be used to open a file on disk and use the content of that file as the body of the probe's requests. Something like this:

```
type HTTPProbe struct {
	// Defaults to 2xx.
	ValidStatusCodes       []int             `yaml:"valid_status_codes,omitempty"`
...
	Body                   string            `yaml:"body,omitempty"`
	BodyPath               string            `yaml:"body_path,omitempty"`
	HTTPClientConfig       config.HTTPClientConfig `yaml:"http_client_config,inline"`
}
```

brian-brazil:
How big is the data you're sending?

emmanuel:
1.6MiB, so far. 

Also, I have multiple endpoints I'm scraping with the same bodies (with different HTTP parameters of other sorts), so the same data is parsed into memory as multiple copies of this large string. The current (in-line string) config option compels this duplication. Plus (and more essentially), YAML is just not a good medium for values of that size.

Turgon37:
Hello I'm also looking for this feature @emmanuel have any time to works on this ?

Thanks in advance

birmogave87:
i am looking for something like this also Feature Request: Is there a way to scrape "Probe Logs" #586

PR #392, opened 2018-12-14
Add support for `body_path` in HTTP probe config.

emmanuel:
Fixes prometheus/blackbox_exporter#391.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -142,10 +142,11 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return errors.New("at most one of body & body_file may be configured")
 	}
 	if s.BodyFile != "" {
-		_, err := os.Open(s.BodyFile)
+		file, err := os.Open(s.BodyFile)
 		if err != nil {
 			return err
 		}
+		defer file.Close()
 	}
 	return nil
 }
--- a/config/config.go
+++ b/config/config.go
@@ -138,8 +138,8 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := s.HTTPClientConfig.Validate(); err != nil {
 		return err
 	}
-	if s.Body != "" && s.BodyFile != "" {
-		return errors.New("Only one of `body` and `body_file` may be set")
+	if len(s.Body) > 0 && len(s.BodyFile) > 0 {
+		return errors.New("at most one of body & body_file may be configured")
 	}
 	if s.BodyFile != "" {
 		_, err := os.Open(s.BodyFile)
--- a/config/config.go
+++ b/config/config.go
@@ -3,7 +3,9 @@ package config
 import (
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"os"
 	"runtime"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -135,6 +139,13 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := s.HTTPClientConfig.Validate(); err != nil {
 		return err
 	}
+	if s.BodyPath != "" {
```
brian-brazil: This should be opened and read in the request. It should also be an error for both to be specified.

config/config.go
```diff
@@ -52,18 +54,20 @@ type Module struct {
...
+	IPProtocolFallback     bool              `yaml:"ip_protocol_fallback,omitempty"`
+	NoFollowRedirects      bool              `yaml:"no_follow_redirects,omitempty"`
+	FailIfSSL              bool              `yaml:"fail_if_ssl,omitempty"`
+	FailIfNotSSL           bool              `yaml:"fail_if_not_ssl,omitempty"`
+	Method                 string            `yaml:"method,omitempty"`
+	Headers                map[string]string `yaml:"headers,omitempty"`
+	FailIfMatchesRegexp    []string          `yaml:"fail_if_matches_regexp,omitempty"`
+	FailIfNotMatchesRegexp []string          `yaml:"fail_if_not_matches_regexp,omitempty"`
+	Body                   string            `yaml:"body,omitempty"`
+	BodyPath               string            `yaml:"body_path,omitempty"`
```
brian-brazil: body_file is the usual naming

config/config.go
emmanuel: Updated to be opened within `ProbeHTTP`. I left the current `os.Open` call in `config.go` in order to fail fast on bad configs. This is different from `BearerToken`/`BearerTokenFile` (https://github.com/prometheus/common/blob/master/config/http_config.go#L81).

Let me know if I should drop opening the file during config validation.

config/config.go
emmanuel: Renamed to `body_file`.

emmanuel:
Added DCO. I'm not sure why the Travis build is failing, but from my (short) look, it appears related to git remotes, and not the content of these changes.

config/config.go
```diff
@@ -135,6 +138,15 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := s.HTTPClientConfig.Validate(); err != nil {
 		return err
 	}
+	if len(s.Body) > 0 && len(s.BodyFile) > 0 {
+		return errors.New("at most one of body & body_file may be configured")
+	}
+	if s.BodyFile != "" {
+		_, err := os.Open(s.BodyFile)
```
brian-brazil: You should close the file

prober/http.go
```diff
@@ -257,10 +258,19 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
 	var body io.Reader
 
-	// If a body is configured, add it to the request.
+	// If a body string is configured, add it to the request.
 	if httpConfig.Body != "" {
 		body = strings.NewReader(httpConfig.Body)
 	}
+	// If a body path is configured, add it to the request.
+	if httpConfig.BodyFile != "" {
+		file, err := os.Open(httpConfig.BodyFile)
```
brian-brazil: Make sure this gets closed

prober/http_test.go
```diff
@@ -164,6 +164,33 @@ func TestPost(t *testing.T) {
...
+	registry := prometheus.NewRegistry()
+	testCTX, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	result := ProbeHTTP(testCTX, ts.URL,
+		config.Module{
+			Timeout: time.Second,
+			HTTP: config.HTTPProbe{
+				IPProtocolFallback: true,
+				Method:             "POST",
+				BodyFile:           "./http.go",
```
brian-brazil: You can create a testdata directory, and a file with known contents there - plus verify that's what's coming through.

brian-brazil: You should still change this to use explicit testdata files, rather than the source itself

brian-brazil:
This has gone stale, let us know if you still wish to work on this.

PR #393, opened 2018-12-15
Update prometheus/common to support --log.format option

shusugmt:
Support for json format logging in `prometheus/common/promlog` has been added since https://github.com/prometheus/common/pull/136
This PR updates `prometheus/common` to the latest version which supports `--log.format` option. 

The update also includes code change to adopt https://github.com/prometheus/common/pull/129
As a result now http prober exposes new metrics described in https://github.com/prometheus/prometheus/pull/3241 via `/metrics` endpoint.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,11 +17,12 @@ require (
 	github.com/mattn/go-isatty v0.0.4 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.1 // indirect
 	github.com/miekg/dns v1.0.1
+	github.com/mwitkow/go-conntrack v0.0.0-20161129095857-cc309e4a2223 // indirect
 	github.com/pkg/errors v0.0.0-20170910134614-2b3a18b5f0fb // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/prometheus/client_golang v0.9.0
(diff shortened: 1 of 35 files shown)


brian-brazil:
Thanks for working on this. The newer version of common also makes all http connections use keepalives, which is undesirable for this particular exporter. That'll need work first.

shusugmt:
I just noticed there are a few similar branches trying to update `prometheus/common`
- https://github.com/prometheus/blackbox_exporter/tree/fish-update-dependencies
- https://github.com/prometheus/blackbox_exporter/tree/bjk/prom_common

And found this PR https://github.com/prometheus/blackbox_exporter/pull/321
Then that leads to https://github.com/prometheus/common/pull/133 which seems blocking this to proceed.

shusugmt:
> The newer version of common also makes all http connections use keepalives, which is undesirable for this particular exporter

Solution seems quite straightforward once https://github.com/prometheus/common/pull/133 got merged. Simply pass `true` as the third parameter to `NewClientFromConfig()` in https://github.com/prometheus/blackbox_exporter/pull/393/files#diff-29df7d46fe6b972c1946a34608c58ab3R219

brian-brazil:
This is obsolete.

PR #407, opened 2019-01-16
Implemented Last-Modified HTTP header metric

joker234:
This feature returns a `probe_http_last_modified` metric with the content of the Last-Modified HTTP response header, if it is set by the server.

That feature would be nice for (blackbox) monitoring, if a service that should regularly update a website, is working or not.

-------

P.S.: someone encountered me to the fact, that registering the gauges shouldn't be repeated in each probe, but only at initializing the exporter. This isn't part of this PR.
https://github.com/prometheus/blackbox_exporter/blob/22be14377a74e81e6c4c77a5577c14f944aa2277/prober/http.go#L181-L187

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -172,6 +172,11 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Name: "probe_failed_due_to_regex",
 			Help: "Indicates if probe failed due to regex",
 		})
+
+		probeHTTPLastModified = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_last_modified_timestamp_seconds",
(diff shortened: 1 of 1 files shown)


prober/http.go
```diff
@@ -172,6 +172,11 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Name: "probe_failed_due_to_regex",
 			Help: "Indicates if probe failed due to regex",
 		})
+
+		probeHTTPLastModified = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_last_modified",
```
brian-brazil: This should end in timestamp_seconds

brian-brazil: Yes

brian-brazil:
Thanks!

Issue #411, opened 2019-01-29
ICMP probes fail when running multiple blackbox_exporters behind NAT


BenoitKnecht:
Let me first describe my setup. I have two hosts on the same subnet, each running their own instance of Prometheus and a blackbox_exporter, running in Docker containers. Each Prometheus instance scrapes the corresponding blackbox_exporter on localhost. The two hosts have internet access through a firewall that performs NAT.

I use Ansible to manage those hosts, and whenever I change the blackbox_exporter configuration, both instances are restarted to pickup the new parameters. But after the restart, the ICMP probes from one of the hosts would fail, while the same probes on the other host were fine; but only the probes to public IP addresses, probes to IP addresses on the LAN were fine (so it's potentially related to #360, I don't know if the author of that bug also has multiple blackbox_exporters running in parallel).

After banging my head on my keyboard for a few days, I finally figured out what's going on.

The blackbox_exporter sets the ICMP echo ID to `os.Getpid() & 0xffff`, which in a Docker container is always `1`, and after a simultaneous restart, `int(getICMPSequence())` will be in sync on both blackbox_exporter instances:

```go
body := &icmp.Echo{
    ID:   os.Getpid() & 0xffff,
    Seq:  int(getICMPSequence()),
    Data: data,
}
```

When the firewall NATs those packets, it creates a session based on the ID and sequence number, and when the target replies, it doesn't know which packet should be NATed back to which source IP address, so all the replies go to a single blackbox_exporter. Restarting one of the blackbox_exporters fixes the issue because it resets `int(getICMPSequence())` so that they're out of sync between the two instances.

To be fair, it's technically a bug in the way this particular firewall (a Palo Alto) NATs ICMP traffic; a Linux firewall not only NATs the source IP address, it also NATs the ICMP echo ID in such a way that it can keep track of the two sessions separately.

However, it's not something that can easily be fixed, and my guess is that there are many more network equipments that would misbehave when faced with such ICMP traffic.

I see several ways in which the blackbox_exporter could help working around this issue:

1. Set `body.ID` to a random integer instead of `os.Getpid()`; maybe only do that if `os.Getpid() == 1`, so that it only applies to instances running in a container.
2. Initialize `icmpSequence` with a random offset, to make sure two instances started at the same time aren't in sync;
[...]

PR #412, opened 2019-01-30
Initialize ICMP echo ID and sequence to random values

BenoitKnecht:
This should help preventing issues with some network devices that have
trouble NATing ICMP packets with the same ID and sequence nubmer but a
different source IP address.

Currently, this can happen if the blackbox_exporter runs in a container
(the ID is set to the PID, which is typically 1 in a container), and
serveral blackbox_exporters are restarted at the same time (the sequence
numbers are reset to zero and stay in sync).

This commit sets the ICMP echo ID to a random value if the PID is 1, and
initializes the sequence number at a random offset.

See #411 for details.

### DIFF
--- a/prober/icmp.go
+++ b/prober/icmp.go
@@ -16,6 +16,7 @@ package prober
 import (
 	"bytes"
 	"context"
+	"math/rand"
 	"net"
 	"os"
 	"sync"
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

PR #413, opened 2019-01-31
Add metric successful_config_load

ev1lm0nk3y:


config/config.go
```diff
@@ -22,22 +25,34 @@ type SafeConfig struct {
 	C *Config
 }
 
+func init() {
+	configReloadSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "successful_config_load",
```
SuperQ: This should have a `Namespace` set to `blackbox_exporter`.

config/config.go
```diff
@@ -22,22 +25,34 @@ type SafeConfig struct {
...
+	})
+	prometheus.MustRegister(configReloadSuccess)
+}
+
 func (sc *SafeConfig) ReloadConfig(confFile string) (err error) {
 	var c = &Config{}
 
 	yamlFile, err := ioutil.ReadFile(confFile)
 	if err != nil {
+		configReloadSuccess.Set(0)
```
SuperQ: Rather than set at each return, we can use a defer to at the top of the function deal with the error output.

```go
defer func() {
  if err != nil {
    configReloadSuccess.Set(0)
  } else {
    configReloadSuccess.Set(1)
}()
```

config/config.go
```diff
@@ -22,22 +25,34 @@ type SafeConfig struct {
 	C *Config
 }
 
+func init() {
+	configReloadSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
```
SuperQ: This block can be moved up to line 17.

```go
var (
  configReloadSuccess = prometheus.NewGauge(
...
)
```

SuperQ:
I might also introduce a counter for counting all config load errors. `blackbox_exporter_config_load_errors_total`.

config/config.go
brian-brazil: config_last_reload_successful is the suffix we use in Prometheus, which we should align with

brian-brazil:
I'm not sure that is useful, however Prometheus has a prometheus_config_last_reload_success_timestamp_seconds

config/config.go
ev1lm0nk3y: done and done

config/config.go
ev1lm0nk3y: made this a tad cleaner

ev1lm0nk3y:
I agree that the `errors_total` is not that useful, but I'm not sure about the `timestamp_seconds` either. I was thinking about adding that metric, but then I read https://godoc.org/github.com/prometheus/client_golang/prometheus#NewMetricWithTimestamp and decided against it.

brian-brazil:
timestamp_seconds would have the value of the time, not a metric timestamp.

config/config.go
```diff
@@ -22,8 +37,20 @@ type SafeConfig struct {
...
 
+func init() {
+	prometheus.MustRegister(configReloadSuccess)
+	prometheus.MustRegister(configReloadSeconds)
+}
+
 func (sc *SafeConfig) ReloadConfig(confFile string) (err error) {
 	var c = &Config{}
+	defer func() {
+		configReloadSuccess.Set(0)
```
brian-brazil: There's a small race here, this should be the other branch of the if

ev1lm0nk3y:
yup, i just realized that. should be fixed now.

config/config.go
```diff
@@ -22,8 +37,21 @@ type SafeConfig struct {
...
+}
+
 func (sc *SafeConfig) ReloadConfig(confFile string) (err error) {
 	var c = &Config{}
+	defer func() {
+		if err != nil {
+			configReloadSuccess.Set(0)
+		} else {
+			configReloadSuccess.Set(1)
+			configReloadSeconds.Set(float64(time.Now().Unix()))
```
brian-brazil: You can use SetToCurrentTime for this.

brian-brazil:
Thanks!

PR #416, opened 2019-02-08
shrank the binary and image size

### LINKED ISSUE #352: MAINTAINER instruction in Dockerfile is deprecated
MAINTAINER should be replaced by LABEL maintainer="..."
See https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

nrvnrvn:
Changes:
- added a new ldflag to make the resulting binary a static ELF;
- removed debug symbols to make it smaller;
- switched to using `SCRATCH` as the base image;
- fixed #352 by coincidence.

### DIFF
--- a/.promu.yml
+++ b/.promu.yml
@@ -12,6 +12,7 @@ build:
         -X github.com/prometheus/common/version.Branch={{.Branch}}
         -X github.com/prometheus/common/version.BuildUser={{user}}@{{host}}
         -X github.com/prometheus/common/version.BuildDate={{date "20060102-15:04:05"}}
+        -s -w -extldflags "-static"
 tarball:
     files:
         - blackbox.yml
--- a/.promu.yml
+++ b/.promu.yml
@@ -7,12 +7,12 @@ repository:
 build:
     flags: -mod=vendor -a -tags netgo
     ldflags: |
+        -s
         -X github.com/prometheus/common/version.Version={{.Version}}
         -X github.com/prometheus/common/version.Revision={{.Revision}}
         -X github.com/prometheus/common/version.Branch={{.Branch}}
         -X github.com/prometheus/common/version.BuildUser={{user}}@{{host}}
         -X github.com/prometheus/common/version.BuildDate={{date "20060102-15:04:05"}}
-        -s -w -extldflags "-static"
 tarball:
     files:
         - blackbox.yml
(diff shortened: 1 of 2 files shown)


.promu.yml
```diff
@@ -12,6 +12,7 @@ build:
         -X github.com/prometheus/common/version.Branch={{.Branch}}
         -X github.com/prometheus/common/version.BuildUser={{user}}@{{host}}
         -X github.com/prometheus/common/version.BuildDate={{date "20060102-15:04:05"}}
+        -s -w -extldflags "-static"
```
SuperQ: This should probably match the upstream [promu config](https://github.com/prometheus/promu/blob/master/.promu.yml#L4-L6).

SuperQ:
I think we do want to keep the debug symbols. Also, I don't think switching to `FROM scratch` is appropriate to do here without a discussion on prometheus-developers@googlegroups.com first. We like to keep our various projects consistent, and this would be a change we would want to have the same everywhere.

SuperQ:
I'm pretty sure this is unecessary, our build default produces a static binary.

```
$ file blackbox_exporter
blackbox_exporter: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, not stripped
```

.promu.yml
nrvnrvn: yep, you were right about the default static. The config now matches the upstream.

SuperQ:
I don't see why this is necessary. This seems to eliminate our busybox image from the image, which means the usual busybox tools will not be available in the image.

nrvnrvn:
The reason is to keep the image as small as possible. When it goes for static linux binaries it is always tempting to use scratch as the base image. See [coredns](https://github.com/coredns/coredns/blob/master/Dockerfile) for instance. If for whatever reason you need to debug you can always inject busybox inside using `kubectl cp`/`docker cp`.

I hadn't looked into other Dockerfiles under https://github.com/prometheus until you mentioned consistency. This is a good point. I would not be happy to bring inconsistency to the official prometheus images. Now I see that most of your images use `https://github.com/prometheus/busybox` as the base image.

Feel free to close this PR and if you think it is worth raising the question of switching from `quay.io/prometheus/busybox:latest` to `scratch` wherever possible in prometheus-developers@googlegroups.com please let me know.

brian-brazil:
What's the size difference here?

SuperQ:
The uClibc busybox image is 1.46MiB.

SuperQ:
Yes, we should have a discussion first.

We already have an extremely compact busybox image. I'm a fan of `from scratch` containers, especially for Go. But I see the utility of having some basics in the container for inspection.

We could always propose having a `debug` type tag on Docker that contains the busybox image, but default to a scratch image.

Looking forward to the discussion.

PR #417, opened 2019-02-14
Add support for tcp probes which you expect to fail

alagendijk-minddistrict:
This PR adds the option expect_connection_fail to the tcp probe. So you can you use the tcp prober to check if ports are closed / unavailable.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -121,6 +121,9 @@ query_response:
 tls_config:
   [ <tls_config> ]
 
+# Fail if connection to port can be made.
+[ expect_connection_fail: <boolean> | default = false ]
+
 ```
 
 ### <dns_probe>
--- a/config/config.go
+++ b/config/config.go
@@ -102,12 +102,13 @@ type QueryResponse struct {
 }
 
 type TCPProbe struct {
-	IPProtocol         string           `yaml:"preferred_ip_protocol,omitempty"`
-	IPProtocolFallback bool             `yaml:"ip_protocol_fallback,omitempty"`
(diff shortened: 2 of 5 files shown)


brian-brazil:
Why not invert your check on probe_success?

alagendijk-minddistrict:
Because not being able to use a general `probe_success == 0`  check means a lot of filtering on labels in the checks. 
And then it can become confusing to know when you expect  `probe_success = 0` vs  `probe_success = 1`
IMO a user should be able to use the blackbox_exporter for different use cases, and this PR adds a option to check a port is indeed closed / unavailable, while still being able to use the great generic alerting rules Prometheus offers

brian-brazil:
> Because not being able to use a general probe_success == 0 check means a lot of filtering on labels in the checks.

That's not a very strong reason to me, you'd have to have different scrape jobs for this anyway to use a different module so there's not a lot of work there.

I'd also rather not add options for every possible way a probe can fail, e.g. DNS, TCP, TLS etc.

alagendijk-minddistrict:
I understand that you don't want to add options for all the ways a probe can fail, and I totally agree with you. But for those other ways (DNS, TLS, etc) I do not see a use case where you want the probe to fail.
Maybe it helps if I explain a little bit more about our use case:
We are using the blackbox_exporter from a node outside our network to blackbox test if all the services our users should have access to are available. But we also want to check that the services that should not be publicly available (shielded by VPN or firewalls), are indeed not available, this is where we are using this tcp probe for.

brian-brazil:
Maybe you want to check that a given name doesn't resolve, or that certain IPs aren't in use. I don't think this level of detail makes sense for the blackbox exporter.

PR #419, opened 2019-02-16
Add regexp matching of HTTP response headers to the http probe

gvsmirnov:
I needed to add the verification of header values for some of the targets probed by the blackbox exporter, but apparently matching is currently only supported for body content.

This pull request adds two new config options: `fail_if_header_matches_regexp` and `fail_if_header_not_matches_regexp`, mirroring the existing `fail_if_matches_regexp` and `fail_if_not_matches_regexp` for the response body. I considered renaming these ones to `fail_if_body_matches_regexp` and `fail_if_body_not_matches_regexp` for consistency, but that would be a breaking change and thus does not seem like a good idea.

I am fairly new to golang, so I just tried to follow the style and apparent conventions in the rest of this project. Please let me know if something should be changed.

I also tested this extensively manually, it worked in all the corner cases I tried.


gvsmirnov:
Oh dear, I just checked the list of open pull requests and found not one, but TWO others with the same changes: https://github.com/prometheus/blackbox_exporter/pull/62 and https://github.com/prometheus/blackbox_exporter/pull/332  😕 Apparently they were abandoned by their authors before all the comments were addressed.

Looking at the comments from @brian-brazil, it seems like all of them are addressed in this PR now. The "check that header is missing" can be achieved by

```
fail_if_header_matches_regexp:
  - header: Not-Supposed-To-Be-Here
    regexp: '.*'
```

config/config.go
```diff
@@ -80,19 +80,27 @@ type Module struct {
...
+	ValidHTTPVersions            []string                `yaml:"valid_http_versions,omitempty"`
+	IPProtocol                   string                  `yaml:"preferred_ip_protocol,omitempty"`
+	IPProtocolFallback           bool                    `yaml:"ip_protocol_fallback,omitempty"`
+	NoFollowRedirects            bool                    `yaml:"no_follow_redirects,omitempty"`
+	FailIfSSL                    bool                    `yaml:"fail_if_ssl,omitempty"`
+	FailIfNotSSL                 bool                    `yaml:"fail_if_not_ssl,omitempty"`
+	Method                       string                  `yaml:"method,omitempty"`
+	Headers                      map[string]string       `yaml:"headers,omitempty"`
+	FailIfMatchesRegexp          []string                `yaml:"fail_if_matches_regexp,omitempty"`
+	FailIfNotMatchesRegexp       []string                `yaml:"fail_if_not_matches_regexp,omitempty"`
```
gvsmirnov: Should these two above be renamed to `FailIfBodyMatchesRegexp` and `FailIfBodyNotMatchesRegexp` ?

SuperQ: Yes, I would rename them.

SuperQ:
Thanks for picking up where those other PRs went stale.

Per your above comment, I think you want the regexp to match `.+` in order to enforce some header text. Or maybe I'm misunderstanding.

```yaml
fail_if_header_matches_regexp:
  - header: Not-Supposed-To-Be-Here
    allow_missing: true
    regexp: '.+'
```

gvsmirnov:
>  I think you want the regexp to match `.+` in order to enforce some header text.
> EDIT: Are headers with no value after the header name valid?

If I read [RFC 7230 section 3.2](https://tools.ietf.org/html/rfc7230#section-3.2) correctly, header value is not optional, but an empty value is valid. That seems to be the generally accepted interpretation, judging by the discussions on the Internet that I found. However, `http.Header.Get()` returns an empty string (`""`) if the header is not set, which is ambiguous. Thankfully, the mapping can be accessed directly, so it's possible to just do something like `len(textproto.MIMEHeader(header)[key]) > 0`.

This leads to another consideration: as per section [section 3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2), there may be multiple header fields with the same field name. Judging by the doc on `http.Response.Header`, they may be concatenated with comma delimiters. It is not clear to me in which cases it would be populated by the http client as a list with multiple string values, and in which cases it would be in a single string value as a comma-separated list.

I would propose the following behaviour:

1. Verify the `allow_missing` by checking if the header with given name is present at least once. If one or more values are empty, do not consider it to be missing.
2. Verify the `regexp` against all the header values for a given header name.  For "fail if matches", at least one match fails the probe. For "fail if does not match", at least one mismatch fails the probe.
3. If multiple values are concatenated into a single comma-separated string, do not do any special handling. Let the user deal with such cases on their own.

I have doubts about the behaviour in (2) for the "fail if does not match" for multi-value headers. "fail if none match" may be more desirable than "fail if any does not match".

What do you think?

config/config.go
gvsmirnov: Done

gvsmirnov:
After some consideration, the most common use of multi-value headers is probably `Set-Cookie` (the RFC even mentions it explicitly). The following use cases seem to be plausible:

1. Verify that a specific cookie is set: the user would want the probe to fail if a desired cookie is not present, but would not want the probe to fail if there are other cookies:
```
fail_if_header_not_matches_regexp:
  header: Set-Cookie
  regexp: '.*Domain=\.example\.com.*'
```
2. Verify that no cookies are set
```
fail_if_header_matches_regexp:
  header: Set-Cookie
  allow_missing: true
  regexp: '.*'
```

Another common multi-value header is `Accept-Encoding`, but in practice it's mostly represented as a single comma-separated value, not as multiple values.

I pushed the changes implementing the behaviour described above. Please let me know if you think it is reasonable.

prober/http.go
```diff
@@ -173,6 +236,11 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Help: "Indicates if probe failed due to regex",
 		})
 
+		probeFailedDueToHeaders = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_failed_due_to_headers",
```
brian-brazil: probe_http_

falied_due_regex should already cover this

prober/http.go
```diff
@@ -69,6 +70,68 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+		for _, val := range values {
+			if re.MatchString(val) {
+				level.Error(logger).Log("msg", "Header matched regular expression", "header", headerMatchSpec.Header,
+					"regexp", headerMatchSpec.Regexp, "value_count", len(values))
+				return false
+			}
+		}
+	}
+	for _, headerMatchSpec := range httpConfig.FailIfHeaderNotMatchesRegexp {
+		values := textproto.MIMEHeader(header)[headerMatchSpec.Header]
```
brian-brazil: This won't canonicalise the key

prober/http.go
gvsmirnov: > `probe_http`
Thee are several other labels that do not have the `http`: `probe_ssl_earliest_cert_expiry` and `probe_failed_due_to_regex`. Should they be changed, too? Or can I just ditch the newly introduced `probe_failed_due_to_headers` and keep the rest as is?

prober/http.go
gvsmirnov: Fixed and added tests

gvsmirnov:
> It's probably worth it.

Alright, done. The breaking change would be caught at the configuration parsing stage, so it's a fail-fast scenario.

prober/http.go
brian-brazil: They should be changed, but at least the ssl one is commonly used for alerting so we need to be a little careful so not now.

config/config.go
```diff
@@ -80,19 +80,27 @@ type Module struct {
...
+	IPProtocol                   string                  `yaml:"preferred_ip_protocol,omitempty"`
+	IPProtocolFallback           bool                    `yaml:"ip_protocol_fallback,omitempty"`
+	NoFollowRedirects            bool                    `yaml:"no_follow_redirects,omitempty"`
+	FailIfSSL                    bool                    `yaml:"fail_if_ssl,omitempty"`
+	FailIfNotSSL                 bool                    `yaml:"fail_if_not_ssl,omitempty"`
+	Method                       string                  `yaml:"method,omitempty"`
+	Headers                      map[string]string       `yaml:"headers,omitempty"`
+	FailIfBodyMatchesRegexp      []string                `yaml:"fail_if_body_matches_regexp,omitempty"`
+	FailIfBodyNotMatchesRegexp   []string                `yaml:"fail_if_body_not_matches_regexp,omitempty"`
+	FailIfHeaderMatchesRegexp    []HeaderMatch           `yaml:"fail_if_header_matches_regexp,omitempty"`
```
brian-brazil: This isn't a string, so I'd remove the _regexp from the name to avoid confusion.

gvsmirnov: Where would the confusion arise from? Is there a convention that properties ending with `_regexp` are always strings that represent regular expression?

I could rename the properties so it would be like this:

```yml
# ...
fail_if_body_matches_regexp:
  - 'Thou shall not pass'
fail_if_header_matches:
  - header: Host
    regexp: example.com
```

It would lose the consistency between property names for header and body, though. Is it worth it?

prober/http.go
gvsmirnov: OK, I will not touch the existing labels.

But I am not sure if representing the failed header matches with `probe_failed_due_to_regex` is the right way. It would be helpful if the user could see immediately from the metrics where the failure originated from. I do not see any drawbacks to this either (the extra disk usage would be negligible).

If you do insist, I can reuse `probe_failed_due_to_regex`. Do you insist?

config/config.go
brian-brazil: Yes, the data structures are different

prober/http.go
brian-brazil: I'd suggest reusing, either that or have two more clearly named metrics.

gvsmirnov: OK, reused existing metric. Renaming existing ones would be a more dangerous breaking change indeed.

config/config.go
gvsmirnov: Done

gvsmirnov:
@brian-brazil @SuperQ is there anything else I can do to get this merged? I would love to use the upstream version instead of own fork.

CONFIGURATION.md
```diff
@@ -58,14 +58,30 @@ The other placeholders are specified separately.
...
+  fail_if_body_matches_regexp:
     [ - <regex>, ... ]
 
-  # Probe fails if response does not match regex.
-  fail_if_not_matches_regexp:
+  # Probe fails if response body does not match regex.
+  fail_if_body_not_matches_regexp:
     [ - <regex>, ... ]
 
+  # Probe fails if response header matches regex. For headers with multiple values, fails if *at least one* matches
```
brian-brazil: Full stop

CONFIGURATION.md
```diff
@@ -58,14 +58,30 @@ The other placeholders are specified separately.
...
+  fail_if_header_matches:
+    [ - [ header: <string>,
+          [ regex: <regex>, ]
```
brian-brazil: Indentation seems off here

prober/http.go
```diff
@@ -69,6 +70,68 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
 
+func matchRegularExpressionsOnHeaders(header http.Header, httpConfig config.HTTPProbe, logger log.Logger) bool {
+	for _, headerMatchSpec := range httpConfig.FailIfHeaderMatchesRegexp {
+		values := header[textproto.CanonicalMIMEHeaderKey(headerMatchSpec.Header)]
+		if len(values) == 0 {
+			if !headerMatchSpec.AllowMissing {
+				level.Error(logger).Log("msg", "Missing required header", "header", headerMatchSpec.Header)
+				return false
+			} else {
+				continue // no need to match any regex on missing headers
```
brian-brazil: "No", and  full stop

CONFIGURATION.md
gvsmirnov: Yeah, does not look quite good. I got it from the prometheus config documentation, e.g. [here](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)

When the same level contains both optional and non-optional parameters, then it looks like this (comments stripped):

```yml
job_name: <job_name>
[ scrape_interval: <duration> | default = <global_config.scrape_interval> ]
```

In here, though, the `regex` should be named `regexp` and be made non-optional, so it would look like this:

```yml
  # Probe fails if response header matches regex. For headers with multiple values, fails if *at least one* matches.
  fail_if_header_matches:
    [ - [ header: <string>,
          regexp: <regex>,
          [ allow_missing: <boolean> | default = false ]
        ], ...
    ]
```

Would that be OK?

brian-brazil: Looking around, the only similar one we have is some of pagerduty for the AM. There we break it out as a different type.

gvsmirnov: Another example right in this repo is `tcp_probe`, it does not break it out as a separate type. That's why I did not do it here. Can extract both or keep inlined like above

brian-brazil: For query_response they're all optional though.

gvsmirnov: OK, extracted only the http header match spec.

CONFIGURATION.md
gvsmirnov: Added here and in other places

prober/http.go
gvsmirnov: Fixed here and in the other place.

gvsmirnov:
Huh, travis build failed when fetching dependencies. Is it possible to rerun it?

gvsmirnov:
@brian-brazil @SuperQ sorry to pester, but I ask again: anything else I can do to get this merged?

CONFIGURATION.md
```diff
@@ -86,14 +94,22 @@ The other placeholders are specified separately.
...
   body: [ <string> ]
 
 
 ```
 
+#### <http_header_match_spec>
+
+```yml
+header: <string>,
+regexp: <regex>,
```
brian-brazil: This is optional

gvsmirnov: Why is it optional? It is not possible to match a header without a regex to match against.

brian-brazil: Unless you allow it to be missing

gvsmirnov: But if you allow it to be missing, then what is the point of having a match rule at all? 

```yml
fail_if_header_matches:
  - header: Transfer-Encoding
     allow_missing: true
fail_if_header_not_matches:
  - header: Transfer-Encoding
     allow_missing: true
```

Neither of these rules make sense to me. But looking at the checks made in `config.go`, apparently they used to. So I think I should change the config check, so it always required `regexp` to be set, regardless of `allow_missing`.

Would that be OK?

gvsmirnov: In a stroke of speculative execution, did just that.

brian-brazil:
Thanks!

gvsmirnov:
Thanks for a thorough code review! I appreciate it.

Is there any estimate of when the next version will be released? I’m fine building from upstream source, asking for the sake of completeness.

brian-brazil:
It's not going to be today, and there's a breaking change in there so Friday is a bad idea. Probably early next week.

piotrkochan:
So now example.yml should be fixed, because for "http" prober option "fail_if_matches_regexp" fail and this is very confusing:

```
msg="Error loading config" err="Error parsing config file: yaml: unmarshal errors:\n  line 8: field fail_if_body_matches_regexp not found in type config.plain"
```

gvsmirnov:
@piotrkochan how should it be fixed? `example.yml` is in line with the renamed config parameter. It used to be `fail_if_matches_regexp`, now it is `fail_if_body_matches_regexp`.

Where do you get this error? I would guess that you are running the old version of `blackbox_exporter` with the new version of config. However, the new version of `blackbox_exporter` has not yet been released, so the documentation is out of sync with the available distributions.

@brian-brazil any way I can help with making a release? This is indeed confusing now.

brian-brazil:
https://github.com/prometheus/blackbox_exporter/pull/382#issuecomment-467026887 is what's holding a release up.

PR #430, opened 2019-03-09
added NTP probe

ac427:


brian-brazil:
I'd prefer that we look at generic UDP support, rather than having code for each specific protocol using it. In addition NTP is not something to hit too often.

rigtorp:
I have a use case for this. I currently use Nagios to check that all my switches, firewalls and servers are correctly synched to my GNSS grandmaster clock. Once every hour it checks that NTP servers on all the devices responds with a time within 100ms of the grandmaster clock. This is really to catch configuration errors. How would I be able to do this with the generic UDP support?

SuperQ:
@rigtorp I think you'd be better off using a stand-alone NTP proxy exporter. I don't know if one exists right now, but it would be easy enough to build.  You might base it on the blackbox_exporter and the node_exporter's [ntp module](https://github.com/prometheus/node_exporter/blob/master/collector/ntp.go).

rigtorp:
Sure, but wouldn't it be more useful for the prometheus ecosystem to provide batteries included blackbox monitoring solution? For example another useful and common probe is to check if tracepath output changed to see if BGP was screwed up or MTU is set incorrectly.

SuperQ:
In Prometheus ecosystem, we prefer smaller, sharper tools rather than all-inclusive bloated agents. :smile:

Tracepath probing also sounds like something much more specialized.

rigtorp:
Yes it's one of the issues I find with using Prometheus for non-application metrics: I need to write all the stuff that comes out of the box in Nagios, Zabbix, commercial tools myself. Now these tools has other downsides of course which is why i'm looking to replace them :).

Tracepath is fairly standard blackbox probing when maintaining large networks.

SuperQ:
Yea, there are lots of corners of the monitoring world we haven't written yet. We started the project from the application monitoring world. From that perspective we have a lot of good coverage. Some stuff we do have good optisons. For example, I wrote a [SmokePing replacement](https://github.com/SuperQ/smokeping_prober) that, IMO, produces better data than smokeping itself.

There are some large network operators out there using Prometheus, getting them working together would help with this.

On the other hand, I think we're _way_ ahead in terms of translating SNMP data into something sane.

rigtorp:
@SuperQ You're smokeping replacement looks good.

My wish is that there was some defacto standard Prometheus exporter for network reachability, path MTU etc. I guess blackbox_exporter will not be that (despite the name). 

Yes getting SNMP and netflow data into Prometheus would also be nice.

SuperQ:
The snmp_exporter is great. Also check out [tflow2](https://github.com/bio-routing/tflow2).

rigtorp:
Thanks you for your comments. It seems to me that there is more information and tools needed regarding how to use Prometheus for network monitoring.

brian-brazil:
As discussed above, there's better ways to do this elsewhere.

PR #440, opened 2019-03-14
Empty in YAML means null.

brian-brazil:
The null then overrides the default config. Avoid this
in the example config file.


juliusv:
:+1:

PR #451, opened 2019-03-27
Docker image generation not requiring building the project outside docker

dcaba:
Introducing multistage build in the main dockerfile so we don't depend on an external build to generate a dockered version of the software

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,14 @@
-FROM        quay.io/prometheus/busybox:latest
+FROM golang:1.12 as builder
+
+WORKDIR /go/src/github.com/prometheus/blackbox_exporter
+ADD . .
+RUN make
+
+FROM quay.io/prometheus/busybox:latest
 LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com>"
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
This file is part of our build system, if you want to change how it all works the promu repo is where to propose that.

dcaba:
Thanks for the info... I see your build system is created with the assumption the binary its build outside docker, and then packaged in all formats you support, including docker, to ensure the binary is the same, so that's why you are not doing a separate build for docker directly in your dockerfile, that could to external users, as it would be easier to build and get an image ready, just with a single command... just to facilitate that use case: what do you think if we just add to the readme that we can run this software by pulling your already build docker image? 

[https://hub.docker.com/r/prom/blackbox-exporter/tags](https://hub.docker.com/r/prom/blackbox-exporter/tags)

Happy to submit a proposal

brian-brazil:
The readme is a bit outdated, using a released build should be first.

dcaba:
Proposal [here](https://github.com/prometheus/blackbox_exporter/pull/452) only covering README updates. Closing this PR. Thanks!

PR #452, opened 2019-03-27
README update so run (without build) instructions are clear

dcaba:
Separating build and run instructions, so users can just download the released software, or opt for the public docker container the project is already maintaining


brian-brazil:
Thanks!

PR #458, opened 2019-04-10
Use timeout for chooseProtocol

thorfour:
If an address DNS resolve takes a while, a probe can easily exceed its timeout limit.

This threads the given context to the probe into the `chooseProtocol` func to exit lookup of the IP address upon timeout.


brian-brazil:
Could you use one of the Resolver methods that take a context?

thorfour:
That's what I initially looked into doing, but couldn't find a Resolver function that supported both protocol, and context. [LookupIPAddr](https://golang.org/pkg/net/#Resolver.LookupIPAddr) doesn't take the protocol, but will return ipv4 and ipv6, which changes the behavior of the chooseProtocol function some. 

And there's currently `// TODO(bradfitz): Timeout time.Duration?` to add timeout to the resolver in general?

brian-brazil:
We could filter the result of LookupIPAddr.

thorfour:
If that's an acceptable change I'd be happy to make that.

thorfour:
I've made the changes to use LookupIPAddr and return the requested IP protocol, or fallback if necessary.

thorfour:
Is travis-ci a flake? Unable to get failures locally

thorfour:
Doesn't appear to be a flake. I wonder if LookupIPAddr wont work since this technically changes the underlying behavior and does an ipv6 lookup too?

prober/utils.go
```diff
@@ -60,31 +55,45 @@ func chooseProtocol(IPProtocol string, fallbackIPProtocol bool, target string, r
...
+
+			// ip6 as fallback
+			fallback = ip
 		}
 	}
 
-	if ip.IP.To4() == nil {
-		probeIPProtocolGauge.Set(6)
-	} else {
+	// Use fallback ip protocol
```
brian-brazil: Shouldn't you check that fallback has been set?

thorfour: Good call

thorfour:
Failure seems to be unique to Travis. I'm unable to get it to fail on the other platforms I've tried

prober/utils.go
```diff
@@ -60,31 +56,50 @@ func chooseProtocol(IPProtocol string, fallbackIPProtocol bool, target string, r
...
-		}
+		level.Error(logger).Log("msg", "Resolution with IP protocol failed", "err", err)
+		return nil, 0.0, err
+	}
 
-		if err != nil {
-			if IPProtocol == "ip6" {
-				probeIPProtocolGauge.Set(6)
-			} else {
+	// Return the IP in the requested protocol format
```
brian-brazil: Full stop. s/format//.

thorfour:
Since Travis doesn't support v6 and it's technically changing the behavior of the code to use the LookupIPAddr I've reverted it to the wrapped ctx method. Hopefully in the future the net package will become more mature and make this easier.

brian-brazil:
The problem with that though is that we don't actually stop the request. I'd prefer the LookupIPAddr method, and then do whatever to get Travis to pass.

thorfour:
Okay, I moved it back to use LookupIPAddr, and skipped the 4 failing tests on the travis environment

brian-brazil:
Thanks!

Issue #147, opened 2017-05-15
rootless ICMP ping


darkk:
[x/net/icmp](https://godoc.org/golang.org/x/net/icmp#ListenPacket) supports root-less operation for ICMP pings on Linux and MacOSX, but blackbox_exporter requires elevated privileges for that.

Are there any non-obvious blockers for using rootless ping sockets? I've looked at the code and I've not noticed any.

brian-brazil:
Those docs seem to be confusing UDP and ICMP.

darkk:
Maybe the doc is wrong, maybe it says "udp" as IPPROTO_ICMP socket is created with SOCK_DGRAM (like UDP socket), but golang/go#9166 explicitly tells that the feature is implemented.

brian-brazil:
The docs are definitely wrong, the only reference to ICMP is for privileged sockets which means root.

I think we need more clarity here, and knowing which kernels support this.

darkk:
> which kernels support this
AFAIK, it's mainlined [since v2.6.39](http://openwall.info/wiki/people/segoon/ping).

I have not figured out earliest MacOSX version supporting `socket(AF_INET, SOCK_DGRAM, IPPROTO_ICMP)` but it was already supported when the patch was merged to Linux mainline.

brian-brazil:
So introduced just under 6 years ago, that's relatively new as kernel features go.

darkk:
> The docs are definitely wrong

The docs match test code at https://github.com/golang/net/blob/master/icmp/ping_test.go#L61

brian-brazil:
The docs indicate you need privileged access to use this feature: "For privileged raw ICMP endpoints, network must be "ip4" or "ip6" followed by a colon and an ICMP protocol number or name."

darkk:
The part of the doc that is relevant to the ticket is located a couple of paragraphs above:

>  For **non-privileged** datagram-oriented ICMP endpoints, network must be "udp4" or "udp6". The endpoint allows to read, write a few **limited ICMP messages such as echo request and echo reply**. Currently only Darwin and Linux support this. 

It also needs some privileges, running process should be in the group within `net.ipv4.ping_group_range`, but it's much lower amount of privileges than `cap_net_raw` capability.

[...]

PR #462, opened 2019-04-17
Do not run as root by default in Docker

der-eismann:
Similar in spirit to prometheus/statsd_exporter#202 and others.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com
 COPY blackbox_exporter  /bin/blackbox_exporter
 COPY blackbox.yml       /etc/blackbox_exporter/config.yml
 
+USER        nobody
 EXPOSE      9115
 ENTRYPOINT  [ "/bin/blackbox_exporter" ]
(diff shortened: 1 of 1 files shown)


SuperQ:
Looks like this is missing the DCO. You can `git commit -s --amend` to add it.

der-eismann:
Just noticed it, it's added now :slightly_smiling_face:

brian-brazil:
We may actually want root here, for ICMP.

der-eismann:
Can't we workaround this with setuid `chmod u+s` or something like that?

brian-brazil:
That makes no difference, it'd still be running as root.

n0guest:
> We may actually want root here, for ICMP.

Maybe [setcap](https://github.com/docker/labs/tree/master/security/capabilities#step-4-extra-for-experts) would help? I.e:

```
RUN setcap cap_net_raw+ep /path/to/binary
```

brian-brazil:
That's an option.

SuperQ:
`setcap` is a good option. We're still waiting for someone to implement https://github.com/prometheus/blackbox_exporter/issues/147 for root-free operation.

PR #479, opened 2019-05-22
Switch from quay.io busybox to docker hub

agolomoodysaada:
There are several DNS issues related to busybox versions 1.28.4 and above. It seems like quay.io has not updated its latest image in 2 months. In order to get bugfixes sooner, it's better to use the up to date versions on docker hub.
More info: https://github.com/docker-library/busybox/issues/48

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM        quay.io/prometheus/busybox:latest
+FROM  busybox:latest
 LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com>"
 
 COPY blackbox_exporter  /bin/blackbox_exporter


brian-brazil:
We've many repos which we're keeping in sync, so if you want to propose changing this you should do so in the prometheus repo which acts as our canonical version.

@SuperQ @simonpasquier FYI

simonpasquier:
`quay.io/prometheus/busybox` is built from this [repository](https://github.com/prometheus/busybox) which uses the upstream busybox image and adds a few files that are deemed necessary for the various Prometheus projects. That being said, we could have a CI job that periodically rebuilds those images whenever a new tag exists for the [upstream](https://hub.docker.com/_/busybox/) image.

SuperQ:
@simonpasquier Yea, another use case for a CircleCI cron job. :)

SuperQ:
I filed https://github.com/prometheus/busybox/issues/28 to track the feature request.

Issue #442, opened 2019-03-18
Add "probe_ssl_cert_expiry" to have expiration date of the actual certificate as in Chrome/openssl (not the earliest expiration date as we have now in "probe_ssl_earliest_cert_expiry" metric)


dkutetsky:
Related to the issue - https://github.com/prometheus/blackbox_exporter/issues/340

brian-brazil:
Did you mean to make a PR?

dkutetsky:
Actually, it was a feature request, not sure how it should be marked properly.

brian-brazil:
This is a duplicate then.

pir8g33k:
so any solution on this it seems we have the same issue 

@brian Brazil it seems the same issue i saw on repo, any other solution how to display the actual information of certificate expiration date ? we need this info as well as we monitor the remaining days of the ssl certificate
https://github.com/prometheus/blackbox_exporter/issues/340
https://github.com/prometheus/blackbox_exporter/pull/483
https://github.com/prometheus/blackbox_exporter/issues/442

dkutetsky:
Hi @pir8g33k ,
You can use [fork with our fix](https://github.com/greg-solutions/blackbox_exporter/tree/feature/extend_cert_expire_data).

pir8g33k:
> Hi @pir8g33k ,
> You can use [fork with our fix](https://github.com/greg-solutions/blackbox_exporter/tree/feature/extend_cert_expire_data).

Alright thank you so much @dkutetsky

Issue #340, opened 2018-07-03
Wrong probe_ssl_earliest_cert_expiry value for certificates with multiple trust chains


oleg-glushak:
We try to monitor cert expiration date for this certificate: tvonline.swb-gruppe.de
With openssl and using Google Chrome browser we get an expected value (Oct 26 23:59:59 2019).

`echo | openssl s_client -servername tvonline.swb-gruppe.de -connect tvonline.swb-gruppe.de:443 2>/dev/null | openssl x509 -noout -dates
notBefore=Oct 26 00:00:00 2017 GMT
notAfter=Oct 26 23:59:59 2019 GMT
`

But with blackbox_exporter we get:

`probe_ssl_earliest_cert_expiry 1.534824e+09`

which is GMT: Tuesday, 21 August 2018 04:00:00.

It seems a similar problem is described here:
https://security.stackexchange.com/questions/66487/what-happens-when-certificates-further-up-the-chain-expires-before-mine-equifa

It's recommended to update openssl to v1.0.2 there which has a fix for this issue but I guess golang/blackbox_exporter use some other mechanism to work with SSL. Is there any workaround or fix for this issue?

brian-brazil:
The metric purposefully works off the cert that expires first, so that you don't get caught out by something in your chain expiring before the main cert. It doesn't understand multiple trust chains, and it's not clear to me if we should change this.

VarunBatraIT:
Just signing in to confirm that I also faced the same problem. Our App, went off for about 3 hours when someone reported the problem to us. Then instead of wildcard-certificate, we shifted to normal one.

dkutetsky:
We have the same issue when Chrome and OpenSSL shows one date, but blackbox shows date earlier. 
How it could be fixed or can we use any workaround for this?

dkutetsky:
It will be great to have another metric to return probe_ssl_cert_expiry to have the same date that we have now in all browsers/openssl.

jrdemasi:
So turns out this is definitely still an issue - has there been any discussion in terms of how to move forward?  The behavior as-is is basically useless.

ribbybibby:
For those of you interested in only alerting on specific certificates in the chain, that can be achieved with: https://github.com/ribbybibby/ssl_exporter.

mr-karan:
https://thesslonline.com/blog/sectigo-addtrust-external-ca-root-expiring-may-30-2020

[...]

PR #483, opened 2019-06-04
Extend information about last cert of chain

vadimgs:
Add to registry information about expiration time fro last certs of chain
probe_ssl_last_cert_expiry

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -225,6 +225,10 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Name: "probe_ssl_earliest_cert_expiry",
 			Help: "Returns earliest SSL cert expiry in unixtime",
 		})
+		probeSSLLastCertExpiry = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_ssl_last_cert_expiry",
+			Help: "Returns last SSL cert expiry date",
+		})
 
 		probeHTTPVersionGauge = prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: "probe_http_version",
@@ -490,7 +494,9 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	if resp.TLS != nil {
 		isSSLGauge.Set(float64(1))
 		registry.MustRegister(probeSSLEarliestCertExpiryGauge)
+		registry.MustRegister(probeSSLLastCertExpiry)
 		probeSSLEarliestCertExpiryGauge.Set(float64(getEarliestCertExpiry(resp.TLS).Unix()))
+		probeSSLLastCertExpiry.Set(float64(getLastCertExpiry(resp.TLS).Unix()))
 		if httpConfig.FailIfSSL {
 			level.Error(logger).Log("msg", "Final request was over SSL")
 			success = false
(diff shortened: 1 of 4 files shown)


brian-brazil:
How is this useful?

vadimgs:
@brian-brazil This is necessary to obtain information on the latest certificate. Since it is necessary in the chain of certificates

brian-brazil:
How does this differ from what we already have?

dkutetsky:
> How does this differ from what we already have?

This is a fix for the issue https://github.com/prometheus/blackbox_exporter/issues/340.
Currently, blackbox returns the earliest expiration in a chain, but in the real world, we need an another behavior - we should get expiration date of the last certificate in the chain (should be the same as in OpenSSL/Chrome).
So, @vadimgs extends the result with another field that is will be useful for getting the same results we had in browsers.
I described exactly implemented behavior here -https://github.com/prometheus/blackbox_exporter/issues/442

brian-brazil:
This doesn't sound like a fix for #340, as you'd (presumably) get an arbitrary cert rather than the earliest chain.

dkutetsky:
I will get exactly site certificate expiration date - we tested it with our case and it works fine :)

algestam:
+1

The last cert in the chain is usually the site/domain specific certificate. Information on the expiry date for this certificate is useful since it is the often the only one in the chain that needs to be re-ordered from the certificate issuer before the validation end date.

brian-brazil:
However if an intermediate cert is going to expire first, you want to know about that.

algestam:
Definitely! Both properties are useful but for different purposes. The earliest expire property gives the time until the earliest expiry in the chain and the new property (last in chain) is useful for planning when a new certificate is to be ordered.

brian-brazil:
In both cases you need to take action, so this new metric wouldn't add anything.

algestam:
I don't agree there :) I have a certificate that is issued to expire in ~2 years. The cert chain however contains an AddTrust External CA Root certificate that expires on May 30th 2020. There is however no action needed from my end as I understand it based on this: https://www.xolphin.com/support/Rootcertificates/Phasing_out_Addtrust_External_CA_Root_certificate

The earliest expiry property correctly returns the time until the earliest expiry which is the AddTrust certificate but I'm also interested in when my site certificate (last cert in chain) expires so that a new cert is ordered in time.

brian-brazil:
You're assuming that everyone in the chain is being diligent, and checking that it's all okay is a human action.

algestam:
I will still continue to use the earliest expiry property to monitor the overall validity of the cert but the last expiry property complements my needs since that is the only certificate in my chain that I need to manually re-order when the expiry date closes in.

brian-brazil:
This isn't a fix for #340, and I'm not seeing how this helps in general.

pir8g33k:
> > How does this differ from what we already have?
> 
> This is a fix for the issue #340.
> Currently, blackbox returns the earliest expiration in a chain, but in the real world, we need another behavior - we should get expiration date of the last certificate in the chain (should be the same as in OpenSSL/Chrome).
> So, @vadimgs expanded the result with another field (probe_ssl_last_cert_expiry) that is will be useful for getting the same results we had in browsers.
> I described exactly implemented behavior here -#442

are you able to resolved or find a solution for this one sir? @dkutetsky

PR #487, opened 2019-06-13
Allow TLS renegotiation

meyergru:
This patch is to address an interop issue with Microsoft IIS when using client certificates. Without this, there will always be an error because TLS renegotiation is disabled by default. This has been addressed in Go a while ago, by implementing the 'Renegotiation' parameter.


brian-brazil:
You shouldn't change vendored files directly, you need to change it in the common repo first. Note that there's work to do before we can re-vendor that repo.

meyergru:
You are correct, I did not know that. I created a pull request in the upstream repo...

PR #489, opened 2019-06-17
Update prometheus/client_golang to v1.0.0

beorn7:
Signed-off-by: beorn7 <beorn@grafana.com>


beorn7:
Hmm, I guess update of dependencies has been in limbo for a while now (see a bunch of other PRs and branches). Is that still because of the HTTP keep-alive? @brian-brazil do you have plans to move forward from the status quo?

brian-brazil:
Yes, it's due to keepalive. There is https://github.com/prometheus/common/pull/133, but it's stale.

beorn7:
I'll update this to use common v0.6.0 and will try to fix the other issues.

brian-brazil:
I've literally just finished that.

beorn7:
Even better. Thanks!

PR #492, opened 2019-06-20
Do not apply offsets for explicit timeouts

### LINKED ISSUE #491: Offset should not be applied for explicitly set timeout
As per the following code, blackbox_exporter always applies `--timeout-offset` (defaulting to `0.5s`) to the timeout:

https://github.com/prometheus/blackbox_exporter/blob/dc53582740bcf20f1a618c806745a91c2d894345/main.go#L75-L92

There's a problem with this setting in the following cases:
- `--timeout-offset=0`: Prometheus might fail to scrape blackbox_exporter as the timeout (when combined with the network latency) might exceed Prometheus scrape timeout (that's the reason it's not `0` by default).
- `--timeout-offset=0.5`: it's not possible to have HTTP modules that expect response faster than `0.5` (as `timeout: 500ms` effectively becomes `timeout: 0ms`).

Would it be possible to apply the offset (`timeoutSeconds -= *timeoutOffset `) *only* when there's no explicitly defined timeout for a module? Meaning, that if the user explicitly set `timeout` field (which, as per the conditions, must be more than 0 and less than Prometheus scrape timeout) for some specific module, we can assume the user knows what he's doing and we use this timeout as is (without applying any offset).

An alternative can also be having per-module `timeout-offset` property, which allows the module to override not only the default time but default offset as well.

I can do a PR if agreed.

tdabasinskas:
Fixes #491 and adds an additional test.


main.go
```diff
@@ -355,3 +343,25 @@ func run() int {
...
+		timeoutSeconds, err = strconv.ParseFloat(v, 64)
+		if err != nil {
+			return 0, err
+		}
+	}
+	if timeoutSeconds == 0 {
+		timeoutSeconds = 10
+	}
+
+	if module.Timeout.Seconds() < timeoutSeconds && module.Timeout.Seconds() > 0 {
```
brian-brazil: Shouldn't we do the subtraction before this?

tdabasinskas: Not sure.

As per the linked issue, the idea is that `offset` has no effect in case the user explicitly set per-module timeout. So, we only do subtraction if there's no explicitly set timeout. The requirement of the specified timeout not exceeding Prometheus scrape-timeout is also met (again, without taking offset into account then).

brian-brazil: We should still take the offset into account when comparing against the scrape timeout, otherwise we're not allowing for network time to the exporter.

tdabasinskas: Would `if module.Timeout.Seconds() < (timeoutSeconds - offset) && module.Timeout.Seconds() > 0 {` be OK?

brian-brazil: Yes, but it'd be clearer to do the subtraction before the if.

tdabasinskas: Done 👌

brian-brazil:
Thanks!

PR #496, opened 2019-07-02
Build ARM container images

simonpasquier:
It also updates the Circle CI configuration to use the Prometheus orb.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,12 +1,8 @@
 ---
 version: 2.1
 
-executors:
-  # Whenever the Go version is updated here, .travis.yml and .promu.yml
-  # should also be updated.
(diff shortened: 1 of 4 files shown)


brian-brazil:
:+1:

Lekensteyn:
Is the Dockerfile supposed to be used by others? Right now `make common-docker` (or `docker build -t foo .`) fails because the default `common-build` target puts the binary in `blackbox_exporter` as opposed to somewhere in `.build/`.

simonpasquier:
@Lekensteyn I can't find the original issue but the idea was that `make build` would build the binary at the root directory and copy/link it to `.build/...`. For now you could do `promu crossbuild linux/amd64` then `make docker`.

PR #504, opened 2019-07-31
updated config docs for no_follow_redirects option

wtip:
For example the following would need to be used.

```
prober: http
http:
  valid_status_codes:
  - 200
  - 202
  - 301
  - 302
  no_follow_redirects: true
```

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -49,7 +49,7 @@ The other placeholders are specified separately.
   headers:
     [ <string>: <string> ... ]
 
-  # Whether or not the probe will follow any redirects.
+  # Whether or not the probe will follow any redirects. You will also need to use the valid_status_codes option to allow 301 and 302
   [ no_follow_redirects: <boolean> | default = false ]
 
   # Probe fails if SSL is present.


brian-brazil:
Thanks for your PR. These are reference docs, and your suggestion isn't correct in all cases.

wtip:
@brian-brazil When the `no_follow_redirects: true` option is used is there a case where the probe can succeed when the `valid_status_codes` don't include the  `302` code and the server being probed returns a 302 redirect?

Is this reference doc not the correct place to document this example or are you just looking for language that covers all cases?
For example when `no_follow_redirects = false` my comment of `You will also need to use the valid_status_codes option to allow 301 and 302` is incorrect.
Would the following be better?
`If true, you might also need to use the valid_status_codes option to allow 301 and 302`

brian-brazil:
> is there a case where the probe can succeed when the valid_status_codes don't include the 302 code and the server being probed returns a 302 redirect?

The cases are when it doesn't return a 302, such as testing that there isn't a redirect.

> Is this reference doc not the correct place to document this example

What you're proposing are usage guides, not reference docs.

wtip:
Ok so maybe this would be better for the `example.yml` file or would the wiki be better?

PR #509, opened 2019-08-16
Bump default timeout to 2m

### LINKED ISSUE #259: Confusing timeout behaviour
I was surprised by the behaviour of timeouts when troubleshooting an internal healthcheck for a very slow service with configuration like this:

```
    https_2xx:
        prober: http
        timeout: 15s
        http:
            fail_if_not_ssl: true
            preferred_ip_protocol: "ip4"
```

The first problem was that the timeout values are not applied by default:
```
$ curl 'http://localhost:9999/probe?module=https_2xx&target=https://example.org/very/slow/healthcheck'
…
probe_duration_seconds 9.500304772
```

Note that while the configuration specifies 15s the actual timeout was just under the default 10 seconds. Adding `-H X-Prometheus-Scrape-Timeout-Seconds:120` allows reaching the configured timeout:

```
probe_duration_seconds 14.500166973
```

brian-brazil:
The current default of 10s is problematic when testing slower probes
from a browser. Cancellation will still work as a fallback,
including for older Prometheus servers that don't send the timeout
header, so there shouldn't be a big buildup of ongoing probes.

Fixes #259

### DIFF
--- a/main.go
+++ b/main.go
@@ -354,7 +354,7 @@ func getTimeout(r *http.Request, module config.Module, offset float64) (timeoutS
 		}
 	}
 	if timeoutSeconds == 0 {
-		timeoutSeconds = 10
+		timeoutSeconds = 120
 	}
 
 	var maxTimeoutSeconds = timeoutSeconds - offset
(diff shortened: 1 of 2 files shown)


juliusv:
:+1:

Issue #475, opened 2019-05-07
Add support for --web.route-prefix/external-url


agrrh:
### Host operating system:

Not applicable.

### blackbox_exporter version: output of `blackbox_exporter -version`

Current v0.14.0 version is affected:
https://github.com/prometheus/blackbox_exporter/blob/e7e450c/main.go#L283-L286

### What is the blackbox.yml module config.

Not applicable.

### What is the prometheus.yml scrape config.

Not applicable.

### What logging output did you get from adding `&debug=true` to the probe URL?

Not applicable.

### What did you do that produced an error?

I'm running blackbox_exporter under following address:

http://example.org/exporters/blackbox/

### What did you expect to see?

I would like index page links lead me to correct URLs, like:

http://example.org/exporters/blackbox/metrics

### What did you see instead?

Broken URLs:

http://example.org/metrics

PR #515, opened 2019-09-04
Adding web.external-url and web.route-prefix flags

snebel29:
Hi,
First all, let's make it clear that this is my first contribution within the Prometheus ecosystem, but hopefully not the last one.

This PR aims to fix https://github.com/prometheus/blackbox_exporter/issues/475 making several assumptions and having different feelings on the way.

- Keep consistency with `alertmanager` and `prometheus` which seems to implement the same flags, I essentially copied the same logic adapting for the specifics of this component.
- I'm not forcing absolute url when `--web.external-url` is set, since all links are anyway relative reusing the original domain, I guess in other components more complex such as prometheus or alertmanager this might be necessary. I'm happy to change the behavior if forcing absolute path is what we want.

Since this component is so simple I've coded this PR using the naive approach, but I feel it probably makes sense to move the following [function](https://github.com/prometheus/blackbox_exporter/compare/master...snebel29:feat/add-support-for-web-route-prefix-external-url?expand=1#diff-7ddfb3e035b42cd70649cc33393fe32cR397-R429) with its [test case](https://github.com/prometheus/blackbox_exporter/compare/master...snebel29:feat/add-support-for-web-route-prefix-external-url?expand=1#diff-53618cae32e23ae47e1c35f10da92654R144-R218) to https://github.com/prometheus/common since the same function is used in the alertmanager as well (In fact It was literally copied from there)

May be creating something like https://github.com/prometheus/common/tree/master/promlog even trying to get the flags packed there, then implement in both `blackbox_exporter` and `alertmanger` and take a look to see if those flags are implemented anywhere else, but I would appreciate feedback on this point before doing anything.


main.go
```diff
@@ -255,25 +280,27 @@ func run() int {
...
-    <p><a href="/metrics">Metrics</a></p>
-    <p><a href="/config">Configuration</a></p>
-    <h2>Recent Probes</h2>
-    <table border='1'><tr><th>Module</th><th>Target</th><th>Result</th><th>Debug</th>`))
+		w.Write([]byte("<html>\n"))
+		w.Write([]byte("    <head><title>Blackbox Exporter</title></head>\n"))
+		w.Write([]byte("    <body>\n"))
+		w.Write([]byte("    <h1>Blackbox Exporter</h1>\n"))
+
+		w.Write([]byte(fmt.Sprintf("    <p><a href=\"%s/probe?target=prometheus.io&module=http_2xx\">Probe prometheus.io for http_2xx</a></p>\n", *routePrefix)))
```
brian-brazil: We could use relative paths here, the blackbox exporter doesn't do anything fancy.

snebel29: Great, since they are already relative urls I'll keep the code in this area the same then

snebel29:
Sorry for bothering again, but let's see if I understood correctly

> Let's go with consistency.

The only places were we could be using the absolute url value coming from `web.external-url` are

- ServeMux pattern of `Handle` and `HandleFunc` functions: But this would be inconsistent with other projects from the ecosystem, and kind of weird
- The html code returned by `/` path pattern matching: We already agreed on using relative path there 

It's therefore my understanding that we are already consistent and ok on this point.

> Probably best to do that with the TLS stuff when it's moved over to common. It's also in the pushgateway.

I'm not sure what you mean by `TLS` in the context of `alertmanager` and `pushgateway`, I can't see any obvious TLS stuff in their main files to move to common on them.

In any case my understanding is that you would want to address moving `extURL` together with "the TLS stuff" in a different PR.

Is this PR then ok to be merged?

brian-brazil:
> The only places were we could be using the absolute url value coming from web.external-url are

I meant consistent in the checks it performs. Having the semantics different across different servers would be confusing.

> I'm not sure what you mean by TLS in the context of alertmanager and pushgateway

It's in progress in the node exporter currently, and it'd make sense for all this code to be together at some future point.

snebel29:
> I meant consistent in the checks it performs. Having the semantics different across different servers would be confusing.

I agree, I took the logic from the `alertmanager`, therefore the same parameters across servers of the ecosystem should produce the same result

snebel29:
I will double-check `prometheus`, `pushgateway` and `alertmanager` behaviour for those two flags, to ensure the semantics are exactly the same (as it was intended).

Here is a recently implemented example https://github.com/prometheus/pushgateway/issues/190 which code I could reuse I guess, allowing to be moved on to common later on.

brian-brazil:
The logic in https://github.com/prometheus/prometheus/blob/d98d4a9bf04217f5b36fd3248cbe88d3046e84e4/cmd/prometheus/main.go#L791 is a bit different

snebel29:
@brian-brazil I reviewed prometheus behavior for `--web.external-url` and `--web.route-prefix` and I think I've matched it accordingly, note that alertmanager from where I originally "took" the logic is slightly different, see [here](https://github.com/prometheus/alertmanager/blob/d5f6f8892824a485663301b9a542c6b6dff91c27/cmd/alertmanager/main.go#L521-L553) so may be as an intermediary step towards leveraging common would be good to migrate the function in the alertmanager as well, at first glance the function signature is different and this [check](https://github.com/prometheus/alertmanager/blob/d5f6f8892824a485663301b9a542c6b6dff91c27/cmd/alertmanager/main.go#L542-L544) is not performed in prometheus.

From what I can see, only one behaviour will remain different between prometheus <> blackbox_exporter after merging this PR, currently blackbox_exporter matches any non registered path to the landing page, see [here](https://github.com/prometheus/blackbox_exporter/blob/ce8e9952a074768995cb67de0bc31bb43c1ec7c1/main.go#L265), while prometheus return 404s when a path wasn't explicitly registered, but this the case already before this PR.

This is hard to match due to using [ServeMux](https://golang.org/pkg/net/http/#ServeMux) in blackbox_exporter vs `github.com/prometheus/common/route` by leveraging [julienschmidt/httprouter](https://github.com/julienschmidt/httprouter) which matches exact patterns only.

I guess in the future, migrating over `github.com/prometheus/common/route` would make things easier if the intention is to behave 100% the same over time, also easier to maintain.

I hope this makes sense now, but please let me know otherwise.

Cheers

main.go
```diff
@@ -255,25 +291,27 @@ func run() int {
...
-    <p><a href="/metrics">Metrics</a></p>
-    <p><a href="/config">Configuration</a></p>
-    <h2>Recent Probes</h2>
-    <table border='1'><tr><th>Module</th><th>Target</th><th>Result</th><th>Debug</th>`))
+		w.Write([]byte("<html>\n"))
+		w.Write([]byte("    <head><title>Blackbox Exporter</title></head>\n"))
+		w.Write([]byte("    <body>\n"))
+		w.Write([]byte("    <h1>Blackbox Exporter</h1>\n"))
+
+		w.Write([]byte("    <p><a href=\"" + path.Join(*routePrefix, "/probe") + "?target=prometheus.io&module=http_2xx\">Probe prometheus.io for http_2xx</a></p>\n"))
```
brian-brazil: You can still leave these as relative.

snebel29:
How about that?

brian-brazil:
Thanks!

PR #517, opened 2019-09-04
Add separate history for expired failed probe results

### LINKED ISSUE #350: be able to persist failed probe history longer

jutley:
This PR adds a new section to the history to hold onto failed probes after they expire from the main history. This allows us to check what debug logs for failed probes in environments with a high rate of successful probes.

I implemented this to be strictly additive for the sake of backwards compatibility. The Web UI has a separate section for these preserved failures.

Resolves #350

### DIFF
--- a/history.go
+++ b/history.go
@@ -32,7 +32,7 @@ type resultHistory struct {
 	mu                     sync.Mutex
 	nextId                 int64
 	results                []*result
-	preservedFailedResults []*result 
+	preservedFailedResults []*result
 	maxResults             uint
 }
 
--- a/history.go
+++ b/history.go
@@ -26,10 +26,12 @@ type result struct {
 }
 
 type resultHistory struct {
-	mu         sync.Mutex
-	nextId     int64
-	results    []*result
-	maxResults uint
+	mu                        sync.Mutex
+	nextId                    int64
+	results                   []*result
+	maxResults                uint
+	preservedFailedResults    []*result
+	maxPreservedFailedResults uint
 }
 
 // Add a result to the history.
@@ -48,6 +50,14 @@ func (rh *resultHistory) Add(moduleName, target, debugOutput string, success boo
 
 	rh.results = append(rh.results, r)
 	if uint(len(rh.results)) > rh.maxResults {
+		if !rh.results[0].success {
+			rh.preservedFailedResults = append(rh.preservedFailedResults, rh.results[0])
+			if uint(len(rh.preservedFailedResults)) > rh.maxPreservedFailedResults {
+				preservedFailedResults := make([]*result, len(rh.preservedFailedResults)-1)
+				copy(preservedFailedResults, rh.preservedFailedResults[1:])
+				rh.preservedFailedResults = preservedFailedResults
+			}
+		}
 		results := make([]*result, len(rh.results)-1)
 		copy(results, rh.results[1:])
 		rh.results = results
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -46,11 +46,12 @@ var (
...
-	configCheck   = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
-	historyLimit  = kingpin.Flag("history.limit", "The maximum amount of items to keep in the history.").Default("100").Uint()
+	configFile                  = kingpin.Flag("config.file", "Blackbox exporter configuration file.").Default("blackbox.yml").String()
+	listenAddress               = kingpin.Flag("web.listen-address", "The address to listen on for HTTP requests.").Default(":9115").String()
+	timeoutOffset               = kingpin.Flag("timeout-offset", "Offset to subtract from timeout in seconds.").Default("0.5").Float64()
+	configCheck                 = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
+	historyLimit                = kingpin.Flag("history.limit", "The maximum amount of items to keep in the history.").Default("100").Uint()
+	historyPreservedFailedLimit = kingpin.Flag("history.preserved-failed-limit", "The maximum amount of failed items to preserve after expiration.").Default("5").Uint()
```
brian-brazil: I don't think we need an extra setting here.

jutley: Without the extra flag it's less obvious to me how this should be implemented. Should the `history.limit` flag be the cardinality of standard history and the cardinality of preserved failed history? Should that flag be the cardinality of the combined history? Should the preserved failed history be a constant size separate from the `history.limit` flag?

I'm open to anything, though I'll admit that without the extra flag, the semantics feel confusing to me: "I set the `history.limit` flag to 10, why do I have 20 items?"

brian-brazil: I'd have it be the limit of each.

history.go
```diff
@@ -26,10 +26,11 @@ type result struct {
...
 
 type resultHistory struct {
-	mu         sync.Mutex
-	nextId     int64
-	results    []*result
-	maxResults uint
+	mu                     sync.Mutex
+	nextId                 int64
+	results                []*result
+	preservedFailedResults []*result
```
brian-brazil: failedResults is enough

history.go
```diff
@@ -59,14 +68,19 @@ func (rh *resultHistory) List() []*result {
 	rh.mu.Lock()
 	defer rh.mu.Unlock()
 
-	return rh.results[:]
+	return append(rh.preservedFailedResults[:], rh.results...)
```
brian-brazil: You need to de-dupe here

jutley: Not true, actually. Maybe there is confusion about how I implemented this. I am adding results to the `preservedFailedResults` slice only after they expire from the main `results` slice. As a result, no result will be in both at the same time. 

If we put failed results into this slice immediately, then I think it would make more sense to have one list for successes and one list for failures, then merge the two on `List`.

brian-brazil: Ah, I see what you're doing now. A brief comment would help.

brian-brazil:
Thanks!

lbrichet:
Hello,

Sorry, I feel a bit stupid but I can't find where I can access this separate history ...

jutley:
@lbrichet It is below the main history section in the blackbox-exporter web ui.

lbrichet:
Doh ... thank you !

PR #520, opened 2019-09-11
Change the front page links to probe logs into relative links

### LINKED ISSUE #519: Support for --web.external-url and --web.route-prefix in 0.15.0 is incomplete
Linux hawkwind.cs 5.2.11-200.fc30.x86_64 #1 SMP Thu Aug 29 12:43:20 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

blackbox_exporter, version 0.15.0 (branch: HEAD, revision: 3f726a8a4489ea84a7163e37ba337f9b8332da63)
  build user:       root@3280fda7d3fc
  build date:       20190911-12:52:20
  go version:       go1.12.9

### The problem

If you run Blackbox with the new --web.external-url and --web.route-prefix options to try to operate it behind a reverse proxy, Blackbox's front page HTML still produces incorrect absolute links for the links to logs for probes, instead of either relative links or correct absolute links that use the external URL you set.

Looking at the code in commit 252a878a2c0e4b5961f04c765ca969224a8e0150, I think it should be prefixing these links with `*externalURL` instead of `*routePrefix` or simply leaving them as relative links. As I understand the settings, --web.route-prefix is the prefix that Blackbox sees in the requests from the front-end server, while --web.external-url is the external address, so visible URLs should always be generated using the latter (while being matched through the former).

Run `blackbox_exporter --config.file blackbox.yml --web.external-url=https://example.org/blackbox/ --web.route-prefix=/`, make a probe (eg `curl "localhost:9115/probe?module=http_2xx&target=https://google.com/"`), and then look at the HTML for Blackbox's front page. In the HTML, logs for probes will be linked as eg `<a href='/logs?id=0'>Logs</a>`, which is an incorrect external URL (it should be `/blackbox/logs?id=0`, even without the host).

siebenmann:
The front page contains a table of links to logs from the most recent
probes. These links were accidentally changed to absolute links in
252a878, in a form that doesn't work when using --web.external-url.
Rather than prefixing these with the external URL (if set), we return
them to being relative links, as they are meant to be (per issue #519).

Fixes #519

Signed-off-by: Chris Siebenmann <cks.git01@cs.toronto.edu>

### DIFF
--- a/main.go
+++ b/main.go
@@ -324,7 +324,7 @@ func run() int {
 			if !r.success {
 				success = "<strong>Failure</strong>"
 			}
-			fmt.Fprintf(w, "<tr><td>%s</td><td>%s</td><td>%s</td><td><a href='"+path.Join(*routePrefix, "/logs")+"?id=%d'>Logs</a></td></td>",
+			fmt.Fprintf(w, "<tr><td>%s</td><td>%s</td><td>%s</td><td><a href='logs?id=%d'>Logs</a></td></td>",
 				html.EscapeString(r.moduleName), html.EscapeString(r.target), success, r.id)
 		}
 


brian-brazil:
Thanks!

PR #523, opened 2019-09-18
Support systemd socket activation

tik-stbuehler:
We would like to run the exporter in a different network namespace (to run the probes in the network we want to monitor), but access the metrics through our (more reliable) default namespace network.

With (systemd) socket activation systemd binds the listening socket in the default namespace, and before the exporter is actually started we change to the target namespace; the already bound listening socket (passed as fd 3) stays in the original namespace.

Various other places still expect a listening address (logging, building URLs?), but I didn't touch those for now.

The code is based on https://vincent.bernat.ch/en/blog/2018-systemd-golang-socket-activation.

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,6 +44,8 @@ import (
 
 	"github.com/prometheus/blackbox_exporter/config"
 	"github.com/prometheus/blackbox_exporter/prober"
+
+	"github.com/coreos/go-systemd/activation"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


brian-brazil:
This is a layering violation, applications should not have to care about what init or supervision system is in use. I'd suggest trying to resolve this at the systemd level.

tik-stbuehler:
System socket activation is a concept driven by systemd, but doesn't actually depend on it. It uses a few environment variables to document how many file descriptors are passed for what reason to which PID, and that's it.

All I'm asking is to support using a file descriptor passed this way as listening socket. All the remaining text is just the motivation, not something `blackbox_exporter` needs to support in any way.

I'm open to using a different way to pass a listening socket from the init system (assuming I can use a wrapper to make it systemd compatible), e.g. an explicit "use fd X" option, but systemd socket activation is probably the most popular one right now.

brian-brazil:
The standard for Prometheus components is to support HTTP over TCP that we bind to in the usual fashion. I don't think it makes sense to add code for every possible other way of doing things.

tik-stbuehler:
That seems like a real poor excuse.

Can you please reread my first post again, and explain *why* you don't think it makes sense to want a different network to "control" the blackbox than what the blackbox monitors? (In our case we want to monitor shaky WLAN, and need a reliable transport (i.e. LAN) to see the results.)

There are other ways of course: you could support binding in other VRFs, or support running blackbox tests in other VRFs and other namespaces.

But the easiest solution is allowing to pass a listening socket (and please just get over the fact that the concept got named "systemd-socket-activation". It isn't based on systemd, and has in fact been in use for a long time for FastCGI applications, long before systemd existed).

PR #537, opened 2019-10-09
Make TCP port configurable

silkeh:
This PR adds a setting for the TCP port. This makes it easier to create probes for generic TCP protocols that have a standard port.

The additional test case is identical to `TestTCPConnection`, but with the port split and added via the configuration.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -154,6 +154,7 @@ type QueryResponse struct {
 }
 
 type TCPProbe struct {
+	Port               string           `yaml:"port,omitempty"`
 	IPProtocol         string           `yaml:"preferred_ip_protocol,omitempty"`
 	IPProtocolFallback bool             `yaml:"ip_protocol_fallback,omitempty"`
 	SourceIPAddress    string           `yaml:"source_ip_address,omitempty"`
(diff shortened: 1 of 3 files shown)


brian-brazil:
I'm not sure there's much need for this, this is as easy to add via relabelling already.

SuperQ:
Yes, this is already something controlled from the service discovery side.

silkeh:
The reason for this is a setup based on a list of URLs which then directly map the protocol to a module. Now these hostnames *could* of course include the ports, but that would be repetitive, error-prone and not as clear. Especially as the probes have to be configured for their specific protocol anyway.

brian-brazil:
This is already possible via relabelling.

silkeh:
@brian-brazil mind givind me a few pointers on how to do that? I can image something like the following, but that seems a bit convoluted (especially repeated for every probe):

```yaml
scrape_configs:
  - job_name: 'blackbox'
    metrics_path: /probe
    params:
      module: [smtp]
    static_configs:
      - targets:
        - example.com
        - example.com:587
    relabel_configs:
        regex: (:(\d+))?$
        replacement: ${2}
        target_label: __tmp_port
      - source_labels: [__tmp_port]
        regex: ^$
        replacement: :25
        target_label: __tmp_port
      - source_labels: [__address__,__tmp_port]
        separator: ""
        target_label: __param_target
      - target_label: __address__
        replacement: 127.0.0.1:9115
```

PR #538, opened 2019-10-09
Export TLS version

silkeh:
This PR adds the TLS version to the exported variables, or NaN when it is unknown (maybe that should be 0 instead?). All currently available TLS versions are covered. SSL3 is not, as it is not TLS and will be removed in Go 1.14 (See https://github.com/golang/go/issues/32716).

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -229,6 +229,14 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Help: "Returns earliest SSL cert expiry in unixtime",
 		})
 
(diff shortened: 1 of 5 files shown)


SuperQ:
Version information should be exposed as an info metric like this:

```
probe_tls_version_info{version="1.3"} 1
```

This avoids the problems with NaN in unknown.

silkeh:
Right! I initially followed the format of `probe_http_version`. I have added two fixup commits that implement it as `probe_tls_version_info`.

brian-brazil:
Thanks!

PR #540, opened 2019-10-23
[BUGFIX] disable ip protocol fallback and tests

genofire:
- fix ip protocol fallback
  - before the argument `fallbackIPProtocol` in `chooseProtocol()` (`prober/utils.go`) was never used
- correct / get default value of `IPProtocol` during testing
  - `ipv4` is a wrong value, it should be `ip4` or `ip6`
  -

### DIFF
--- a/prober/dns_test.go
+++ b/prober/dns_test.go
@@ -96,21 +96,24 @@ func TestRecursiveDNSResponse(t *testing.T) {
 	}{
 		{
 			config.DNSProbe{
-				IPProtocol: "ipv4",
-				QueryName:  "example.com",
+				IPProtocol:         "ip4",
+				IPProtocolFallback: true,
+				QueryName:          "example.com",
 			}, true,
 		},
 		{
 			config.DNSProbe{
-				IPProtocol:  "ipv4",
-				QueryName:   "example.com",
-				ValidRcodes: []string{"SERVFAIL", "NXDOMAIN"},
+				IPProtocol:         "ip4",
+				IPProtocolFallback: true,
+				QueryName:          "example.com",
+				ValidRcodes:        []string{"SERVFAIL", "NXDOMAIN"},
 			}, false,
 		},
 		{
 			config.DNSProbe{
-				IPProtocol: "ipv4",
-				QueryName:  "example.com",
+				IPProtocol:         "ip4",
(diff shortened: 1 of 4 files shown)


prober/utils_test.go
```diff
@@ -110,3 +115,30 @@ func generateTestCertificate(expiry time.Time, IPAddressSAN bool) ([]byte, []byt
 	pemKey := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(privatekey)})
 	return pemCert, pemKey
 }
+
+func TestChooseProtocol(t *testing.T) {
+	ctx := context.TODO()
```
brian-brazil: Background is fine here

prober/dns_test.go
```diff
@@ -226,27 +230,23 @@ func TestAuthoritativeDNSResponse(t *testing.T) {
 	}{
 		{
 			config.DNSProbe{
-				IPProtocol: "ipv4",
-				QueryName:  "example.com",
+				QueryName: "example.com",
```
brian-brazil: Leave these as v4 please.

prober/dns_test.go
```diff
@@ -96,21 +96,23 @@ func TestRecursiveDNSResponse(t *testing.T) {
...
+				IPProtocolFallback: true,
+				QueryName:          "example.com",
+				ValidRcodes:        []string{"SERVFAIL", "NXDOMAIN"},
 			}, false,
 		},
 		{
 			config.DNSProbe{
-				IPProtocol: "ipv4",
-				QueryName:  "example.com",
+				IPProtocol:         "ip4",
```
brian-brazil: Hmm, are we missing a sanity check on the value of this?

prober/dns_test.go
genofire: with fallback?

brian-brazil: It doesn't matter either way.

genofire: thats the problem, on my notebook some test failures if there is no fallback and it try only ipv4

prober/dns_test.go
genofire: jep, but please let somebody else do it ...

PR #544, opened 2019-11-05
exclude CA

SilentGo:
Check the certificate expiration date, exclude the CA authority, and get the last TLS expiration date by default. Sometimes it is the CA authority, and the expiration date of the certificate is not accurate

### DIFF
--- a/prober/tls.go
+++ b/prober/tls.go
@@ -21,7 +21,7 @@ import (
 func getEarliestCertExpiry(state *tls.ConnectionState) time.Time {
 	earliest := time.Time{}
 	for _, cert := range state.PeerCertificates {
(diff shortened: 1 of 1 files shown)


brian-brazil:
I don't see why we should exclude certs just because they have permission to grant other certs, they can still expire.

SilentGo:
The probe detects the expiration time of the corresponding domain name certificate, now , the return time is the expiration time of the CA organization, not the expiration time of the domain name's certificate. I think the expiration time of the CA organization's certificate is less than the expiration time of the domain name certificate. What's your idea?

brian-brazil:
Any cert in the chain expiring is a problem.

PR #545, opened 2019-11-05
Add http endpoint support

mblaschke:
Add support for custom endpoint (IP/Port; skips DNS resolve) to make probes possible in specific and complex environments:

- possiblity to check a service internally even if a public PaaS cloud gateway is in front of it to check outages in the PaaS/network chain -> outage detection using Prometheus service discovery
- checking services behind eg. Cloudflare and send the probes to the server instead to the CDN
- cloud environments where no DNS split horizon is available or not stable (eg. Azure)
- checking if hosts are NOT available on a loadbalancer
- corporate environments where external DNS requests are not allowed

example module configuration:
```
  http_2xx_endpoint_example:
    prober: http
    timeout: 5s
    http:
      endpoint: "1.1.1.1:443"
      valid_http_versions: ["HTTP/1.1", "HTTP/2"]
      method: GET
      fail_if_ssl: false
      fail_if_not_ssl: false
      ip_protocol_fallback: false  # no fallback to "ip6"
```

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -45,6 +45,9 @@ The other placeholders are specified separately.
   # The HTTP method the probe will use.
   [ method: <string> | default = "GET" ]
 
+  # Endpint IP (IP or IP:Portand port) to send HTTP requests, skip DNS lookup.
+  [ endpoint: <string> ]
+
   # The HTTP headers set for the probe.
   headers:
     [ <string>: <string> ... ]
--- a/config/config.go
+++ b/config/config.go
@@ -131,6 +131,7 @@ type HTTPProbe struct {
 	NoFollowRedirects            bool                    `yaml:"no_follow_redirects,omitempty"`
 	FailIfSSL                    bool                    `yaml:"fail_if_ssl,omitempty"`
 	FailIfNotSSL                 bool                    `yaml:"fail_if_not_ssl,omitempty"`
+	Endpoint                     string                  `yaml:"endpoint,omitempty"`
 	Method                       string                  `yaml:"method,omitempty"`
 	Headers                      map[string]string       `yaml:"headers,omitempty"`
 	FailIfBodyMatchesRegexp      []string                `yaml:"fail_if_body_matches_regexp,omitempty"`
--- a/example.yml
+++ b/example.yml
@@ -52,6 +52,18 @@ modules:
       method: GET
       tls_config:
         ca_file: "/certs/my_cert.crt"
+  http_2xx_endpoint_example:
+    prober: http
+    timeout: 5s
+    http:
+      endpoint: "1.1.1.1:443"
+      valid_http_versions: ["HTTP/1.1", "HTTP/2"]
+      valid_status_codes: []  # Defaults to 2xx
+      method: GET
+      fail_if_ssl: false
+      fail_if_not_ssl: false
+      preferred_ip_protocol: "ip4" # defaults to "ip6"
+      ip_protocol_fallback: false  # no fallback to "ip6"
   tls_connect:
     prober: tcp
     timeout: 5s
--- a/prober/http.go
+++ b/prober/http.go
@@ -228,6 +228,7 @@ func (t *transport) GotFirstResponseByte() {
 
 func ProbeHTTP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var redirects int
+	var ip *net.IPAddr
 	var (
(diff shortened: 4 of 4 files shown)


brian-brazil:
Thank you for your PR. As I've already indicated this doesn't make sense in
terms of the blackbox exporter's design, and is already possible via
existing functionality.

mblaschke:
Here an example what's possible with this PR with full Prometheus service discovery.
It's not possible to use Prometheus service discovery if every check needs an own module configuration. It's not possible via the existing functionality with the current feature set as it makes Prometheus service discovery completely unusable.

blackbox-exporter

Also there are environments where DNS resolving is not possible so you have to set a dedicated endpoint because of corporate global security rules.

mblaschke:
Maybe I don't see a proper solution so i here my summary:

## Situation

We have multiple hundreds of ingresses (hosts and subpaths) which we want to check against different endpoints using Prometheus service discovery to ensure the connection chain is running (this is only one usecase).

```
ingresses: 
- foo.example.com
- foo.example.com/subpath
- bar.example.com
- foobar.example.com
[...]
```

## solution without PR

The blackbox_exporter only supports `target` and `module` as parameter for probes so we can pass the `module` name and either the URL or the endpoint.

So if we want to run 3 different checks with let's say 250 ingress definitions we need 750 modules inside blackbox_exporter and 3 scraping configurations within Prometheus with service discovery:

```
  http_check1_foo.example.com:
    prober: http
    timeout: 30s
    http:
      headers:
        Host foo.example.com
  http_check2_foo.example.com:
    prober: http
    timeout: 30s
    http:
      headers:
        Host foo.example.com
  http_check3_foo.example.com:
    prober: http
    timeout: 30s
    http:
      headers:
        Host foo.example.com
[...]
```

blackbox_exporter isn't capable of service discovery so it needs another service which is generating the configuration and needs somehow to be in sync with the Prometheus service discovery (to avoid failing probes).

In Prometheus we have to make for each endpoint check one scaping configuration and inject (somehow) the ingress domain/path into the `module` name and pass the endpoint IP as `target`.
Can we escape the ingress domain/path so we don't have an special chars here that needs to be escaped? Maybe md5/sha1 hash for the module name? But that makes it very difficult to debug these checks...


## solution with PR

blackbox_exporter defines one module per endpoint:

```
  http_check1:
    prober: http
    timeout: 30s
    http:
      endpoint: 1.2.3.4:443

  http_check2:
    prober: http
    timeout: 30s
    http:
      endpoint: 1.2.3.5:443


  http_check3:
    prober: http
    timeout: 30s
    http:
      endpoint: some-api-gateway:443
```

In Prometheus we have to make for each endpoint check one scaping configuration and inject the `module` name `http_check1`, `http_check2` and `http_check3`.

## Forgot something?
You told me that this is already possible in blackbox_exporter, maybe I don't see a solution here but I don't think that throwing hundreds or thousands of modules against blackbox_exporter is a clean or an elegant solution if we can solve this with a simple configuration option.

brian-brazil:
> You told me that this is already possible in blackbox_exporter, maybe I don't see a solution here but I don't think that throwing hundreds or thousands of modules against blackbox_exporter is a clean or an elegant solution

It is a solution though. The blackbox exporter is better for smoketests than exhaustive regression testing.

mblaschke:
Hm.. I would not call a workaround a "solution" which is misusing service discovery from Prometheus which also needs an additional service discovery with an increasing number of configuration.
Even we have demand for this feature for smoke tests as i mentioned that it's only one use case in a corporate environment. 
I don't understand what you mean with regression tests here becasue it's used for smoke tests. I currently don't have the feeling that the use case for this PR is fully clear?

The proposed solutions with the workaround makes 100 times more work and is difficult to debug than just these bunch of lines. At least I'm not the only one if I look at the reactions.

Disappointed for not getting an elegant cloud solution here 😞

mblaschke:
After a another evaluation, I still want to push this PR for smoke tests.
This PR solves problems were other solutions would lead to wrong probes and so to wrong metrics.

It should not be the case to write for each test one module. This feels like 90s where you write a big textfile with urls for wget to check and download...

brian-brazil:
That's exactly what modules are, tests. It's designed for when you have e.g. many ingress servers that need testing that they're working at all rather than that a single ingress server has hundreds of settings perfect. As I've indicated above I'm afraid I can't accept this change.

mblaschke:
> It's designed for when you have e.g. many ingress servers that need testing that they're working at all rather than that a single ingress server 

And that's what we're trying but we cannot rely on the published DNS entries because of a corporate environment and some PaaS services. Even redirecting with iptables is PITA.

Thanks for not trying to understand a corporate use case 😞

PR #552, opened 2019-11-20
Add none/all matches to DNS RRset validation

wodity:
The current validators for the DNS probers (if_matches/if_not_matches) pass if _any_ record matches/does not match. This patch adds the opposite sense of each, so that a validation will pass if _no_ records match, or _all_ records match. This is much more useful for example with TXT records where you might want to check if a particular record is present or not without failing if any record does not match.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -165,7 +165,13 @@ validate_answer_rrs:
   fail_if_matches_regexp:
     [ - <regex>, ... ]
 
+  fail_if_all_match_regexp:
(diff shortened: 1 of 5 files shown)


brian-brazil:
Can you do a make format to make CI happy?

wodity:
Updated.

brian-brazil:
The configuration docs need updating too.

wodity:
Ah, missed that. Updated those too.

brian-brazil:
Thanks!

Issue #553, opened 2019-11-21
module HTTP: hostname resolved locally even when using a proxy


Scapal:
When using a proxy in a http probe, the http target address is resolved locally instead of letting the proxy server resolve it. 

Would it be possible to add an option that doesn't replace the targetURL with the IP address ?

https://github.com/prometheus/blackbox_exporter/blob/3a08e67780bcb728a42cbf00ba03e0bd363a8e8e/prober/http.go#L370

brian-brazil:
Dupe of #553.

Scapal:
@brian-brazil, it has nothing to do with forcing https described in #503 even if I first thought it was related.

brian-brazil:
So it is, #264 is probably closest then.

Issue #543, opened 2019-10-29
Feature request: Custom endpoint IP for http checks


mblaschke:
Allow custom endpoint IP addresses for HTTP checks.
Sometimes services are exposed behind a NAT or not reachable via the published DNS address so it would be nice to specify a custom IP address for overriding the DNS request.

possible example:

```
  http_2xx:
    prober: http
    timeout: 30s
    http:
      valid_http_versions: ["HTTP/1.1", "HTTP/2"]
      preferred_ip_protocol: "ip4"
      endpoint: "10.1.2.3"
```

The current solution (creating a module for each check) is too much work and creates a huge configuration :(

brian-brazil:
You can already override the host header, and running the blackbox exporter behind the NAT would also make things simpler.

> The current solution (creating a module for each check) is too much work and creates a huge configuration :(

Automation is your friend.

mblaschke:
We have around 5 LoadBalancer and several hundreds of host which we want to check.
So that means we have to create hundreds of seperated scraping configs AND blackbox modules.

Sure this is possible but it's not really elegant just to throw configuration against the system :)

When we can override the host header why not override the endpoint ip address? :)

brian-brazil:
> When we can override the host header why not override the endpoint ip address? :)

That's what the target URL parameter is.

mblaschke:
Ah sure but we also need for each host one scape configuration as we also have to specify the module every time.

Also Prometheus supports service discovery (eg. `kubernetes_sd` or `file_sd`) with watches, but blackbox exporter lacks this feature, right?

In a cloud environment where the hosts are constantly changing this doesn't feel right to regenerate the configuration every time the cloud changes.

It would be nice to override the endpoint IP like we can override the headers instead of generating the configuration and reloading the blackbox exporter. Debugging in prometheus is much easier then in blackbox exporter.

mblaschke:
Would also sponsor a beer or two 😊

brian-brazil:
> Also Prometheus supports service discovery (eg. kubernetes_sd or file_sd) with watches, but blackbox exporter lacks this feature, right?

Yes, that's Prometheus's responsibility.

This sounds like something that you can do with relabelling on the Prometheus end.

mblaschke:
It's possible to set the `Host` header override via http parameter for the probe? hmm...haven't found it yet.

We just want to avoid a huge generated configuration if there is a possibility to override the endpoint.

[...]

PR #554, opened 2019-11-27
Add use_proxy_dns to bypass local DNS resolution

Scapal:
Addressing issue #553 and possibly #264 .

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -92,6 +92,9 @@ The other placeholders are specified separately.
   # HTTP proxy server to use to connect to the targets.
   [ proxy_url: <string> ]
 
+  # Skip local DNS resolution when using an HTTP proxy server.
+  [ use_proxy_dns: <boolean> | default = false ]
+
   # The IP protocol of the HTTP probe (ip4, ip6).
   [ preferred_ip_protocol: <string> | default = "ip6" ]
   [ ip_protocol_fallback: <boolean> | default = true ]
--- a/config/config.go
+++ b/config/config.go
@@ -138,6 +138,7 @@ type HTTPProbe struct {
 	FailIfHeaderMatchesRegexp    []HeaderMatch           `yaml:"fail_if_header_matches,omitempty"`
 	FailIfHeaderNotMatchesRegexp []HeaderMatch           `yaml:"fail_if_header_not_matches,omitempty"`
 	Body                         string                  `yaml:"body,omitempty"`
+	UseProxyDNS                  bool                    `yaml:"use_proxy_dns,omitempty"`
 	HTTPClientConfig             config.HTTPClientConfig `yaml:"http_client_config,inline"`
 }
 
--- a/prober/http.go
+++ b/prober/http.go
@@ -367,16 +367,20 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		httpConfig.Method = "GET"
 	}
 
-	// Replace the host field in the URL with the IP we resolved.
+	// Replace the host field in the URL with the IP we resolved unless letting a proxy server do the name resolution.
 	origHost := targetURL.Host
-	if targetPort == "" {
-		if strings.Contains(ip.String(), ":") {
-			targetURL.Host = "[" + ip.String() + "]"
+	skipDNS := module.HTTP.UseProxyDNS && module.HTTP.HTTPClientConfig.ProxyURL.URL != nil
+
+	if !skipDNS {
+		if targetPort == "" {
+			if strings.Contains(ip.String(), ":") {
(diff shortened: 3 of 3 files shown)


brian-brazil:
As indicated this breaks other features, so doesn't make sense. As I've previously stated I'd consider a feature to better test HTTP proxies, but DNS resolution of whatever is passed in with the URL needs to stay inside the blackbox exporter.

Scapal:
I really don't get it why you states it breaks other features.
The purpose of black box monitoring is end to end testing: "does it works from here".
Even now when using the existing proxy option you actually don't control the IP protocol used to connect to the proxy, only the target host.
If you use the option, you know it bypasses the local DNS resolution and IP protocol selection, what is wrong with that?

Please reconsider accepting this PR.

brian-brazil:
The purpose of the blackbox exporter is to test many many replicas of a service, including knowing what it was that you actually tested.

If a proxy does resolution we don't know what was tested, and as mentioned before that http proxies are supported at all is because it happens to come from the http library we use - it was never an intended feature of the blackbox exporter. If you need to test things at the far end of a proxy, you can have Prometheus use the proxy_url and put the blackbox exporter on the other side of the proxy.

Scapal:
We are not authorised to install an exporter on the far end of the proxy and the purpose is to test wether the service is working from here.
For example, one of my customer HQ moved servers to another network segment. It was transparent for users on the headquarter network but stopped working for branch offices forced to use the proxy.
Blackbox_exporter should imho stay a Swiss army knife. Monitoring many many replicas of a services being just one possible usage.
Even in your use case, using a Docker service IP address would also go through load-balancers...

brian-brazil:
We can't cover every possible deployment model to work in the easiest way possible I'm afraid, and we end up presuming that there is only one DNS view. Monitoring many many replicas of a service is explicitly the design goal, in this case the idea would be to test that each of multiple proxy servers work for getting to some HTTP fixed endpoint - which isn't supported as noone has proposed a PR/design yet.

I believe what I proposed would have caught that particular issue, as Prometheus wouldn't have been able to get to the blackbox exporter.

> We are not authorised to install an exporter on the far end of the proxy and the purpose is to test wether the service is working from here.

Unfortunately we can't help with problems arising due to internal policies/politics, as otherwise that becomes a general argument for everything.

Scapal:
Could we have additional opinions about this PR or are you the sole contributor to have a voice for this repository ?

brian-brazil:
I'm the sole maintainer of this repo, per our governance you can bring it up on the developers list.

mblaschke:
Hm.. that would maybe also solve our issue within the corporate network.

Something similar I already tried to address with https://github.com/prometheus/blackbox_exporter/issues/543 and https://github.com/prometheus/blackbox_exporter/pull/545

pbeaumontQc:
We also need this and have been monitoring answer since issue #303 ... We have a list of third party providers that we need to monitor, as a breakdown from one of them directly breaks our ability to function. Our instances are hosted on AWS EC2 and all go through a proxy to the 3rd party providers... Any update on this much desired feature?

brian-brazil:
As I've already indicated this feature as-stated will not be accepted. I'd suggest looking at moving where the exporter runs, or fixing your proxy server to work with IP addresses.

lyz-code:
Although it's stated that it won't be implemented, I'd also need this feature.

The ugly workaround I'm implementing is creating "fake" DNS records in my internal DNS server so the probe sees they exist and then the probe works as expected

riuvshyn:
This is frustrating... I also need this feature.

riuvshyn:
> As indicated this breaks other features

@brian-brazil  can you please elaborate on what exact features will be impacted by this change?

Most proxy setups do ACLs based on DNS names because IPs are not constants. 
The proposed change is optional and will not affect users who are not using HTTP proxy but those who using proxy can use Blackbox exporter without this at all...

brian-brazil:
As indicated above, it breaks the ability to see what you talked to.

I'll be closing this now, as there's no plans to accept any feature like this. As previously stated I am willing to consider features to allow probing of http proxies.

dbluxo:
We would also be happy about this feature :disappointed:

riuvshyn:
@dbluxo that is why there are 552 forks ;)

dbluxo:
Like this PR, [PR-338](https://github.com/prometheus/blackbox_exporter/pull/338) was also rejected. Also with the hint to discuss it on the prometheus developers list. Has anyone tried it yet?

riuvshyn:
@dbluxo changes in this PR works perfectly fine.

ChristianCiach:
Can you guys recommend a fork that fixes this? I don't have much trust in a project that is maintained by a single person anyway, but given his comments on this and other issues I've lost any remaining trust that I may have had before.

PR #556, opened 2019-12-04
Add cmd probe

atompi:
Add cmd probe. Allows users scrape metrics by execute their specified commands. Now you only need to modify part of the configuration file.


brian-brazil:
We explicitly don't shell out in this exporter, this is a network exporter, plus you're ignoring the target. I'd suggest looking at the node exporter's textfile collector.

Issue #555, opened 2019-12-03
Please specify the user-agent header in the HTTP requests.


pvalsecc:
### What did you expect to see?

I'm expecting to be able to filter out the requests from blackbox exporter in my access logs.
So, please, add a user agent header that container the "blackbox exporter" string somewhere.

### What did you see instead?

I see User-Agent: Go-http-client/1.1
Not easy to filter in the access logs...

pvalsecc:
Looks like it's possible:
```yaml
modules:
  http_2xx:
    prober: http
    http:
      #preferred_ip_protocol: "ip4" # defaults to "ip6"
      #ip_protocol_fallback: false  # no fallback to "ip6"
      headers:
        User-Agent: "blackbox-monitoring"
```

brian-brazil:
Would you like to send a PR? Note that we must allow this to be overridden, and see the format we use over in Prometheus for scapes.

pauloczava:
This Issue seems to be already resolved since you can change headers.

`modules:
  http_2xx:
    prober: http
    timeout: 30s
    http:
      valid_status_codes: [] #default to 2xx
      method: GET
      headers:
        User-Agent: potatohead1.1`

tdm4:
This is not working for me:
```
modules:
  http_2xx:
    prober: http
    timeout: 60s
    http:
      headers:
        User-Agent: "blackbox_exporter"
```

Restarted prometheus and it continues to use `Go-http-client/1.1` regardless.

mnacharov:
@pvalsecc @electron0zero shouldn't this issue be closed as resolved?

electron0zero:
thank you, going to close this.

`User-Agent` header can be confiured via `headers` config, see more: https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md#http_probe

PR #557, opened 2019-12-12
Change default user agent

nikonhub:
Hello.

According to prometheus scrape behavior [here](https://github.com/prometheus/prometheus/blob/48d25e6fe7d960c76c01763303c38f11168b1bba/scrape/scrape.go#L520) Change default go http client user-agent
from "Go-http-client/1.1" to "BlackBox Exporter/{{ version }}"

Issue : https://github.com/prometheus/blackbox_exporter/issues/555
Maintainers : @brian-brazil

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -227,7 +227,7 @@ func (t *transport) GotFirstResponseByte() {
 	t.current.responseStart = time.Now()
 }
 
-var userAgentDefaultHeader = fmt.Sprintf("BlackBox Exporter/%s", version.Version)
+var userAgentDefaultHeader = fmt.Sprintf("Blackbox Exporter/%s", version.Version)
 
 func ProbeHTTP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var redirects int
@@ -406,7 +406,8 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		request.Header.Set(key, value)
 	}
 
-	if request.Header.Get("User-Agent") == "" {
+	_, hasUserAgent := request.Header["User-Agent"]
+	if !hasUserAgent {
 		request.Header.Set("User-Agent", userAgentDefaultHeader)
 	}
 
--- a/prober/http.go
+++ b/prober/http.go
@@ -35,6 +35,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	pconfig "github.com/prometheus/common/config"
+	"github.com/prometheus/common/version"
 	"golang.org/x/net/publicsuffix"
 
 	"github.com/prometheus/blackbox_exporter/config"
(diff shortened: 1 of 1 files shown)


prober/http.go
```diff
@@ -403,6 +406,10 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		request.Header.Set(key, value)
 	}
 
+	if request.Header.Get("User-Agent") == "" {
```
brian-brazil: Hmm, what if the user explicitly wants no User-Agent header?

prober/http.go
```diff
@@ -226,6 +227,8 @@ func (t *transport) GotFirstResponseByte() {
 	t.current.responseStart = time.Now()
 }
 
+var userAgentDefaultHeader = fmt.Sprintf("BlackBox Exporter/%s", version.Version)
```
brian-brazil: Blackbox

prober/http.go
Nikosmonaut: Initially I was looking for Has method to check if key exists but it is not exported.
What do you think about this solution ?
(Since all headers are added with Set() we are sure that the key was formatted to canonical key)

```
	_, hasUserAgent := request.Header["User-Agent"]
	if !hasUserAgent {
		request.Header.Set("User-Agent", userAgentDefaultHeader)
	}
```
blackbox.yaml
```
headers:
  // Result : User-Agent: "My user agent"
  User-Agent: "My user agent"

headers:
  // Result : User-Agent: ""
  User-Agent: ""

headers:
  // Result : User-Agent: "Blackbox Exporter/0.16"
  User-Agent:

headers:
  // Result : User-Agent: "Blackbox Exporter/0.16"
  User-Agent: null
```

brian-brazil: I think it'll do for now, we can tune the null case in future if someone asks.

Nikosmonaut:
By the way I checked the case of null value.
I think it goes beyond the simple test of key exists because it's not in module.HTTP.

I tried to remove the omitempty but no changes
I found that a nil value can not be assigned to a string.
```
Headers                      map[string]string       `yaml:"headers,omitempty"`
```

brian-brazil:
omitempty affects marshalling, not unmarshalling. Null values are fun, we've had many bugs due to them.

edevil:
This seems like a useful change. Is there some work to do besides fixing the conflicts?

brian-brazil:
Hmm, looks like that's the case. Better handling of not sending the header at all would also be good.

Nikosmonaut:
Yo.

Would like me to do anything ? Honestly I'm not sure what to do about this PR. I followed this issue #555, and it's been one year with no update on this PR nor the issue. Is it still useful ?

Personnaly I just set the User-Agent to BlackBox in config yml to get the same behaviour.

brian-brazil:
It is still useful. It's technically breaking, but it doesn't hurt to help track where requests are coming from.

Resolving the conflicts, and allowing for no header to be set if practical, would be good.

cofyc:
any progress?

megamorf:
I'd also like to see this merged. 

**Use case:** 
Identify monitoring traffic and exclude it from application logs.

SuperQ:
Going to close and re-open this to see if the tests still pass.

SuperQ:
Seems like this is ready to go.

PR #566, opened 2020-01-03
Skip network dependent test in short mode

dswarbrick:
Skip the network-dependent `TestChooseProtocol` test when testing in short mode. Also fix spelling of "dependent" elsewhere in tests.


brian-brazil:
Thanks!

PR #575, opened 2020-02-25
Add resolved IP address as a label on the protocol field

imduffy15:
Adds the resolved IP as a label to allow for tracking IP changes.


brian-brazil:
This would be a breaking change, and in addition may pose cardinality issues.

imduffy15:
Can we come to a resolution or should I close and fork?

brian-brazil:
What do you propose? This sort of thing is generally best handled via logging, a given endpoint could easily have thousands of IPs over time.

imduffy15:
Closing in favor of logs.

Issue #579, opened 2020-03-05
[suggestion] Omitting metrics if not success


zerkms:
At the moment if the http probe fails `probe_success=0` - the exporter still exports

```
probe_http_duration_seconds{phase="connect"} 0
probe_http_duration_seconds{phase="processing"} 0
probe_http_duration_seconds{phase="resolve"} 0
probe_http_duration_seconds{phase="tls"} 0
probe_http_duration_seconds{phase="transfer"} 0
```

Which is a lie - none of those phases completed in a blazing fast 0 seconds.

Why this is a problem: if you perform any analytics over those metrics (eg percentiles, or min) it would be skewed.

Suggestion: what if the exporter does not export metrics it could not really scrape instead of exporting zero values?

brian-brazil:
That a probe failed doesn't mean that the HTTP request failed, the status code could have been wrong for example.

zerkms:
@brian-brazil that's right, but imagine the following query:

`min_over_time(probe_http_duration_seconds{phase="connect"}[1h])` which semantically means: "the quickest tcp connect over period", practically it would return `0` though, which is wrong.

My core point here is - if a particular metric was not actually scraped (for whatever reason: be it a transport error, or http status code error, or any other error) - the complete metric must be removed from the response, not the zero value returned as it's now.

brian-brazil:
The TCP connection time doesn't become invalid because a subsequent HTTP request failed, this isn't sane in general.

zerkms:
@bbigras 

> The TCP connection time doesn't become invalid because a subsequent HTTP request failed, this isn't sane in general.

That's NOT what I said.

I probably am terrible at explaining, but I will try once again.

If **a particular metric** was not actually scraped **for whatever reason** - it **must not be** exported, at all, since `0` is a lie.

Returning to my initial example:

```
probe_http_duration_seconds{phase="connect"} 0
probe_http_duration_seconds{phase="processing"} 0
probe_http_duration_seconds{phase="resolve"} 0
probe_http_duration_seconds{phase="tls"} 0
probe_http_duration_seconds{phase="transfer"} 0
```

Here are the metrics exported for a target with not-existing DNS address. Do you think those zeroes make any sense?

SuperQ:
@brian-brazil I think the point is that exposing TCP connection duration isn't valid if we fail early, for example at URL parsing or DNS lookup.

[...]

Issue #485, opened 2019-06-11
Bogus probe_http_duration_seconds{phase="tls"} values if more than one redirect


dswarbrick:
### Host operating system: output of `uname -a`
```
Linux ber-prom1 4.19.0-5-amd64 #1 SMP Debian 4.19.37-3 (2019-05-15) x86_64 GNU/Linux
```
### blackbox_exporter version: output of `blackbox_exporter -version`
```
blackbox_exporter, version 0.14.0+ds (branch: debian/sid, revision: 0.14.0+ds-1~ionos1)
  build user:       pkg-go-maintainers@lists.alioth.debian.org
  build date:       20190606-12:27:16
  go version:       go1.11.6
```
### What is the blackbox.yml module config.
```yaml
  http:
    prober: http
    http:
      preferred_ip_protocol: ip4
      headers:
        User-Agent: blackbox_exporter
```
### What is the prometheus.yml scrape config.
```yaml
- job_name: blackbox-http
  metrics_path: /probe
  params:
    module: [http]
  file_sd_configs:
  - files:
    - targets/http.yml
  relabel_configs:
  - source_labels: [__address__]
    target_label: __param_target
  - source_labels: [__param_target]
    target_label: instance
  - target_label: __address__
    replacement: localhost:9115
```
### What logging output did you get from adding `&debug=true` to the probe URL?
```
Logs for the probe:
ts=2019-06-11T09:52:35.312192849Z caller=main.go:118 module=http target=http://backup.profitbricks.com level=info msg="Beginning probe" probe=http timeout_seconds=9.5
ts=2019-06-11T09:52:35.312481462Z caller=utils.go:42 module=http target=http://backup.profitbricks.com level=info msg="Resolving target address" ip_protocol=ip4
ts=2019-06-11T09:52:35.31408387Z caller=utils.go:75 module=http target=http://backup.profitbricks.com level=info msg="Resolved target address" ip=81.169.163.183
ts=2019-06-11T09:52:35.314177504Z caller=http.go:349 module=http target=http://backup.profitbricks.com level=info msg="Making HTTP request" url=http://[81.169.163.183] host=backup.profitbricks.com
ts=2019-06-11T09:52:35.322419932Z caller=http.go:305 module=http target=http://backup.profitbricks.com level=info msg="Received redirect" url=https://backup.profitbricks.com/
[...]

PR #581, opened 2020-03-07
Don't init http phase values

SuperQ:
Don't init http probe phase labels for phases we haven't observed.
Avoids exposing 0 value samples for things we haven't measured.

Fixes: https://github.com/prometheus/blackbox_exporter/issues/579

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -288,10 +288,6 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		})
 	)
 
-	for _, lv := range []string{"resolve", "connect", "tls", "processing", "transfer"} {
-		durationGaugeVec.WithLabelValues(lv)
-	}
-
 	registry.MustRegister(durationGaugeVec)
 	registry.MustRegister(contentLengthGauge)
 	registry.MustRegister(bodyUncompressedLengthGauge)


brian-brazil:
I'm not sure about this, having time series not predictably present (e.g. tls might be missing if a redirect breaks) could be problematic for users to handle correctly and this is also a breaking change. There's also issues like #485 around multiple requests.

We also do this for all metrics, not just this one, as this design decision predates staleness. So it'd have to be considered more broadly. In general I view these metrics as audit logging to assist figuring out why a particular probe failed, not as something you'd do math on.

SuperQ:
These specific metrics are used for math, so we should be more careful about them.

We could also move the init for these metrics later in the probe phase. That way we keep the init, but don't create them when things fail early.

brian-brazil:
`probe_success` and `probe_ssl_earliest_cert_expiry` are the only metrics intended for math, everything else is audit log as far as I'm concerned. We can't record a full log of a probe in Prometheus, but we can use metrics to help narrow down what went wrong after the fact.

If we're going to change the approach of this in the code we design it wholesale, rather than metric by metric.

zerkms:
To me as a metrics consumer it's awkward to see some metrics to be _deliberately_ unreliable.

If something is exposed for debugging only purposes only - it could have had a `_debug` or something in its name. At this moment those numbers look like any other number in any other exporter :man_shrugging:

brian-brazil:
> If something is exposed for debugging only purposes only - it could have had a _debug or something in its name

The vast majority of all metrics are only for debugging, so that'd be a bit redundant :)

electron0zero:
Q: can we close this PR in favor of https://github.com/prometheus/blackbox_exporter/pull/865?

PR #584, opened 2020-03-18
add probe_ip_addr_hash to detect if the IP changes

Wing924:
## WHAT

add `probe_ip_addr_hash` metrics to specify the hash of IP address

## WHY

We want to detect if the IP is changed by `change(probe_ip_addr_hash[10h])`

### Use Case

Assume you use CDN and you want to monitoring your web server.

```
blackbox_exporter --> CDN edge server --> origin server
```

if the edge server failover, blackbox exporter returns error and the IP address will be changed.
With `probe_ip_addr_hash`, we can tell the error happen due to edge failover.

### DIFF
--- a/prober/utils.go
+++ b/prober/utils.go
@@ -16,6 +16,7 @@ package prober
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
 	"net"
 	"time"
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can we use a better hash?

Wing924:
@brian-brazil this hash is copied from java's `String.hashCode()`.
http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/String.java#l1452

Why I use this simple hash instead of SHA256:
* we don't need slow cryptographic hash function because IP address is not secret
* we can't set 256bit to value because value is `float64`
* unsigned 32bit integer looks good when you convert to float64

But I'm not have strong opinion on using this simple hash.
Do you have any recommendation?

brian-brazil:
Use any hash that Go provides and truncate it to 32 bits.

Wing924:
@brian-brazil I've changed the hash func to Go provided.

prober/utils.go
```diff
@@ -37,8 +38,20 @@ func chooseProtocol(ctx context.Context, IPProtocol string, fallbackIPProtocol b
 		Name: "probe_ip_protocol",
 		Help: "Specifies whether probe ip protocol is IP4 or IP6",
 	})
+
+	probeIPAddr := prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
brian-brazil: This IP could have high cardinality, it's not safe to expose it.

Wing924: Understood. I've removed that.

brian-brazil:
Thanks!

basroovers:
Looking for some documentation on the metric on how this can be used lead me to this pull request.
I just wanted to share how an alert rule would work (1st post already showed it but this won't work since "change" is a non-existent function, "changes" is):
```
changes(probe_ip_addr_hash[10h])
```

PR #595, opened 2020-04-10
support 'params_from_url'

talebook:
With this option, prometheus can past specific param on url, which make it easier to build dynamic task

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,65 @@
+# Attentions
+This branch has a new feature that original mater branch does not plan to added:
+* A New Config: `params_from_url`
+
+By enable this config, the specific param list whill be read from url query arguments ( if provided ).
+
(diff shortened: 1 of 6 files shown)


brian-brazil:
Thank you for your PR. For sanity and to prevent becoming an open proxy, this is not a feature that is ever planned to be added. If you need different configurations, create a module each.

talebook:
i'm sorry that i don't understand why this feature make it becoming an open proxy?

In my case, I have many blackbox_exporter instances and dns tasks which make the blackbox.yml difficult to maintain. With this new feature the job can be done by just modify the prometheus config. I think it may be helpful for other users.

Please let me know more of your plans, thanks.

brian-brazil:
As I said this isn't something that I plan on adding. You can use configuration management to generate your config.

PR #605, opened 2020-04-22
Add password file field to document

tottoto:
## Summary

- Add `password_file` field to document of http probe.
- Ref. https://github.com/prometheus/blackbox_exporter/blob/v0.16.0/vendor/github.com/prometheus/common/config/http_config.go#L43


brian-brazil:
Thanks!

PR #610, opened 2020-04-28
Last ssl cert expiry

Novota15:
Checks the expiry date of the ssl certificate at the end of the chain.


brian-brazil:
This has the same issue as #483, just because a cert is the last one in the list doesn't mean that it's part of the relevant chain.

vitalyu:
This fix is very helpful! Thanks @Novota15 
Can you done all requirements so it can be merged to master?

brian-brazil:
As I've already explained on #483 PRs like this which only consider the last cert in the chain will not be accepted.

Issue #624, opened 2020-05-16
pass server_name as query parameter


sepich:
### blackbox_exporter version: output of `blackbox_exporter -version`
0.16.0

We have to monitor many certificates installed on groups of hosts and being managed separately.
For example multiple backends behind single Load Balancer, answering by the same name.
Another use-case is monitoring HTTPS port on nodes behind Load Balancer, where certificate does not have specific node names, but only main VIP name.

Right now we have to manage both `blackbox-exporter` and `prometheus` configs like this:

### What is the blackbox.yml module config.
```yaml
# many sections like
  elk:
    prober: tcp
    timeout: 5s
    tcp:
      preferred_ip_protocol: ipv4
      tls: true
      tls_config:
        server_name: es.domain.local
# basically only differ in server_name
```

### What is the prometheus.yml scrape config.
```yaml
- job_name: blackbox
  metrics_path: /probe
  static_configs:
# many groups like this
    - labels:
        service: elk
        __params_module: [elk]
      targets: # es.domain.local
        - es-logs-node1.domain.local:443
        - es-logs-node2.domain.local:443
        - es-logs-node3.domain.local:443
        - es-logs-node1.domain.local:9300
        - es-logs-node2.domain.local:9300
        - es-logs-node3.domain.local:9300
        - es-logs-warm-node1.domain.local:9300
# only differ in hosts and module name
  relabel_configs:
  ...
```

### What did you expect to see?
Having ability to pass `server_name` via query parameter (like `module`) would greatly simplify things. So, we could manage config only on `prometheus` side.

brian-brazil:
This has been requested in many forms, but this is not something that is going to be added. Each configuration option only comes from one place, and as your example shows you can already do this via modules. This is a matter for your configuration management.

AndresPineros:
Just because a problem can be solved, doesn't mean that design can't be improved. Imagine if I said "ServiceMonitors aren't needed because the person can just manually pass configurations to a prometheus.yaml".... Well, yeha, but why centralize a configuration when it could be extended by dynamic environments?

roidelapluie:
Hello, hostname can now be passed as query parameter, as of last release, which I expect should fix this.

PR #625, opened 2020-05-16
add server_name param

sepich:
Signed-off-by: Alexander Ryabov <i@sepa.spb.ru>

Closes: https://github.com/prometheus/blackbox_exporter/issues/624
Add ability to pass `server_name` for http/tcp probers via query string parameter `server_name`, which will override value set in config (if any) for this specific request.

### DIFF
--- a/main.go
+++ b/main.go
@@ -78,6 +78,19 @@ func probeHandler(w http.ResponseWriter, r *http.Request, c *config.Config, logg
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thank you for your PR. There are no plans to make any settings configurable other than via the config file, as that ways lies madness and open proxies. You can already create a module per setting you need.

PR #630, opened 2020-05-27
Keep blackbox exporter's address in instance label

yann-soubeyrand:
This avoids conflicts when two instances of the blackbox exporter are probing the same target.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ scrape_configs:
       - source_labels: [__address__]
         target_label: __param_target
       - source_labels: [__param_target]
-        target_label: instance
+        target_label: target
       - target_label: __address__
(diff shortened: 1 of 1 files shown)


brian-brazil:
This wouldn't allow you to distinguish targets. Usually it's the job name you'd vary here.

yann-soubeyrand:
You're right, my bad. I fixed it.

Isn't it interesting to have the blackbox exporter itself as instance to be able to correlate its own metrics with the metrics related to probing?

Also, semantically speaking, this is the blackbox which is producing the metrics, not the target, and it appeared more logical to me to have the blackbox in the instance label.

brian-brazil:
The blackbox exporter is an exporter, it's the actual target which you're getting metrics about. I don't see a need to change this example.

PR #633, opened 2020-06-05
Fix the HTTP/2 example

nightmared:
I spent some time trying to understand why I was getting errors like `msg="Invalid HTTP version number" version=2` while my config looked fine: `valid_http_versions: ["HTTP/1.1", "HTTP/2"]`.

Turns out that HTTP/2 is not called HTTP/2 but HTTP/2.0 in the golang HTTP client, so this fixes the example to showcase that.

### DIFF
--- a/example.yml
+++ b/example.yml
@@ -3,7 +3,7 @@ modules:
     prober: http
     timeout: 5s
     http:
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks!

PR #635, opened 2020-06-06
[dns] Allow to specify query Class

Annih:
**Context**
In some cases we might want to probe other DNS class than the INET one; for instance we could probe the server version often exposed in a `TXT CH version.bind` record.

This PR offers the ability to do so by setting `query_class` in your prober configuration, the default is `"IN"` as it was before.

**Implementation note**
Most of the code is a copy/paste/replace of the `query_type` setting; except that the `setQuestion` call has been replaced by its "[slightly more verbose, but more flexible](https://godoc.org/github.com/miekg/dns)" alternative.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -156,6 +156,7 @@ tls_config:
 query_name: <string>
 
 [ query_type: <string> | default = "ANY" ]
+[ query_class: <string> | default = "IN" ]
 
 # List of valid response codes.
 valid_rcodes:
--- a/prober/dns.go
+++ b/prober/dns.go
@@ -141,6 +141,16 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 	registry.MustRegister(probeDNSAuthorityRRSGauge)
 	registry.MustRegister(probeDNSAdditionalRRSGauge)
 
+	qc := uint16(dns.ClassINET)
+	if module.DNS.QueryClass != "" {
+		var ok bool
+		qc, ok = dns.StringToClass[module.DNS.QueryClass]
+		if !ok {
+			level.Error(logger).Log("msg", "Invalid query class", "Class seen", module.DNS.QueryClass, "Existing classes", dns.ClassToString)
+			return false
+		}
+	}
+
 	qt := dns.TypeANY
 	if module.DNS.QueryType != "" {
 		var ok bool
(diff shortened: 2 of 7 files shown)


example.yml
```diff
@@ -129,6 +129,15 @@ modules:
     dns:
       query_name: "prometheus.io"
       query_type: "SOA"
+  dns_bind_version:
```
brian-brazil, line 132: This isn't really testing anything about the version. This is a niche enough use case that we can probably do without an example in the config.

prober/dns.go
```diff
@@ -141,6 +141,16 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 	registry.MustRegister(probeDNSAuthorityRRSGauge)
 	registry.MustRegister(probeDNSAdditionalRRSGauge)
 
+	qc := uint16(dns.ClassINET)
+	if module.DNS.QueryClass != "" {
+		var ok bool
+		qc, ok = dns.StringToClass[module.DNS.QueryClass]
+		if !ok {
+			level.Error(logger).Log("msg", "Invalid query class", "Class seen", module.DNS.QueryClass, "Existing classes", dns.ClassToString)
```
brian-brazil, line 149: This should be done in the config so that it can be caught when the config is loaded.

CONFIGURATION.md
```diff
@@ -153,6 +153,8 @@ tls_config:
 
 [ transport_protocol: <string> | default = "udp" ] # udp, tcp
 
+[ query_class: <string> | default = "IN" ]
```
brian-brazil, line 156: I'd put this below the type, as it'll be less common.

prober/dns.go
Annih: Would that mean the [QueryType check L148](https://github.com/prometheus/blackbox_exporter/pull/635/files/1f04ae59feacb65cd358ff52a665b9e2da193dc6#diff-26673a9bcc326ad193c4d9832bfc4798L148) (should be also done in the config?
Would you accept a 2nd PR to move both checks outside this file?

example.yml
Annih: Removed :)

CONFIGURATION.md
Annih: Moved :)

prober/dns.go
brian-brazil: Indeed, it should be. It's a small enough change it can all be done here.

brian-brazil:
Did you get a chance to make that final change? I'd like to get a release out soon and would like to include this PR.

prober/dns.go
Annih: done, sorry for the delay.
I added config checks but also kept the existing code in the prober.

brian-brazil:
Thanks!

PR #636, opened 2020-06-07
Add new probe_ssl_latest_verified_chain_expiry metric

itkq:
Resolves https://github.com/prometheus/blackbox_exporter/issues/340

Based on the discussion in the issue above, this metric will help determine
when the SSL/TLS certificate expiration error actually happens on clients
like a browser that attempts to verify certificates by building one or
more chains from peer certificates.

### DIFF
--- a/prober/tcp.go
+++ b/prober/tcp.go
@@ -94,9 +94,9 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_ssl_earliest_cert_expiry",
 		Help: "Returns earliest SSL cert expiry date",
 	})
-	probeSSLLatestVerifiedChainExpiry := prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "probe_ssl_latest_verified_chain_expiry",
-		Help: "Returns latest SSL verified chain expiry date",
+	probeSSLLastChainExpiryTimestampSeconds := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_ssl_last_chain_expiry_timestamp_seconds",
+		Help: "Returns last SSL chain expiry in timestamp seconds",
 	})
 	probeTLSVersion := prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
@@ -129,9 +129,9 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 	}
 	if module.TCP.TLS {
 		state := conn.(*tls.Conn).ConnectionState()
-		registry.MustRegister(probeSSLEarliestCertExpiry, probeSSLLatestVerifiedChainExpiry, probeTLSVersion)
+		registry.MustRegister(probeSSLEarliestCertExpiry, probeSSLLastChainExpiryTimestampSeconds, probeTLSVersion)
 		probeSSLEarliestCertExpiry.Set(float64(getEarliestCertExpiry(&state).Unix()))
-		probeSSLLatestVerifiedChainExpiry.Set(float64(getLatestVerifiedChainExpiry(&state).Unix()))
+		probeSSLLastChainExpiryTimestampSeconds.Set(float64(getLastChainExpiry(&state).Unix()))
 		probeTLSVersion.WithLabelValues(getTLSVersion(&state)).Set(1)
 	}
 	scanner := bufio.NewScanner(conn)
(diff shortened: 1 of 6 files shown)


prober/tls.go
```diff
@@ -28,6 +28,23 @@ func getEarliestCertExpiry(state *tls.ConnectionState) time.Time {
 	return earliest
 }
 
+func getLatestVerifiedChainExpiry(state *tls.ConnectionState) time.Time {
+	latestVerifiedChainExpiry := time.Time{}
```
brian-brazil, line 32: What does this output if there's no chains? You'll probably need insecure_skip_verify to test this.

prober/tcp.go
```diff
@@ -94,6 +94,10 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_ssl_earliest_cert_expiry",
 		Help: "Returns earliest SSL cert expiry date",
 	})
+	probeSSLLatestVerifiedChainExpiry := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_ssl_latest_verified_chain_expiry",
```
brian-brazil, line 98: That's a little of a mouthful, and naming conventions have matured since the previous one was named.

I'd go for something like probe_ssl_last_chain_expiry_timestamp_seconds.

itkq:
Changed the name of the metric a bit and also added a unit test.

prober/tcp.go
```diff
@@ -129,10 +129,14 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
...
 	if module.TCP.TLS {
 		state := conn.(*tls.Conn).ConnectionState()
-		registry.MustRegister(probeSSLEarliestCertExpiry, probeSSLLatestVerifiedChainExpiry, probeTLSVersion)
+		registry.MustRegister(probeSSLEarliestCertExpiry, probeTLSVersion)
 		probeSSLEarliestCertExpiry.Set(float64(getEarliestCertExpiry(&state).Unix()))
-		probeSSLLatestVerifiedChainExpiry.Set(float64(getLatestVerifiedChainExpiry(&state).Unix()))
 		probeTLSVersion.WithLabelValues(getTLSVersion(&state)).Set(1)
+
+		if lastChainExpiry, ok := getLastChainExpiry(&state); ok {
+			registry.MustRegister(probeSSLLastChainExpiryTimestampSeconds)
```
brian-brazil, line 137: Let's keep this consistent with the other tls metric, both in terms of registration and what happens when there's no certs.

prober/tcp.go
```diff
@@ -94,9 +94,9 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_ssl_earliest_cert_expiry",
 		Help: "Returns earliest SSL cert expiry date",
 	})
-	probeSSLLatestVerifiedChainExpiry := prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "probe_ssl_latest_verified_chain_expiry",
-		Help: "Returns latest SSL verified chain expiry date",
+	probeSSLLastChainExpiryTimestampSeconds := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_ssl_last_chain_expiry_timestamp_seconds",
+		Help: "Returns last SSL chain expiry in timestamp seconds",
```
brian-brazil, line 99: in unixtime

Looks like the existing metric help could do with some consistency too.

prober/tcp_test.go
```diff
@@ -193,6 +201,158 @@ func TestTCPConnectionWithTLS(t *testing.T) {
...
+
+	rootPrivatekey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(fmt.Sprintf("Error creating rsa key: %s", err))
+	}
+
+	// Prepare certificates to simulate a situation where
+	// the root certificate in the chain sent by the server
+	// expired but a verified certificate chain can be found
+	// because a new one has installed on the local machine.
```
brian-brazil, line 227: "one" is ambiguous here. Root?


This also feels like we're testing Go's root handling rather than testing our own code. Having two chains where one is expired would seem like a better test.

itkq: Right. I made a change to do that in https://github.com/prometheus/blackbox_exporter/pull/636/commits/d45159b5485d39fd06cdad71f2c9fc56c630a6ff

brian-brazil:
Thanks!

PR #643, opened 2020-06-17
Add support for DoT to DNS probes

edevil:
This adds a new DNS probe option `dns_over_tls` to enable DoT on DNS probes.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -153,6 +153,9 @@ tls_config:
 
 [ transport_protocol: <string> | default = "udp" ] # udp, tcp
 
+# Whether to use DNS over TLS. This only works with TCP.
+[ dns_over_tls: <boolean | default = false> ]
+
 query_name: <string>
 
 [ query_type: <string> | default = "ANY" ]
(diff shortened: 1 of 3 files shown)


brian-brazil:
Hmm, I can see this being confusing if it silently doesn't do anything when UDP is asked for.

Docs also need updating.

edevil:
@brian-brazil do you want me to error out when UDP and DoT is chosen? I’ll update the docs shortly.

brian-brazil:
An error would do it.

edevil:
@brian-brazil Added error case, and docs.

CONFIGURATION.md
```diff
@@ -153,6 +153,9 @@ tls_config:
 
 [ transport_protocol: <string> | default = "udp" ] # udp, tcp
 
+# Whether to use DNS over TLS. This only makes sense with TCP transport protocol.
```
brian-brazil, line 156: This only works with TCP.

edevil: Done.

brian-brazil:
Thanks!

PR #644, opened 2020-06-17
Add option to disable TLS certificate check in DoT

edevil:
Similar to TLS probe

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -156,6 +156,10 @@ tls_config:
 # Whether to use DNS over TLS. This only works with TCP.
 [ dns_over_tls: <boolean | default = false> ]
 
+# Configuration for TLS protocol of DNS over TLS probe.
+tls_config:
+  [ <tls_config> ]
+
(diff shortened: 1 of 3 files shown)


brian-brazil:
Should we have all of the standard tls config stuff here?

edevil:
@brian-brazil I had that question. But since there already is a TLS probe with those options, I suppose the logical thing is to use that one to test for custom TLS stuff, if you need it, and just disable TLS verification in DoT.

In other words, use the DNS and TLS probe in tandem. What do you think?

brian-brazil:
My point is that we already have a standard way of producing a `tls.Config`, so you should use that.

edevil:
I added the TLSConfig block to the DNS probe.

brian-brazil:
Thanks!

edevil:
@brian-brazil No, thank you! Are there any plans to release a new version of blackbox_exporter?

brian-brazil:
Hopefully in the next few days.

PR #645, opened 2020-06-19
Deal with URLs with literal IPv6 addresses

edevil:
Recent versions of Go have started enforcing the usage of
brackets in URLs with literal IPv6 addresses. This change
prevents the attempt to resolve an address in the form
[address]. URL.Hostname() takes care of striping the brackets.

Signed-off-by: André Cruz <acruz@cloudflare.com>


brian-brazil:
Thanks!

PR #652, opened 2020-06-25
Don't return probe_duration_seconds on timeout

edevil:
When a timeout occurs it does not make sense to return `probe_duration_seconds`, since the probe did not actually finish.
This will just skew the metrics with values == timeout.

What do you think?


brian-brazil:
Thanks for your PR, I don't think this makes sense. That a probe failed does not make its duration cease to exist, so the metric is still valid. If you want to exclude these then PromQL would be a better place to do that.

edevil:
Fair enough!

Issue #646, opened 2020-06-19
DNS-over-TLS requests should default to port 853 (not 53)


appliedprivacy:
**Thank you for adding DoT support!**

DNS-over-TLS's is [specified](https://tools.ietf.org/html/rfc7858) to operate on port 853/tcp.

blackbox exporter tries to talk DoT on port 53 by default, which is reserved for plain DNS.
 
We are proposing to change blackbox_exporter's default port of DoT connections to 853, when `dns_over_tls` is `true`.
even though it is easy to workaround this issue by adding the port to the target
eg. dot1.appliedprivacy.net:853 

After implementing this change it should still be possible to specify the port via the target (since we want to test our DoT service on port 443 as well).

### Host operating system:

```4.19.0-9-amd64 #1 SMP Debian 4.19.118-2+deb10u1 (2020-06-07) x86_64 GNU/Linux```

### blackbox_exporter version: 

```
blackbox_exporter, version 0.17.0 (branch: HEAD, revision: 1bc768014cf6815f7e9d694e0292e77dd10f3235)
  build user:       root@626fb3899f41
  build date:       20200619-11:54:41
  go version:       go1.14.4
```

### What is the blackbox.yml module config.

```
  dns_tcp:
    prober: dns
    dns:
      preferred_ip_protocol: "ip4" # defaults to "ip6"
      query_name: "www.prometheus.io"
      query_type: "A"
      transport_protocol: "tcp"
      dns_over_tls: true
      tls_config:
        ca_file: "/etc/blackbox_exporter/DSTRootCAX3.crt"
        server_name: "dot1.appliedprivacy.net"
```

### What did you do that produced an error?

```
curl "http://localhost:9115/probe?target=dot1.appliedprivacy.net&module=dns_tcp&debug=true"
```

### What did you expect to see?

from debug log:
```
caller=main.go:169 module=dns_tcp target=dot1.appliedprivacy.net level=debug msg="Making DNS query" target=94.130.106.88:853 dial_protocol=tcp4-tls query=www.prometheus.io type=1 class=1
```

### What did you see instead?

from the debug log:
```
caller=main.go:169 module=dns_tcp target=dot1.appliedprivacy.net level=debug msg="Making DNS query" target=94.130.106.88:53 dial_protocol=tcp4-tls query=www.prometheus.io type=1 class=1
```

brian-brazil:
Would you like to send a PR?

appliedprivacy:
[...]

PR #655, opened 2020-06-26
Change DoT default port to 853

kkc:
This change the default port of DNS-over-TLS to 853.
Fix https://github.com/prometheus/blackbox_exporter/issues/646


brian-brazil:
Thanks!

PR #656, opened 2020-06-26
Conditionally register and set probe_ssl_last_chain_expiry_timestamp_seconds

### LINKED ISSUE #653: Negative timestamps in probe_ssl_last_chain_expiry_timestamp_seconds
### blackbox_exporter version:
v0.17.0

Replaced `probe_ssl_earliest_cert_expiry ` with `probe_ssl_last_chain_expiry_timestamp_seconds` in our alert definitions to fix issue #340 

A positive timestamp, equal to `probe_ssl_earliest_cert_expiry` in most cases.

### What did you see instead?
Value `-62135596800` everywhere -- the value does not seem to ever change, unlike `probe_ssl_earliest_cert_expiry`. The formatted version of that timestamp is `1677-09-21 00:12:43.146 +0000 UTC`.

bgagnon:
If verification is disabled, there won't get any chains to verify, and a
timestamp value cannot be produced. Skip the metric altogether.

Fixes #653

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -549,10 +549,13 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 
 	if resp.TLS != nil {
 		isSSLGauge.Set(float64(1))
-		registry.MustRegister(probeSSLEarliestCertExpiryGauge, probeTLSVersion, probeSSLLastChainExpiryTimestampSeconds)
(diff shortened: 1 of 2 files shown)


brian-brazil:
I think it'd be better to return 0, so that if you are in this state your alerts will still tell you that something is wrong.

mem:
@bgagnon any responses to Brian's most recent comment?

Issue #658, opened 2020-06-30
Failed probe with HTTP/2 included in valid_http_versions


allenyhu:
### Host operating system: output of `uname -a`
Darwin Kernel Version 19.5.0
### blackbox_exporter version: output of `blackbox_exporter -version`

0.17.0
### What is the blackbox.yml module config.
```
modules:
    http_2xx:
      prober: http
      timeout: 30s
      http:
        valid_http_versions: ["HTTP/1.1", "HTTP/2"]
        fail_if_ssl: false
        fail_if_not_ssl: true
        no_follow_redirects: false
        preferred_ip_protocol: "ip4"
```
### What is the prometheus.yml scrape config.
```
scrape_configs: 
    - job_name: 'blackbox'
      metrics_path: /probe
      params:
        module: [http_prometheus]
      static_configs:
        - targets:
          - prometheus.io
      relabel_configs:
        - source_labels: [__address__]
           target_label: __param_target
        - source_labels: [__param_target]
           target_label: instance
        - target_label: __address__
           replacement: 127.0.0.1:9115
```
### What logging output did you get from adding `&debug=true` to the probe URL?
```
Logs for the probe:
ts=2020-06-30T19:33:07.400956Z caller=main.go:304 module=http_2xx target=prometheus.io level=info msg="Beginning probe" probe=http timeout_seconds=30
ts=2020-06-30T19:33:07.401093Z caller=http.go:323 module=http_2xx target=prometheus.io level=info msg="Resolving target address" ip_protocol=ip4
ts=2020-06-30T19:33:07.482731Z caller=http.go:323 module=http_2xx target=prometheus.io level=info msg="Resolved target address" ip=104.24.120.60
ts=2020-06-30T19:33:07.482937Z caller=client.go:252 module=http_2xx target=prometheus.io level=info msg="Making HTTP request" url=http://104.24.120.60 host=prometheus.io
ts=2020-06-30T19:33:07.545754Z caller=client.go:474 module=http_2xx target=prometheus.io level=info msg="Received redirect" location=https://prometheus.io/
ts=2020-06-30T19:33:07.545903Z caller=client.go:252 module=http_2xx target=prometheus.io level=info msg="Making HTTP request" url=https://prometheus.io/ host=
ts=2020-06-30T19:33:07.545939Z caller=client.go:252 module=http_2xx target=prometheus.io level=info msg="Address does not match first address, not sending TLS ServerName" first=104.24.120.60 address=prometheus.io
[...]

PR #659, opened 2020-07-01
Log http version string we compare with, rather than float we put in metric.

brian-brazil:
See #658


roidelapluie:
:+1:

PR #662, opened 2020-07-09
Expose probe_dns_duration_seconds metric

mem:
This follows the same pattern as probe_http_duration_seconds and
probe_icmp_duration_seconds: it captures the time it takes to perform
the actual request, excluding the time it takes to set up things.
probe_duration_seconds includes the time to perform the request as well
as all the time it takes to set it up. probe_dns_lookup_time_seconds
refers to the time it takes to resolve the target's address, so it
doesn't capture the time it takes to make one request for that target.

### DIFF
--- a/prober/dns.go
+++ b/prober/dns.go
@@ -126,6 +126,10 @@ func validRcode(rcode int, valid []string, logger log.Logger) bool {
 
 func ProbeDNS(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) bool {
 	var dialProtocol string
+	probeDNSDurationGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_dns_duration_seconds",
+		Help: "Duration of DNS request",
+	})
 	probeDNSAnswerRRSGauge := prometheus.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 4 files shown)


prober/dns.go
```diff
@@ -246,7 +251,14 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 	level.Info(logger).Log("msg", "Making DNS query", "target", targetIP, "dial_protocol", dialProtocol, "query", module.DNS.QueryName, "type", qt, "class", qc)
 	timeoutDeadline, _ := ctx.Deadline()
 	client.Timeout = time.Until(timeoutDeadline)
+	// we don't use the rtt value returned fom client.Exchange because that
+	// includes only the time to exchange messages with the server _after_
+	// the connection is created. We would in principle have three phases:
+	// resolve (probe_dns_lookup_time_seconds), connect (request - rtt),
```
brian-brazil, line 257: To keep things similar to the metric in the other probes, we should break this out into its phases.

mem: OK, will do

mem: I added the phase label and added the three time measurements, as indicated in the original comment.

brian-brazil:
Thanks!

Issue #166, opened 2017-06-13
Please document what metrics are exposed


nafg:

brian-brazil:
It's not normal to document metrics, as they come from a number of different pieces of code which you rarely have control of all of. The HELP and name of the metric should give you enough to work with, and at the end of the day the only real way to understand a metric is to look at the code. Duplicating that elsewhere is just going to get stale.

nafg:
How is it different than any API? What's the point of the readme or any
prose docs if the code will change?

SuperQ:
I disagree, we should have good reference docs for our exporters.  "Read the code" is not sufficient.

brian-brazil:
In my experience read the code is the only thing that ultimately works for metrics. Every attempt I've seen to comprehensively document metrics has failed.

nafg:
The same is true of APIs, yet we​ try...

brian-brazil:
APIs are (usually) explicitly for consumption by others, whereas most metrics are added on a whim for potential future use in debugging/analysis.

SuperQ:
For the blackbox_exporter, it's very much an API.

brian-brazil:
There's a fair few subtleties, which basically boils down to re-rewriting the code in English.

nafg:
Prometheus is a debugging tool?

If there are such subtleties doesn't that make it more important?

Would "rewriting the code in English" make it easier to maintain
documentation... or are you conceding that currently reading the code is
not a solution?

All I want is to set up Prometheus alerting when one of my servers goes
down. Is that like an unsupported corner case??

brian-brazil:
If the metric name and HELP aren't enough, it's unlikely anything short of the source code will help you.

> If there are such subtleties doesn't that make it more important?
> Would "rewriting the code in English" make it easier to maintain
documentation... or are you conceding that currently reading the code is
not a solution?

Such subtleties mean you'd just be writing the code a 2nd time in English, which would add maintenance burden without much benefit to the user.

> All I want is to set up Prometheus alerting when one of my servers goes
down. Is that like an unsupported corner case??

That is supported.That is not however what you have requested in this issue.

nafg:
I want a list of the metric names and what they do. Yes it is included in
the metrics page, but I needed it before it was running. (Everything is
managed via salt in version control and applied on push. I didn't want to
have to push twice and wait each time.) And I didn't expose the port to the
world so I can't easily see them. And how do I know the metrics I'm seeing
[...]

Issue #601, opened 2020-04-17
example of the metrics that are returned


ghostsquad:
Could an example of the metrics that are returned be provided somewhere? maybe also link in the README?

matclab:
Also see #166 and PR #663.

ghostsquad:
ah.. from #166 the very tired argument of "we don't need docs, just read the code". Documentation is like sex. Even bad documentation is better than nothing.

reegnz:
Examples, documentation, tutorials... The argument is always 'Just read the code dummy, why isn't that enough? Duplication BAD! DRY GOOD!'.

ghostsquad:
Made an argument to reopen the PR and continue the path of providing metrics documentation here https://github.com/prometheus/blackbox_exporter/pull/663?ts=4#issuecomment-658853584

electron0zero:
closing this in favour of https://github.com/prometheus/blackbox_exporter/issues/856, please comment in https://github.com/prometheus/blackbox_exporter/issues/856 if you would like to discuss this issue.

Issue #241, opened 2017-10-05
Document metric names and meanings exposed by the probers


reegnz:
As a user, I would like to have a clear reference about what metrics each probe exposes and what each metric means.
To get to this information I currently either need to grok the source code.
It would be desirable to have documentation about it eg. in an .md file.

Maxeh:
Is there still no documentation? There are so many commands and I just do not know what they are there for? How should I know, how all the probers work and what to do with them? Does someone has a cheat sheet?

matclab:
Also see #166 and PR #663.

reegnz:
This is still open? LOL!
I'm going to close this issue, I've given up hope years ago. Too many resistance from the maintainers for this to ever get proper traction... 
#663 just shows that usability is not a priority for this project. The documentation is **generated from the code** FFS.

Sad.

PR #663, opened 2020-07-13
166 auto document metrics

matclab:
This PR enable to generate a markdown files `metric.md` describing the metrics collected by blackbox_exporter.

The document is auto-generatd when doing `make` and a job is added to the CI to ensure that the commited `metrics.md` is up to date (Note, that I'm not sure about the circle-ci config file and I'm not sure how to test it, so it may not works at first try).

See #166.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,10 +29,27 @@ jobs:
           sudo service docker restart
     - run: docker run --rm -t -v "$(pwd):/app" "${DOCKER_TEST_IMAGE_NAME}" -i github.com/prometheus/blackbox_exporter -T
 
+  checkdoc:
+    # Ensure that doc is uptodate in git repo
+    docker:
+      - image: alpine/make:latest
+    steps:
+      - checkout
+      - run: 
+          name: "verify metrics.md is up to date"
+          command: |
+            cp metrics.md  metrics.md.git
+            make metrics.md
+            diff metrics.ms metrics.md.git
+
 workflows:
   version: 2
   blackbox_exporter:
     jobs:
+    - checkdoc:
+        filters:
+          tags:
+            only: /.*/
     - test:
         filters:
           tags:
@@ -45,6 +62,7 @@ workflows:
     - prometheus/publish_master:
         context: org-context
         requires:
+        - checkdoc
         - test
         - build
         filters:
@@ -53,6 +71,7 @@ workflows:
     - prometheus/publish_release:
         context: org-context
         requires:
+        - checkdoc
         - test
         - build
         filters:
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,15 @@
 # Needs to be defined before including Makefile.common to auto-generate targets
 DOCKER_ARCHS ?= amd64 armv7 arm64 ppc64le
 
+GOFILES = ${shell find . -name "*.go" | grep -v "^\./vendor"}
+
+allanddoc: all metrics.md
+
 include Makefile.common
 
 DOCKER_IMAGE_NAME       ?= blackbox-exporter
+
+
+metrics.md: ${GOFILES} scripts/extract_metric_doc.sh
+	scripts/extract_metric_doc.sh ${GOFILES} > $@
+
--- a/README.md
+++ b/README.md
@@ -43,6 +43,9 @@ will return debug information for that probe.
 After a successful local build:
 
     docker build -t blackbox_exporter .
+## Metrics
+
+The blackbox_exporter metrics are documented in [this file](metrics.md).
 
 ## [Configuration](CONFIGURATION.md)
 
--- a/metrics.md
+++ b/metrics.md
@@ -1,0 +1,58 @@
+# Blackbox Exporter Metric Documentation
+
+*Those tables are automatically extracted from source code. Do not modify
+manually!*
+
+
+## Dns metrics
+
+| metric name              | doc                                                             |
+|--------------------------|-----------------------------------------------------------------|
+| probe_dns_additional_rrs | Returns number of entries in the additional resource record list|
+| probe_dns_answer_rrs     | Returns number of entries in the answer resource record list    |
+| probe_dns_authority_rrs  | Returns number of entries in the authority resource record list |
+| probe_dns_serial         | Returns the serial number of the zone                           |
+
+## General metrics
+
+| metric name                                  | doc                                                                               |
+|----------------------------------------------|-----------------------------------------------------------------------------------|
+| config_last_reload_successful                | Blackbox exporter config loaded successfully.                                     |
+| config_last_reload_success_timestamp_seconds | Timestamp of the last successful configuration reload.                            |
+| probe_dns_lookup_time_seconds                | Returns the time taken for probe dns lookup in seconds                            |
+| probe_duration_seconds                       | Returns how long the probe took to complete in seconds                            |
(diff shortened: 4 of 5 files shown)


brian-brazil:
Thanks for your PR. An listing of HELP strings doesn't seem very useful to me, as they're only one-liner descriptions you can already get from the exporter itself or grepping the source code. To be useful the documentation needs to describe what these metrics actually are, including their full semantics. Additionally it should not be mixing exporter, overall probe, and prober metrics.

Something like this would need to be manually created and maintained.

matclab:
> Thanks for your PR. An listing of HELP strings doesn't seem very useful to me, as they're only one-liner descriptions you can already get from the exporter itself or grepping the source code. To be useful the documentation needs to describe what these metrics actually are, including their full semantics. Additionally it should not be mixing exporter, overall probe, and prober metrics.

I agree that a documentation with full semantic would be better. Nonetheless, grepping in the source code is not user firendly, and to get the information from the exporter you should already have one running.

The `metrics.md` file allows for people looking for a solution to their problem to see if blackbox_exporter may be a solution or not. This was a need I had at the time. It was very difficult for me to understand exactly what I will get with blackbox_exporter.

According to #601, #241, and the numerous comments of #166 I understand this is a frequent request, and this is why I propose this PR.
Also maybe that *browse the sources* is not really what we want for a successful open source project.

Perhaps we can improve on this PR by adding a comment in the code beside the "Help" and "Name"  that we could automatically extract and that will provide explanation of the full semantic. However I think that the `metrics.md` file is already useful.

You can find example of such simple documentation provided by other prometheus exporters: [memcached](https://github.com/prometheus/memcached_exporter#user-content-collectors), [elasticsearch](https://github.com/justwatchcom/elasticsearch_exporter#user-content-metrics), [couchdb](https://github.com/gesellix/couchdb-prometheus-exporter/blob/master/README_metrics.md), [rabbitmq](https://github.com/kbudde/rabbitmq_exporter#user-content-metrics), [apache](https://github.com/Lusitaniae/apache_exporter#user-content-collectors),…

Of course, this is just a proposal and you do what you deem useful with it.

Happy codding.

brian-brazil:
> According to #601, #241, and the numerous comments of #166 I understand this is a frequent request, and this is why I propose this PR.

"Documentation" is a request, but there's quite a bit of variance in what people mean by that. Just because something classifies as documentation doesn't mean it'll meet their stated need.

> The metrics.md file allows for people looking for a solution to their problem to see if blackbox_exporter may be a solution or not.

The configuration docs already indicate what it can check. The metrics are largely for forensics after a probe has failed.

> Perhaps we can improve on this PR by adding a comment in the code beside the "Help" and "Name" that we could automatically extract and that will provide explanation of the full semantic

At that point you're duplicating the code in English. The blackbox exporter is not a normal exporter in many ways, what it does is not equivalent to direct instrumentation due to the various configured and runtime modes to the extent that this sort of listing would be quite misleading. I'm dubious of simple listings of metrics for other exporters, and it's going to be even less useful here as there's just so many ways the metrics can actually turn out. For example different things happen metric-wise depending on whether the probe ends up using TLS or not - which may only be determined at runtime.

matclab:
> > The metrics.md file allows for people looking for a solution to their problem to see if blackbox_exporter may be a solution or not.
> 
> The configuration docs already indicate what it can check. The metrics are largely for forensics after a probe has failed.

Do you mean that `probe_tls_version_info`, `probe_ssl_last_chain_expiry_timestamp_seconds`, `probe_ssl_earliest_cert_expiry` or `probe_http_status_code`  should be use only for debugging when probe failed and not for monitoring/alerting ? 
I'd like to know, because today I'm using alerting rules like the one promoted on [awesome prometehus alerts](https://awesome-prometheus-alerts.grep.to/rules#blackbox).

> > Perhaps we can improve on this PR by adding a comment in the code beside the "Help" and "Name" that we could automatically extract and that will provide explanation of the full semantic
> 
> At that point you're duplicating the code in English. 

Indeed, but isn't what is the doc about at the end?  What about duplicating `config.go` in `CONFIGURE.md`? (note I'm really happy this doc exists).

> The blackbox exporter is not a normal exporter in many ways, what it does is not equivalent to direct instrumentation due to the various configured and runtime modes to the extent that this sort of listing would be quite misleading. I'm dubious of simple listings of metrics for other exporters, and it's going to be even less useful here as there's just so many ways the metrics can actually turn out. For example different things happen metric-wise depending on whether the probe ends up using TLS or not - which may only be determined at runtime.

As you can guess, this is not clear to me as I'm not familiar with the code base. You are certainly the one who knows. I won't comment anymore as I think you have exposed your points and our time is precious.

Anyway thank you very much for this nice piece of software and for having taken time to discuss this subject. Do not hesitate to ask if you think I can be of some help on this subject on a way or another.

brian-brazil:
> should be use only for debugging when probe failed and not for monitoring/alerting ?

The two cert timestamp ones are explicitly for alerting. Status code checking is already supported via configuration, you shouldn't be using metrics for that (and that alert you linked is fragile for other reasons). TLS version is a new one to me, I guess if that's important it could be added as a check you could do - though that feels a bit more like something that belongs in a more dedicated security scanning tool as if you care about that there's probably quite a few other settings you'd care about which the blackbox exporter doesn't cover.

> Indeed, but isn't what is the doc about at the end? 

Not quite, it depends on the level of the docs. To produce accurate docs you'd need to painstakingly lay out how the metrics work, which boils down to duplicating the code. As I indicated, this is not a typical exporter here - it's not a matter of "this is gauge X returned from command Y" so more simplistic approaches you could take elsewhere may cause more harm than good in this case.

> What about duplicating config.go in CONFIGURE.md?

That is a point, the docs there provide more than the grep in this PR does and semantics are much cleaner for the config though.

> Do not hesitate to ask if you think I can be of some help on this subject on a way or another.

If you want to have a go at producing docs I can review the PRs. I'd advise leaving HTTP to last, that's the most complex one due to all the interlocking features.

ghostsquad:
> "Documentation" is a request, but there's quite a bit of variance in what people mean by that. Just because something classifies as documentation doesn't mean it'll meet their stated need.

So please, let something be created, and let the community decide if it meets their needs. If it does not, improve it. This is what agile is all about.

> A listing of HELP strings doesn't seem very useful to me, as they're only one-liner descriptions you can already get from the exporter itself or grepping the source code.

This is an insane reason to close this PR. He literally _is_ grepping the source code to provide this information. Also, even if it's not useful to you doesn't mean it's not useful to others. Clearly the author @matclab thinks it's useful, which is why he submitted the PR. I'm also another user for the blackbox_exporter, and I'm looking for more information.

> Something like this would need to be manually created and maintained.

If you are so confident that "just look at the source code" is good enough, then there should be a way to automatically extract this relevant information.

> which boils down to duplicating the code.

DRY is a fallacy. If keeping DRY reduces the value you provided, then this is equivalent to shooting yourself in the foot.

> so more simplistic approaches you could take elsewhere may cause more harm than good in this case.

Let the community decide if that's true

> That is a point, the docs there provide more than the grep in this PR does and semantics are much cleaner for the config though.

Let's get _something_ in place and then iterate. It doesn't need be perfect out of the box.

> At that point you're duplicating the code in English.

Yes. This is correct. This provides values. As they say: Documentation is like sex. Even bad documentation is better than nothing.

---

Please reopen this PR.

PR #674, opened 2020-08-13
Fix issue #673: incorrect content length reporting

### LINKED ISSUE #673: Incorrect Content-Length when using regular expression validations

```
Linux clara 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

```
blackbox_exporter, version 0.17.0 (branch: master, revision: 349af4729b7c24eff5589e50fde9ce7dce5cfc23)
  build user:       marcelo@clara
  build date:       20200812-16:57:08
  go version:       go1.14.7
```

```
modules:
  http:
    prober: http
    http:
      method: GET
      fail_if_body_matches_regexp:
        - "Could not connect to database"
      preferred_ip_protocol: "ip4"
      ip_protocol_fallback: false
```

None, accessing `http://localhost:9115/probe?target=https://prometheus.io/&module=http` directly.

```
Logs for the probe:
ts=2020-08-12T17:06:21.028387865Z caller=main.go:304 module=http target=https://prometheus.io/ level=info msg="Beginning probe" probe=http timeout_seconds=119.5
ts=2020-08-12T17:06:21.028635611Z caller=http.go:321 module=http target=https://prometheus.io/ level=info msg="Resolving target address" ip_protocol=ip4
ts=2020-08-12T17:06:21.029658213Z caller=http.go:321 module=http target=https://prometheus.io/ level=info msg="Resolved target address" ip=172.67.141.23
ts=2020-08-12T17:06:21.029751745Z caller=client.go:252 module=http target=https://prometheus.io/ level=info msg="Making HTTP request" url=https://172.67.141.23/ host=prometheus.io
ts=2020-08-12T17:06:21.58301342Z caller=main.go:119 module=http target=https://prometheus.io/ level=info msg="Received HTTP response" status_code=200
ts=2020-08-12T17:06:21.583538001Z caller=main.go:119 module=http target=https://prometheus.io/ level=info msg="Response timings for roundtrip" roundtrip=0 start=2020-08-12T11:06:21.029868789-06:00 dnsDone=2020-08-12T11:06:21.029868789-06:00 connectDone=2020-08-12T11:06:21.082736393-06:00 gotConn=2020-08-12T11:06:21.24414616-06:00 responseStart=2020-08-12T11:06:21.582728926-06:00 end=2020-08-12T11:06:21.583515684-06:00
ts=2020-08-12T17:06:21.583814089Z caller=main.go:304 module=http target=https://prometheus.io/ level=info msg="Probe succeeded" duration_seconds=0.555312085



probe_dns_lookup_time_seconds 0.000999765
probe_duration_seconds 0.555312085
# HELP probe_failed_due_to_regex Indicates if probe failed due to regex
# TYPE probe_failed_due_to_regex gauge
probe_failed_due_to_regex 0
# HELP probe_http_content_length Length of http content response
# TYPE probe_http_content_length gauge
probe_http_content_length -1
# HELP probe_http_duration_seconds Duration of http request by phase, summed over all redirects
# TYPE probe_http_duration_seconds gauge
probe_http_duration_seconds{phase="connect"} 0.05286772
probe_http_duration_seconds{phase="processing"} 0.338582835
probe_http_duration_seconds{phase="resolve"} 0.000999765
probe_http_duration_seconds{phase="tls"} 0.214277414
probe_http_duration_seconds{phase="transfer"} 0.00078672
# HELP probe_http_redirects The number of redirects
# TYPE probe_http_redirects gauge
probe_http_redirects 0
# HELP probe_http_ssl Indicates if SSL was used for the final redirect
# TYPE probe_http_ssl gauge
probe_http_ssl 1
# HELP probe_http_status_code Response HTTP status code
# TYPE probe_http_status_code gauge
probe_http_status_code 200
# HELP probe_http_uncompressed_body_length Length of uncompressed response body
# TYPE probe_http_uncompressed_body_length gauge
probe_http_uncompressed_body_length 0
# HELP probe_http_version Returns the version of HTTP of the probe response
# TYPE probe_http_version gauge
probe_http_version 2
probe_ip_addr_hash 2.601109704e+09
probe_ip_protocol 4
# HELP probe_ssl_earliest_cert_expiry Returns earliest SSL cert expiry in unixtime
# TYPE probe_ssl_earliest_cert_expiry gauge
probe_ssl_earliest_cert_expiry 1.628424e+09
# HELP probe_ssl_last_chain_expiry_timestamp_seconds Returns last SSL chain expiry in timestamp seconds
# TYPE probe_ssl_last_chain_expiry_timestamp_seconds gauge
probe_ssl_last_chain_expiry_timestamp_seconds 1.628424e+09
# TYPE probe_success gauge
probe_success 1
# HELP probe_tls_version_info Contains the TLS version used



Module configuration:
prober: http
http:
    preferred_ip_protocol: ip4
    method: GET
    fail_if_body_matches_regexp:
        - Could not connect to database
tcp:
    ip_protocol_fallback: true
icmp:
    ip_protocol_fallback: true
dns:
    ip_protocol_fallback: true
```

Add `fail_if_body_matches_regexp` to config.

Based on this:

```
$ curl -o prometheus.gz -H "Accept-Encoding: gzip" --silent https://prometheus.io/

$ file prometheus.gz
prometheus.gz: gzip compressed data, from Unix, original size modulo 2^32 15908

$ ls -l prometheus.gz
-rw-rw-r-- 1 marcelo marcelo 4585 Aug 12 11:11 prometheus.gz

$ gunzip prometheus.gz

$ ls -l prometheus
-rw-rw-r-- 1 marcelo marcelo 15908 Aug 12 11:11 prometheus
```

I would expect to see:

```
probe_http_content_length 4585
probe_http_uncompressed_body_length 15908
```

or

```
probe_http_content_length -1
probe_http_uncompressed_body_length 15908
```

(because adding the Transfer-Encoding header causes the Content-Length one to not be included)

### What did you see instead?

probe_http_content_length -1
probe_http_uncompressed_body_length 0

mem:
Without regexp validations, the current code reads the response from the
server in order to figure out the number of bytes we received. With
regular expression validations, it's necessary to read the whole
response in order verify it if matches or not. In this case, since the
response is completely consumed, the current code is reporting a length
of 0.

Modify the code so that we can keep track of the content length in
either case: wrap the io.ReadCloser that corresponds to the response
body with another io.ReadCloser that keeps track of the number of bytes
read. After completely consuming the reponse, the wrapper is able to
report the total number of bytes read.

Add TestContentLength that checks the content lenght is reported
correctly. This includes checking cases with content encoding values we
might get back from servers (brotli, deflate, gzip and identity) to make
sure we don't change behavior if one of the underlying libraries does.

Add code in existing tests dealing the regular expression validations to
avoid regressing issue 673.

Fixes #673

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -226,6 +226,19 @@ func (t *transport) GotFirstResponseByte() {
 	t.current.responseStart = time.Now()
 }
 
+// byteCounter implements an io.ReadCloser that keeps track of the total
+// number of bytes it has read.
+type byteCounter struct {
+	io.ReadCloser
+	n int64
+}
+
+func (bc *byteCounter) Read(p []byte) (int, error) {
+	n, err := bc.ReadCloser.Read(p)
+	bc.n += int64(n)
+	return n, err
+}
+
 func ProbeHTTP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var redirects int
 	var (
@@ -458,8 +471,10 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			}
 		}
 
+		byteCounter := &byteCounter{ReadCloser: resp.Body}
+
 		if success && (len(httpConfig.FailIfBodyMatchesRegexp) > 0 || len(httpConfig.FailIfBodyNotMatchesRegexp) > 0) {
-			success = matchRegularExpressions(resp.Body, httpConfig, logger)
+			success = matchRegularExpressions(byteCounter, httpConfig, logger)
 			if success {
 				probeFailedDueToRegex.Set(0)
 			} else {
@@ -468,13 +483,19 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		}
 
 		if resp != nil && !requestErrored {
-			respBodyBytes, err = io.Copy(ioutil.Discard, resp.Body)
+			_, err = io.Copy(ioutil.Discard, byteCounter)
 			if err != nil {
 				level.Info(logger).Log("msg", "Failed to read HTTP response body", "err", err)
 				success = false
 			}
 
-			resp.Body.Close()
+			respBodyBytes = byteCounter.n
+
+			if err := byteCounter.Close(); err != nil {
+				// We have already read everything we could from the server. The error here might be a
+				// TCP error. Log it in case it contains useful information as to what's the problem.
+				level.Info(logger).Log("msg", "Error while closing response from server", "error", err.Error())
+			}
 		}
 
 		// At this point body is fully read and we can write end time.
--- a/prober/http_test.go
+++ b/prober/http_test.go
@@ -14,6 +14,7 @@
 package prober
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
@@ -94,6 +95,100 @@ func TestValidHTTPVersion(t *testing.T) {
 		if result != test.ShouldSucceed {
 			t.Fatalf("Test %v had unexpected result: %s", i, body)
 		}
+	}
+}
+
+func TestContentLength(t *testing.T) {
+	type testdata struct {
+		contentLength          int
+		uncompressedBodyLength int
+		handler                http.HandlerFunc
+		expectFailure          bool
+	}
+
+	testmsg := []byte(strings.Repeat("hello world", 10))
+
+	notfoundMsg := []byte("not found")
+
+	testcases := map[string]testdata{
+		"identity": {
+			contentLength:          len(testmsg),
+			uncompressedBodyLength: len(testmsg),
+			handler: func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Add("Content-Encoding", "identity")
+				w.WriteHeader(http.StatusOK)
+				w.Write(testmsg)
+			},
+		},
+
+		"no content-encoding": {
+			contentLength:          len(testmsg),
+			uncompressedBodyLength: len(testmsg),
+			handler: func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+				w.Write(testmsg)
+			},
+		},
+
+		// Unknown Content-Encoding, we should let this pass thru.
+		"unknown content-encoding": {
+			contentLength:          len(testmsg),
+			uncompressedBodyLength: len(testmsg),
+			handler: func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Add("Content-Encoding", "xxx")
+				w.WriteHeader(http.StatusOK)
+				w.Write(bytes.Repeat([]byte{'x'}, len(testmsg)))
+			},
+		},
+
+		// 401 response, verify that the content-length is still computed correctly.
+		"401": {
+			expectFailure:          true,
+			contentLength:          len(notfoundMsg),
+			uncompressedBodyLength: len(notfoundMsg),
+			handler: func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusNotFound)
+				// Send something in the body to make sure that this get reported as the content length.
+				w.Write(notfoundMsg)
+			},
+		},
+	}
+
+	for name, tc := range testcases {
+		t.Run(name, func(t *testing.T) {
+			ts := httptest.NewServer(tc.handler)
+			defer ts.Close()
+
+			testCTX, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+
+			registry := prometheus.NewRegistry()
+			var logbuf bytes.Buffer
+			result := ProbeHTTP(testCTX,
+				ts.URL,
+				config.Module{
+					Timeout: time.Second,
+					HTTP:    config.HTTPProbe{IPProtocolFallback: true},
+				},
+				registry,
+				log.NewLogfmtLogger(&logbuf))
+			if !tc.expectFailure && !result {
+				t.Fatalf("probe failed unexpectedly: %s", logbuf.String())
+			} else if tc.expectFailure && result {
+				t.Fatalf("probe succeeded unexpectedly: %s", logbuf.String())
+			}
+
+			mfs, err := registry.Gather()
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			expectedResults := map[string]float64{
+				"probe_http_content_length":           float64(tc.contentLength),
+				"probe_http_uncompressed_body_length": float64(tc.uncompressedBodyLength),
+			}
+			checkRegistryResults(expectedResults, mfs, t)
+		})
 	}
 }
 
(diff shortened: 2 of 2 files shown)


prober/http.go
```diff
@@ -450,13 +474,37 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			}
 		}
 
+		if resp != nil && !requestErrored && strings.EqualFold(resp.Header.Get("Content-Encoding"), "gzip") {
```
brian-brazil, line 477: I don't like that we're special casing gzip compression, can we not let Go handle this for us?

prober/http.go
```diff
@@ -226,6 +228,28 @@ func (t *transport) GotFirstResponseByte() {
 	t.current.responseStart = time.Now()
 }
 
+// chainedReadCloser implementes an io.ReadCloser that refers to two io.ReadCloser instances, forwarding all Read
```
brian-brazil, line 231: A reader that counted the bytes that passed through it would seem a bit simpler than a unusual tee.

prober/http.go
mem: My bad, while chasing where the -1 was coming from I misread a diff that was doing something different.

A little bit of a braindump here...

In `github.com/prometheus/common@v0.10.0/config/http_config.go`  there's:

```
// NewRoundTripperFromConfig returns a new HTTP RoundTripper configured for the
// given config.HTTPClientConfig. The name is used as go-conntrack metric label.
func NewRoundTripperFromConfig(cfg HTTPClientConfig, name string, disableKeepAlives bool) (http.RoundTripper, error) {
        newRT := func(tlsConfig *tls.Config) (http.RoundTripper, error) {
                // The only timeout we care about is the configured scrape timeout.
                // It is applied on request. So we leave out any timings here.
                var rt http.RoundTripper = &http.Transport{
                        // ...
                        DisableCompression:  true,
                        // ...
                }

```

this function is called from `prober/http.go`

```
client, err := pconfig.NewClientFromConfig(httpClientConfig, "http_probe", true)
```

which is calling the above:

```
rt, err := NewRoundTripperFromConfig(cfg, name, disableKeepAlives)
```

That `DisableCompression` setting has been there for quite a while. My bad.

That said, the code eventually hits this:

```
        if cs.requestedGzip && res.Header.Get("Content-Encoding") == "gzip" {
                res.Header.Del("Content-Encoding")
                res.Header.Del("Content-Length")
                res.ContentLength = -1
                res.Body = &gzipReader{body: res.Body}
                res.Uncompressed = true
        }
```

requestedGzip is set, among other conditions, when DisableCompression is _false_, so that path is never taken for BBE.

This bit of code is relevant because the `res.Body` bit near the end is what causes the body to be automatically decompressed.

So with the current code BBE could be getting a compressed body (if the server provides one, even if one was not requested), `Content-Length` should be populated in the common case (even for compressed bodies), and `Content-Encoding` could be populated (with various values, but the only one that is handled by the existing code is `gzip`).

prober/http.go
mem: The intention behind that code is to be able to read from the gzip Reader, but at the same time be able to close both Closer's, if they exist. The original body needs to be closed to be able to reuse the TCP connection and free the file descriptor; the gzip reader needs to be closed in order to check if there are decompression errors.

I think your comment refers to the use of io.TeeReader in a different part of the change. I'll fix that, because it's true that I'm abusing the fact that `matchRegularExpressions` is using ioutil.ReadAll.

prober/http.go
mem: Thinking this thru, I must be hitting some servers where `Content-Length` is not reported, and that's why I'm seeing a -1:

```
$ curl --include --silent https://prometheus.io/
HTTP/2 200
date: Thu, 13 Aug 2020 22:21:23 GMT
content-type: text/html; charset=UTF-8
set-cookie: __cfduid=daaa32207a8b1cc2df0d447face9aa8a41597357283; expires=Sat, 12-Sep-20 22:21:23 GMT; path=/; domain=.prometheus.io; HttpOnly; SameSite=Lax
cache-control: public, max-age=0, must-revalidate
age: 2605
vary: Accept-Encoding
x-nf-request-id: d59078e9-d2ba-4178-98c3-2d05e5b46e3c-5951011
cf-cache-status: DYNAMIC
cf-request-id: 048b82ef9600001297350f5200000001
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
strict-transport-security: max-age=2592000; preload
x-content-type-options: nosniff
server: cloudflare
cf-ray: 5c25d42c2c4e1297-MIA
```

In this case `probe_http_content_length` is reported as -1 and `probe_http_uncompressed_body_length` is reported as whatever number of bytes were read.

brian-brazil: > So with the current code BBE could be getting a compressed body (if the server provides one, even if one was not requested),

That's a bug on the part of the server then, and not something we should worry about. Any failures due to this are true positives. I don't see why we need any special casing here, and of just one compression algorithm.

mem: I understand.

The reason why this is needed is because browsers will pay attention to the Content-Encoding header and decompress the body even if they didn't ask for compression, so there are servers out there relying on this behavior, and if we want to apply the regular expression validations, we need to decompress the body.

I have added support for decompressing the currently known values except for "compress" (mostly because the RFCs that talk about it don't specify the compression parameters that should be used, and those are necessary during decompression; also because this lost support from browsers around 20 years ago).

prober/http.go
mem: OK, removed.

prober/http.go
```diff
@@ -226,6 +227,45 @@ func (t *transport) GotFirstResponseByte() {
...
+		return err
+	}
+	return err2
+}
+
+// byteCounter implements an io.ReadCloser that keeps track of the total
+// number of bytes it has read.
+type byteCounter struct {
+	rc io.ReadCloser
+	n  *int64 // the location used to keep the running count
```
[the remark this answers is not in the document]
mem: I removed that.

Basically I was trying to keep the code using resp.Body and not having to type-cast to another type to be able to access the count. Obviously that makes the code less clear, so I changed it.

go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/prometheus/blackbox_exporter
 
 require (
+	github.com/andybalholm/brotli v1.0.0
```
mem, line 4: this is needed to support decompressing `br`

prober/http_test.go
```diff
@@ -97,6 +101,153 @@ func TestValidHTTPVersion(t *testing.T) {
...
+		}(),
+
+		"xxx": func() testdata {
+			msg := "hello world"
+			buf := []byte("xxx")
+			return testdata{
+				expectError:            true,
+				msg:                    msg,
+				contentLenght:          len(buf),
+				uncompressedBodyLength: len(buf), // compressed lenght is reported
```
mem, line 203: is this ok, or is it better to report -1?

mem:
I'm going to be away from the keyboard for ~ 1 week. Apologies for going silent.

prober/http.go
brian-brazil: Then that probe should fail, due to the servers not following the HTTP spec. We should never be adding code to pretend that non-compliant HTTP implementations are compliant.

prober/http.go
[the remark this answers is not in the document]
brian-brazil: The comment is still here.

brian-brazil:
> I'm going to be away from the keyboard for ~ 1 week. Apologies for going silent.

Don't worry about it, there's no rush here.

prober/http_test.go
brian-brazil: Hmm, hard to tell. We can't decode it so we can't know.

prober/http.go
mem: Going back to the RFC 7231 just to make sure what's conformant behavior and what isn't, I see:

> 5.3.4.  Accept-Encoding
> 
>    A server tests whether a content-coding for a given representation is
>    acceptable using these rules:
> 
>    1.  If no Accept-Encoding field is in the request, any content-coding
>        is considered acceptable by the user agent.
> 
>    2.  If the representation has no content-coding, then it is
>        acceptable by default unless specifically excluded by the
>        Accept-Encoding field stating either "identity;q=0" or "*;q=0"
>        without a more specific entry for "identity".
> 
>    3.  If the representation's content-coding is one of the
>        content-codings listed in the Accept-Encoding field, then it is
>        acceptable unless it is accompanied by a qvalue of 0.  (As
>        defined in Section 5.3.1, a qvalue of 0 means "not acceptable".)
> 
>    4.  If multiple content-codings are acceptable, then the acceptable
>        content-coding with the highest non-zero qvalue is preferred.

So not specifying Accept-Encoding is equivalent to `Accept-Encoding: *` per 5.3.4.1. I was in error when I said that servers rely on the browser's behavior.

Since we are not adding an `Accept-Encoding` header to the request, and the prometheus/common code is setting DisableCompression to true, what happens is:

* Accept-Encoding is not present, the server will treat that as `Accept-Encoding: *` sending back compressed content if it chooses to.

* We might get compressed content, but we don't decompress it, so we are not able to apply regular expression validations.

* If we were to set DisableCompression to false, the Go client would only handle gzip for us.

Sorry about the confusion I created, I should have started by going to the RFC to see why I was getting compressed content when I wasn't requesting it.

prober/http.go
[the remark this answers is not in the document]
mem: Really fixed now.

prober/http.go
brian-brazil: Hmm, it was better worded in RFC 2616: "If no Accept-Encoding field is present in a request, the server MAY assume that the client will accept any content coding. In this case, if "identity" is one of the available content-codings, then the server SHOULD use the "identity" content-coding..."

> prometheus/common code is setting DisableCompression to true

I'm not sure why that is set, it's not immediately clear from the history - but the right setting for the blackbox exporter.

If a server sends us arbitrary encodings in the hope that we understand them, I don't think we've any obligation to try and decode them - and the user can always specify headers explicitly. For an exporter like this I'd prefer not to be magic, this should be as raw HTTP as is practical.

mem: OK, cool, I'll remove the content-encoding handling from the PR.

mem: I have updated the PR and removed the `Content-Encoding` handling. If the server replies with compressed content, the regular expression validations, if any, will be applied to the compressed content and therefore they will not match. I will open another issue so we can discuss what to do with compressed responses.

I have left the code in the tests that verifies the behavior for compressed responses as a safeguard against the behavior of underlying libraries changing.

mem: The new issue is #684.

go.mod
brian-brazil: This is no longer needed.

prober/http_test.go
```diff
@@ -97,6 +101,158 @@ func TestValidHTTPVersion(t *testing.T) {
...
 
+func TestContentLength(t *testing.T) {
+	type testdata struct {
+		msg                    string
+		contentLenght          int
+		uncompressedBodyLength int
+		handler                http.HandlerFunc
+	}
+
+	// These test cases cover known encodings (brotli, deflate,
```
brian-brazil, line 112: These tests don't relate to this change.

go.mod
mem: I need this for the tests.

I thought about just reporting `Content-Encoding: br` and not providing a body that's actually compressed using that, but the intention behind leaving those tests there is to catch behavior changes in underlying libraries (like automatically decompressing content) so I need to provide something that is actually compressed like that.

brian-brazil: Let's not be pulling in dependencies just for tests of things we don't support.

mem: OK, done

prober/http_test.go
```diff
@@ -97,6 +100,140 @@ func TestValidHTTPVersion(t *testing.T) {
...
+	// identity) so changes in the underlying libraries don't cause
+	// undetected changes in behavior.
+	//
+	// They also cover unknown encodings (xxx) to verify that those
+	// are handled correctly.
+	testcases := map[string]testdata{
+		"deflate": func() testdata {
+			msg := strings.Repeat("hello world", 10)
+			var buf bytes.Buffer
+			// the only error path is an invalid compression level
```
brian-brazil, line 121: Comments should be full sentences with capital letters and full stops.

prober/http_test.go
```diff
@@ -97,6 +100,140 @@ func TestValidHTTPVersion(t *testing.T) {
...
+			fw, _ := flate.NewWriter(&buf, flate.DefaultCompression)
+			fw.Write([]byte(msg))
+			fw.Close()
+			return testdata{
+				msg:                    msg,
+				contentLenght:          buf.Len(), // content lenght is the length of the compressed buffer
```
brian-brazil, line 127: length

prober/http_test.go
mem: removed

prober/http_test.go
mem: Thanks.

This part of the code was removed because of the other comments.

prober/http_test.go
mem: Thanks. Always gets me.

prober/http_test.go
```diff
@@ -257,12 +353,15 @@ func TestFailIfBodyMatchesRegexp(t *testing.T) {
 		t.Fatal(err)
 	}
 	expectedResults := map[string]float64{
-		"probe_failed_due_to_regex": 1,
+		"probe_failed_due_to_regex":           1,
+		"probe_http_content_length":           float64(len(msg)), // issue #673, check that this is correctly populated with regex validations
```
brian-brazil, line 357: All these comments need formatting too.

mem: Thanks for spotting this.

While fixing this I noticed that the other two cases should be checking the results, too. I added that, and since the code was getting a little repetitive, I reorganized the code in the test a little. I simply moved the first block of code into an anonymous function and added t.Run to call that for each test case.

Solved.

brian-brazil:
Thanks!

PR #676, opened 2020-08-13
tcp prober: add support of raw bytes in query_response

cryptobioz:
This PR aims to make the blackbox-exporter support any protocol by allowing to pass raw bytes in `query_response`.

Our use case was to monitor PostgreSQL certificate expiration. The PostgreSQL SSL handshake is not very standard. First, we must send a bytes sequence to ask the server if it supports SSL. If so, it will respond with one byte which in the letter `S` in ASCII. Then, we can start a STARTTLS handshake.

I haven't be able to do these things with the current `send` and `except` functions. Thus, I added two *functions* `send_bytes` and `expect_bytes`. The idea was to write something simple and quite generic as we don't have to write specific code in the blackbox-exporter project for every custom procotol we want to support.
With this PR, the blackbox-exporter can support any protocol as long as the user describes it into a module.

Based on this PR, we wrote the following module for our PostgreSQL SSL handshake's use case:

```yaml
postgresql_ssl:
  prober: tcp
  timeout: 5s
  tcp:
    preferred_ip_protocol: ip4
    tls_config:
      insecure_skip_verify: true
    query_response:
      - send_bytes: "\u0000\u0000\u0000\u0008\u0004\u00d2\u0016\u002f"
      - expect_bytes: "S"
      - starttls: true
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -149,9 +149,11 @@ type HeaderMatch struct {
 }
 
 type QueryResponse struct {
-	Expect   string `yaml:"expect,omitempty"`
-	Send     string `yaml:"send,omitempty"`
-	StartTLS bool   `yaml:"starttls,omitempty"`
+	Expect      string `yaml:"expect,omitempty"`
+	ExpectBytes string `yaml:"expect_bytes,omitempty"`
+	Send        string `yaml:"send,omitempty"`
+	SendBytes   string `yaml:"send_bytes,omitempty"`
+	StartTLS    bool   `yaml:"starttls,omitempty"`
 }
 
 type TCPProbe struct {
--- a/prober/tcp.go
+++ b/prober/tcp.go
@@ -15,6 +15,7 @@ package prober
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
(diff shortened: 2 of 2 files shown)


brian-brazil:
Thanks for your PR. I don't think it makes sense to try and handle arbitrary binary protocols in this exporter.

As I've said for previous people trying to work with Postgres's bespoke TLS solution, a custom exporter is what I'd recommend.

cryptobioz:
Why supporting arbitrary binary protocols would make less sense than what already exist with `query_response`'s `send` and `except`?

brian-brazil:
Text based query response protocols are much simpler. For binary you basically need a full programming language.

raphink:
This PR doesn't intend to support the binary protocol as an abstraction though. It simply provides a way for users to use an advanced mode to send binary sequences of their own, so the complexity doesn't reside in the blackbox exporter, but on the user side, if any.

brian-brazil:
I'm not a fan of adding code to support one particular piece of software's way of doing things, that way lies madness. As I've already said, no matter how it is dressed up Postgres-specific code does not belong in this exporter.

mbornoz:
IMHO this is not a Postgres-specific code but a simple way to address binary protocol. What's the logical reason to limit this exporter to text based query response protocols? 

@brian-brazil you can admit that it's not very easy to define precisely the scope of an exporter called "blackbox_exporter", no?

brian-brazil:
I'm not interested in extended philosophical argument, I've already stated several times this feature does not belong in this exporter. If you wish to support non-standard protocols, then a separate exporter is recommended.

PR #678, opened 2020-08-18
prober/tls: adding metric to expose certificate fingerprint info

xinau:
this change adds a new metric `probe_ssl_fingerprint_info` to both tcp
and http probes. the metric always returns 1 similar to the tls version
metric and contains the leaf certificates sha256 fingerprint (hex) as a
label value.

this change allows users to validate in prometheus if a particular
certificate is being served.

### DIFF
--- a/prober/tcp.go
+++ b/prober/tcp.go
@@ -98,6 +98,13 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_ssl_last_chain_expiry_timestamp_seconds",
 		Help: "Returns last SSL chain expiry in unixtime",
 	})
+	probeSSLLastInformation := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "probe_ssl_last_chain_info",
+			Help: "Contains SSL leaf certificate information",
+		},
+		[]string{"fingerprint_sha256"},
+	)
 	probeTLSVersion := prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "probe_tls_version_info",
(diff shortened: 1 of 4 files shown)


prober/tcp.go
```diff
@@ -98,6 +98,13 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_ssl_last_chain_expiry_timestamp_seconds",
 		Help: "Returns last SSL chain expiry in unixtime",
 	})
+	probeSSLFingerprint := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "probe_ssl_fingerprint_info",
```
brian-brazil, line 103: This should mention that it's the leaf. Also you generally want to avoid an info metric per label name, so I'd not put fingerprint in the name.

xinau: @brian-brazil thanks for the hint on the first part, going to change it. Could you please elaborate a bit more what you mean by "info metric per label name" as I'm unable to follow you?

brian-brazil: Don't put the word fingerprint into the metric name, as that prevents us putting other leaf cert information in here too.

xinau: Okay got you. Going to change it to `probe_ssl_leaf_info` or do you have a better idea?

brian-brazil: probe_ssl_last_chain_info would be consistent with other metrics.

xinau: Is implemeted in [84d73fc](https://github.com/prometheus/blackbox_exporter/pull/678/commits/84d73fcf5ac1637711901d1a70036a410bae5972).

prober/tcp.go
```diff
@@ -98,6 +98,13 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_ssl_last_chain_expiry_timestamp_seconds",
 		Help: "Returns last SSL chain expiry in unixtime",
 	})
+	probeSSLLastInformation := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "probe_ssl_last_chain_info",
+			Help: "Contains SSL leaf certificate information",
+		},
+		[]string{"sha256"},
```
brian-brazil, line 106: sha256 is a bit vague

xinau: Going to change it to `fingerprint_sha256` or do you got something better in mind?

brian-brazil: That sounds good to me.

xinau: Changed it in [92dfa8f](https://github.com/prometheus/blackbox_exporter/pull/678/commits/92dfa8f4d4eaa114ab26241a5301f9a1cff7d6af)

brian-brazil:
Thanks!

xinau:
Thank you too.

PR #681, opened 2020-08-19
prober/tls: fix probe_ssl_last_chain_expiry_timestamp_seconds

ribbybibby:
Apologies if I've misunderstood the purpose of this metric but I believe that it should be reporting the earliest expiry date of the chain that expires last out of all the verified chains. Presently it's reporting the earliest expiry of the chain that expires first.

This wasn't identified by the test for this metric because the test is using an expired root certificate which is not included in the `VerifiedChains`, so there was only one chain for the method to iterate over and therefore `getLastChainExpiry` returned the same regardless of the logic.

### DIFF
--- a/prober/tcp_test.go
+++ b/prober/tcp_test.go
@@ -224,26 +224,26 @@ func TestTCPConnectionWithTLSAndVerifiedCertificateChain(t *testing.T) {
 		panic(fmt.Sprintf("Error creating rsa key: %s", err))
 	}
 
-	rootCertExpiry := time.Now().AddDate(0, 0, 2)
+	rootCertExpiry := time.Now().AddDate(0, 0, 3)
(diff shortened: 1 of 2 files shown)


brian-brazil:
Your change seems correct, @itkq did you miss this in your testing?

The unittest still doesn't seem right, we should have a test with different values for the two metrics.

ribbybibby:
@brian-brazil I've tweaked the test. I think this captures the situation this metric was put in place to address now: the server presents a root cert that expires before the server cert, despite the client holding another, valid, longer-lived root.

itkq:
This change seems right for me. I'm sorry, I should have noticed that `VerifiedChains` didn't have multiple chains in testing.

brian-brazil:
Thanks!

Issue #691, opened 2020-09-16
Regression in 0.17.0 when handling HTTP/2 responses


chancez:
### blackbox_exporter version: output of `blackbox_exporter -version`
v0.17.0

### What is the blackbox.yml module config.

```
modules:
  http_2xx:
    prober: http
    timeout: 5s
    http:
      valid_http_versions: ["HTTP/1.1", "HTTP/2"]
      no_follow_redirects: false
      preferred_ip_protocol: "ip4"
```

### Error

When https://github.com/prometheus/blackbox_exporter/pull/622 was added it also added HTTP2 support which caused the above config to fail due the issue mentioned in https://github.com/prometheus/blackbox_exporter/pull/633.

When we updated blackbox exporter from 0.16.0 to 0.17.0 a few of our HTTP probes started failing because our config used `HTTP/2` instead of `HTTP/2.0`. 

### Expected Behavior

Blackbox exporter should handle `HTTP/2` as `HTTP/2.0`, or the underlying HTTP2 library should be fixed to return the correct response.

### Suggestion

Simply treat `HTTP/2` as `HTTP/2.0` in https://github.com/prometheus/blackbox_exporter/blob/master/prober/http.go#L517.

brian-brazil:
Thanks for your report, this is not a bug nor regression in this code as this never worked. It would be incorrect to accept HTTP/2 as that is not what is on the wire (per RFC 7540 3.5), nor should we be 2nd guessing upstream, or having two ways to do the same thing as that'll ultimately cause confusion. The documentation bug was already fixed, so I don't see anything that needs to be changed here.

chancez:
> this is not a bug nor regression in this code as this never worked

Fair enough. I still think having users specify `HTTP/2.0` is strictly bad UX/surprising behavior if they want to match against `HTTP/2`. It makes no sense.

> It would be incorrect to accept HTTP/2 as that is not what is on the wire (per RFC 7540 3.5)

The fact is that cloudfront, google, cloudflare, etc returns `HTTP/2` so I think it's probably worth considering. I've yet to find an example (besides Go's) that returns `HTTP/2.0` in the version header after checking with a number of different sites on different CDNs.

SuperQ:
@chancez Interesting, maybe we should document both options, since there seems to be some RFC violations out there.

chancez:
[...]

PR #692, opened 2020-09-16
prober/http: Handle HTTP/2 as HTTP/2.0 for backwards compatibly

chancez:
Fixes https://github.com/prometheus/blackbox_exporter/issues/691


brian-brazil:
Thanks for your PR. This doesn't have anything to do with backwards compatibility, as this never worked - and per #691 doesn't make sense generally.

PR #694, opened 2020-09-25
Add probe_icmp_reply_ttl_total

luizluca:
IPv4 TTL (IPv6 hoplimit) identifies the number of routers used by
an arriving packet. It can be used to detect routing problems.

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

### DIFF
--- a/prober/icmp.go
+++ b/prober/icmp.go
@@ -65,14 +65,21 @@ func getICMPSequence() uint16 {
 
 func ProbeICMP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var (
-		socket      net.PacketConn
-		requestType icmp.Type
-		replyType   icmp.Type
+		requestType     icmp.Type
+		replyType       icmp.Type
+		icmpConn        *icmp.PacketConn
+		v4RawConn       *ipv4.RawConn
+		hopLimitFlagSet bool = true
 
 		durationGaugeVec = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "probe_icmp_duration_seconds",
 			Help: "Duration of icmp request by phase",
 		}, []string{"phase"})
+
+		hopLimitGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_icmp_reply_hop_limit",
+			Help: "Replied packet hop limit (TTL for ipv4)",
+		})
 	)
 
 	for _, lv := range []string{"resolve", "setup", "rtt"} {
@@ -81,7 +88,8 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 
 	registry.MustRegister(durationGaugeVec)
 
-	ip, lookupTime, err := chooseProtocol(ctx, module.ICMP.IPProtocol, module.ICMP.IPProtocolFallback, target, registry, logger)
+	dstIPAddr, lookupTime, err := chooseProtocol(ctx, module.ICMP.IPProtocol, module.ICMP.IPProtocolFallback, target, registry, logger)
+
 	if err != nil {
 		level.Warn(logger).Log("msg", "Error resolving address", "err", err)
 		return false
@@ -104,7 +112,7 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 	// Unprivileged sockets are supported on Darwin and Linux only.
 	tryUnprivileged := runtime.GOOS == "darwin" || runtime.GOOS == "linux"
 
-	if ip.IP.To4() == nil {
+	if dstIPAddr.IP.To4() == nil {
 		requestType = ipv6.ICMPTypeEchoRequest
 		replyType = ipv6.ICMPTypeEchoReply
 
@@ -112,7 +120,6 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 			srcIP = net.ParseIP("::")
 		}
 
-		var icmpConn *icmp.PacketConn
 		if tryUnprivileged {
 			// "udp" here means unprivileged -- not the protocol "udp".
 			icmpConn, err = icmp.ListenPacket("udp6", srcIP.String())
@@ -130,8 +137,12 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 				return
 			}
 		}
+		defer icmpConn.Close()
 
-		socket = icmpConn
+		if err := icmpConn.IPv6PacketConn().SetControlMessage(ipv6.FlagHopLimit, true); err != nil {
+			level.Debug(logger).Log("msg", "Failed to set Control Message for retrieving Hop Limit", "err", err)
+			hopLimitFlagSet = false
+		}
 	} else {
 		requestType = ipv4.ICMPTypeEcho
 		replyType = ipv4.ICMPTypeEchoReply
@@ -143,21 +154,25 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 		if module.ICMP.DontFragment {
 			// If the user has set the don't fragment option we cannot use unprivileged
 			// sockets as it is not possible to set IP header level options.
-			icmpConn, err := net.ListenPacket("ip4:icmp", srcIP.String())
+			netConn, err := net.ListenPacket("ip4:icmp", srcIP.String())
 			if err != nil {
 				level.Error(logger).Log("msg", "Error listening to socket", "err", err)
 				return
 			}
+			defer netConn.Close()
 
-			rc, err := ipv4.NewRawConn(icmpConn)
+			v4RawConn, err = ipv4.NewRawConn(netConn)
 			if err != nil {
 				level.Error(logger).Log("msg", "Error creating raw connection", "err", err)
 				return
 			}
-			socket = &v4Conn{c: rc, df: true}
-		} else {
-			var icmpConn *icmp.PacketConn
+			defer v4RawConn.Close()
 
+			if err := v4RawConn.SetControlMessage(ipv4.FlagTTL, true); err != nil {
+				level.Debug(logger).Log("msg", "Failed to set Control Message for retrieving TTL", "err", err)
+				hopLimitFlagSet = false
+			}
+		} else {
 			if tryUnprivileged {
 				icmpConn, err = icmp.ListenPacket("udp4", srcIP.String())
 				if err != nil {
@@ -174,16 +189,18 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 					return
 				}
 			}
+			defer icmpConn.Close()
 
-			socket = icmpConn
+			if err := icmpConn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true); err != nil {
+				level.Debug(logger).Log("msg", "Failed to set Control Message for retrieving TTL", "err", err)
+				hopLimitFlagSet = false
+			}
 		}
 	}
 
-	defer socket.Close()
-
-	var dst net.Addr = ip
+	var dst net.Addr = dstIPAddr
 	if !privileged {
-		dst = &net.UDPAddr{IP: ip.IP, Zone: ip.Zone}
+		dst = &net.UDPAddr{IP: dstIPAddr.IP, Zone: dstIPAddr.Zone}
 	}
 
 	var data []byte
@@ -215,7 +232,26 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 	durationGaugeVec.WithLabelValues("setup").Add(time.Since(setupStart).Seconds())
 	level.Info(logger).Log("msg", "Writing out packet")
 	rttStart := time.Now()
-	if _, err = socket.WriteTo(wb, dst); err != nil {
+
+	if icmpConn != nil {
+		_, err = icmpConn.WriteTo(wb, dst)
+	} else {
+		// Only for IPv4 raw. Needed for setting DontFragment flag.
+		header := &ipv4.Header{
+			Version:  ipv4.Version,
+			Len:      ipv4.HeaderLen,
+			Protocol: 1,
+			TotalLen: ipv4.HeaderLen + len(wb),
+			TTL:      64,
+			Dst:      dstIPAddr.IP,
+			Src:      srcIP,
+		}
+
+		header.Flags |= ipv4.DontFragment
+
+		err = v4RawConn.WriteTo(header, wb, nil)
+	}
+	if err != nil {
 		level.Warn(logger).Log("msg", "Error writing to socket", "err", err)
 		return
 	}
@@ -243,13 +279,52 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 
 	rb := make([]byte, 65536)
 	deadline, _ := ctx.Deadline()
-	if err := socket.SetReadDeadline(deadline); err != nil {
+	if icmpConn != nil {
+		err = icmpConn.SetReadDeadline(deadline)
+	} else {
+		err = v4RawConn.SetReadDeadline(deadline)
+	}
+	if err != nil {
 		level.Error(logger).Log("msg", "Error setting socket deadline", "err", err)
 		return
 	}
 	level.Info(logger).Log("msg", "Waiting for reply packets")
 	for {
-		n, peer, err := socket.ReadFrom(rb)
+		var n int
+		var peer net.Addr
+		var err error
+		var hopLimit float64 = -1
+
+		if dstIPAddr.IP.To4() == nil {
+			var cm *ipv6.ControlMessage
+			n, cm, peer, err = icmpConn.IPv6PacketConn().ReadFrom(rb)
+			// HopLimit == 0 is valid for IPv6, although go initialize it as 0.
+			if cm != nil && hopLimitFlagSet {
+				hopLimit = float64(cm.HopLimit)
+			} else {
+				level.Debug(logger).Log("msg", "Cannot get Hop Limit from the received packet. 'probe_icmp_reply_hop_limit' will be missing.")
+			}
+		} else {
+			var cm *ipv4.ControlMessage
+			if icmpConn != nil {
+				n, cm, peer, err = icmpConn.IPv4PacketConn().ReadFrom(rb)
+			} else {
+				var h *ipv4.Header
+				var p []byte
+				h, p, cm, err = v4RawConn.ReadFrom(rb)
+				if err == nil {
+					copy(rb, p)
+					n = len(p)
+					peer = &net.IPAddr{IP: h.Src}
+				}
+			}
+			if cm != nil && hopLimitFlagSet {
+				// Not really Hop Limit, but it is in practice.
+				hopLimit = float64(cm.TTL)
+			} else {
+				level.Debug(logger).Log("msg", "Cannot get TTL from the received packet. 'probe_icmp_reply_hop_limit' will be missing.")
+			}
+		}
 		if err != nil {
 			if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
 				level.Warn(logger).Log("msg", "Timeout reading from socket", "err", err)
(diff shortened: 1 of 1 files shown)


prober/icmp.go
```diff
@@ -80,6 +80,11 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 	}
 
 	registry.MustRegister(durationGaugeVec)
+	ttlGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_icmp_reply_ttl_total",
```
brian-brazil, line 84: Gauges shouldn't end in total, that's for counters.

prober/icmp.go
```diff
@@ -80,6 +80,11 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
+		Help: "Number of router hops returning from the target",
```
brian-brazil, line 85: No it's not, it's the TTL of the received packet - these are very different things.

prober/icmp.go
```diff
@@ -131,6 +137,7 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 			}
 		}
 
+		icmpConn.IPv6PacketConn().SetControlMessage(ipv6.FlagHopLimit, true)
```
brian-brazil, line 140: Are we changing the outbound packet here? The docs don't indicate what this actually does.

prober/icmp.go
```diff
@@ -287,18 +312,23 @@ type v4Conn struct {
...
 	copy(b, p)
 	n := len(b)
 	if len(p) < len(b) {
 		n = len(p)
 	}
-	return n, &net.IPAddr{IP: h.Src}, nil
+	return n, cm, &net.IPAddr{IP: h.Src}, nil
+}
+
+func (c *v4Conn) ReadFrom(b []byte) (int, net.Addr, error) {
```
brian-brazil, line 329: We don't need a trivial wrapper like this, just change the existing function.

prober/icmp.go
```diff
@@ -249,7 +256,24 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 	}
 	level.Info(logger).Log("msg", "Waiting for reply packets")
 	for {
-		n, peer, err := socket.ReadFrom(rb)
+		var n int
+		var peer net.Addr
+		var err error
+		var ttl int
+
+		if ip.IP.To4() == nil {
```
brian-brazil, line 264: The main point of having v4Conn is to make v4 and v6 sockets look the same, this code here is completely undoing that polymorphism so we end up with the costs of that without any of the benefits.

brian-brazil (commented):
All codepaths here will need manual testing on both Linux and Windows.

prober/icmp.go
luizluca: OK. I renamed it to probe_icmp_reply_hop_limit as even IPv6 header renamed ttl to hop_limit

prober/icmp.go
luizluca: Now 'Replied packet hop limit (TTL for ipv4)'

prober/icmp.go
luizluca: I added a new version without the v4Conn wrapper class. I would need to forcely morph a ipv4.RawConn into a ipv4.PacketConn. It might be a good reason for go not putting a common public interface for RawConn and PacketConn.

Without v4Conn, it looks cleaner with less LoC.

prober/icmp.go
luizluca: SetControlMessage asks for extra info when a datagram is received (inside controlMessage). It might be not fully documented inside go because it is not something special for it. It is a UNIX thing.

> ...
> The  field  msg_control, which has length msg_controllen, points to **a buffer for other protocol control-related messages or miscellaneous ancillary data**.  When recvmsg() is called, msg_controllen should contain the length of the available buffer in msg_control; upon return from a successful call it will contain the length of the control message sequence.
> ...
> recvmsg(2) manual

> ...
> IP_RECVTTL (since Linux 2.2)
> When this flag is set, pass a IP_TTL control message with the time-to-live field of the received packet as a 32 bit integer.  Not supported for SOCK_STREAM sockets.
> ...
> ip(7) manual

> ...
>IPV6_RTHDR, IPV6_AUTHHDR, IPV6_DSTOPTS, IPV6_HOPOPTS, IPV6_FLOWINFO, **IPV6_HOPLIMIT**
>Set delivery of control messages for incoming datagrams containing extension headers from the received packet.  IPV6_RTHDR delivers the routing header, IPV6_AUTHHDR delivers the authentication header, IPV6_DSTOPTS delivers the destination options, IPV6_HOPOPTS delivers the hop options, IPV6_FLOWINFO delivers an integer containing the flow ID, **IPV6_HOPLIMIT delivers an integer containing the hop count of the packet**. The  control messages have the  same type as the socket option. All these header options can also be set for outgoing packets by putting the appropriate control message into the control buffer of sendmsg(2). Allowed only for SOCK_DGRAM or SOCK_RAW sockets.  Argument is a pointer to a boolean value.
> ...
> ipv6(7) manual

AFAIK, at least for those flags, it should not affect outgoing datagrams, even only checking go docs:

> TTL     int    // time-to-live, **receiving only**
https://github.com/golang/net/blob/master/ipv4/control.go#L35
> FlagHopLimit                              // pass the hop limit on the **received packet**
https://github.com/golang/net/blob/master/ipv6/control.go#L32

It might not work with Windows machines. I don't have a windows machine to test Prometheus but it will be 0 if missing (hoplimit 0 will never naturally occur as it will get dropped before reaching us).

How to deal with missing data in Prometheus? Null?? Hide the property? Use zero? -1?

prober/icmp.go
luizluca: I could not change the signature as it would violate the interface.

prober/icmp.go
brian-brazil: The question is more will setting this break icmp on Windows. This will need to be tested manually.

prober/icmp.go
```diff
@@ -143,21 +153,23 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
+			rawConn, err = ipv4.NewRawConn(netConn)
 			if err != nil {
 				level.Error(logger).Log("msg", "Error creating raw connection", "err", err)
 				return
 			}
-			socket = &v4Conn{c: rc, df: true}
+			if err := rawConn.SetControlMessage(ipv4.FlagTTL, true); err != nil {
+				level.Debug(logger).Log("msg", "Failed to set Control Message for retrieving TTL", "err", err)
+			}
+			defer rawConn.Close()
```
brian-brazil, line 171: Shouldn't any defer be right after it is instantiated? That reduces the risk of it being missed.

prober/icmp.go
```diff
@@ -174,16 +186,16 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 					return
 				}
 			}
-
-			socket = icmpConn
+			if err := icmpConn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true); err != nil {
+				level.Debug(logger).Log("msg", "Failed to set Control Message for retrieving TTL", "err", err)
+			}
+			defer icmpConn.Close()
```
brian-brazil, line 192: Ditto

prober/icmp.go
```diff
@@ -215,7 +227,25 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 	durationGaugeVec.WithLabelValues("setup").Add(time.Since(setupStart).Seconds())
 	level.Info(logger).Log("msg", "Writing out packet")
 	rttStart := time.Now()
-	if _, err = socket.WriteTo(wb, dst); err != nil {
+
+	if icmpConn != nil {
+		_, err = icmpConn.WriteTo(wb, dst)
+	} else {
+		header := &ipv4.Header{
```
brian-brazil, line 234: A comment that this is only for DF v4 would be useful, it takes a bit to figure out from the code.

brian-brazil (commented):
That looks better, this will all need manual testing as we've been caught out before.

prober/icmp.go
```diff
@@ -65,23 +65,31 @@ func getICMPSequence() uint16 {
 
 func ProbeICMP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var (
-		socket      net.PacketConn
 		requestType icmp.Type
 		replyType   icmp.Type
+		icmpConn    *icmp.PacketConn
+		rawConn     *ipv4.RawConn
```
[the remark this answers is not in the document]
luizluca: Done!

prober/icmp.go
luizluca: There is just one place it could fail with an open connection before ipv4.NewRawConn(). I added the extra Close() as I guess multiple Close() will not break anything. Anyway, I moved all defer to before SetControlMessage.

prober/icmp.go
luizluca: // Only for IPv4 raw (needed for setting DontFragment flag)

prober/icmp.go
luizluca: For Windows it will (currently) return a NotImplemented error (https://github.com/golang/net/blob/5d4f7005572804eaf7f5ecdd2473a62557f733ba/ipv4/control_windows.go#L11), although there is a "// TODO(mikio): implement this". It might need to map UNIX specific code to Windows.

When NotImplemented is returned, my code will write a msg to log. Nothing else will change.

The cm was already being "returned" but ignored (the "_" here https://github.com/prometheus/blackbox_exporter/pull/694/files#diff-57e28fd17986190d6a1ba451a5131874L291

The cm is initialized inside packetHandler.ReadFrom() if message.NN is positive (The NN field indicates the number of bytes read or written from/to OOB). For windows, I'm not sure if OOB is empty or not. So, I need to consider it might be nil.
I added an if before using it.

Now, it might not break anything even in Windows. I just don't have a Windows to test.

The question that remains is about missing data. According to [this](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics), I should give a metric even if missing. I could use NaN or 0. If I do nothing, it will be 0. ipv4 TTL==0 is not valid even for localhost while IPv6 HopLimit could be 0 between hosts in same network. As I'm getting it from an ICMP echo-reply, HopLimit == 0 might be valid when the answer reaches the probe exactly one hop from being dropped. It might be too rare to bother. Someone might manually set echo-request HopLimit/TTL to count hops but not echo-response HopLimit/TTL (which is what I'm monitoring). Other metrics, like probe_icmp_duration_seconds, give me 0 when they fail, even when 0 is a "valid duration". NaN seems to be better to avoid "missing/ipv6 last hop" dubious meaning. However, it would make sense if all other metrics also had the same semantic. I guess changing a metric semantic (0 => NaN) is dangerous as it might break someone consuming it.

luizluca:
New push:

- Renamed  rawConn -> v4RawConn, ttlGauge -> hopLimitGauge
- check if ControlMessage is null (possibly in Windows)

brian-brazil:
That push didn't seem to change anything, can you try again?

prober/icmp.go
```diff
@@ -215,7 +230,26 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
 	durationGaugeVec.WithLabelValues("setup").Add(time.Since(setupStart).Seconds())
 	level.Info(logger).Log("msg", "Writing out packet")
 	rttStart := time.Now()
-	if _, err = socket.WriteTo(wb, dst); err != nil {
+
+	if icmpConn != nil {
+		_, err = icmpConn.WriteTo(wb, dst)
+	} else {
+		// Only for IPv4 raw (needed for setting DontFragment flag)
```
brian-brazil, line 237: full stop, and remove the parens.

prober/icmp.go
```diff
@@ -243,13 +277,47 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
-		n, peer, err := socket.ReadFrom(rb)
+		var n int
+		var peer net.Addr
+		var err error
+		var hopLimit float64
+
+		if dstIPAddr.IP.To4() == nil {
+			var cm *ipv6.ControlMessage
+			n, cm, peer, err = icmpConn.IPv6PacketConn().ReadFrom(rb)
+			if cm != nil {
```
brian-brazil, line 299: How can this be nil?

Don't guess what happens on Windows, test it out.

prober/icmp.go
luizluca: OK

prober/icmp.go
luizluca: When there is no OOB data, m.NN will be zero here:
https://github.com/golang/net/blob/4acb6c075d100d6ce8757083ec7382423f2a7126/ipv6/payload_cmsg.go#L41

If m.NN was never zero, there should not exist that "if m.NN > 0". I'm not an expert here and maybe ICMP might always returns something as OOB. But I cannot be sure.

Now I discovered that windows crossbuilding with go is trivial. I managed to run it in windows. As expected, SetControlMessage returns err="not implemented on windows/amd64", generating a debug message.

In Windows, cm==nil for all my tests. I'll add a debug message when it cm is null.

prober/icmp.go
brian-brazil: The right way to do it would be not to expose it when it's missing, how the blackbox exporter generally handles this predates staleness. NaN shouldn't be used here.

prober/icmp.go
```diff
@@ -143,21 +153,25 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
-			icmpConn, err := net.ListenPacket("ip4:icmp", srcIP.String())
+			netConn, err := net.ListenPacket("ip4:icmp", srcIP.String())
 			if err != nil {
 				level.Error(logger).Log("msg", "Error listening to socket", "err", err)
 				return
 			}
+			defer netConn.Close()
 
-			rc, err := ipv4.NewRawConn(icmpConn)
+			level.Info(logger).Log("msg", "Creating RawConn")
```
brian-brazil, line 163: I don't think we need this.

brian-brazil (commented):
I take it you've manually tested the various code paths?

prober/icmp.go
luizluca: Removed.

prober/icmp.go
luizluca: OK, I'll register the gauge only when setControlMessage works, and controlMessage is not nil.

luizluca:
New push:
* conditionally register the hopLimit prop
* remove the RawConn msg

prober/icmp.go
```diff
@@ -243,13 +279,51 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
 		level.Error(logger).Log("msg", "Error setting socket deadline", "err", err)
 		return
 	}
 	level.Info(logger).Log("msg", "Waiting for reply packets")
 	for {
-		n, peer, err := socket.ReadFrom(rb)
+		var n int
+		var peer net.Addr
+		var err error
+		var hopLimit float64 = -1
```
brian-brazil, line 296: Could we use just this rather than also needing the bool?

brian-brazil (commented):
> I forced 'dontFragment' flag modifying the code as I could not pass it using URL variables.

It comes from the config file.

> Do I need to test anything else?

The unprivileged sockets should also be tested.

> BTW, the timeout for non-existing targets is quite long 

It used to be much shorter, however that made it difficult to test slow targets from the browser.

prober/icmp.go
luizluca: OK, I changed hopLimit to a global var. '-1' means "no hopLimit". I need to ignore cm info when cm is nil or hopLimit is -1 (SetControlMessage failed). cm could theoretically have a value whenever there is any oob data not necessary about hopLimit.

luizluca: > The unprivileged sockets should also be tested.

All Linux tests where already using unprivileged sockets. I changed the code to force using privileged sockets. And it works as expected. Windows always uses privilege sockets.

luizluca:
New Push:
* removed hasHopLimit. Now it uses hopLimit==-1 as hasHopLimit==false
* hopLimit initialized as 0 to emphasizes its initial state (although it should be 0 by default)

prober/icmp.go
```diff
@@ -65,14 +65,21 @@ func getICMPSequence() uint16 {
 
 func ProbeICMP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
 	var (
-		socket      net.PacketConn
 		requestType icmp.Type
 		replyType   icmp.Type
+		icmpConn    *icmp.PacketConn
+		v4RawConn   *ipv4.RawConn
+		hopLimit    float64 = 0
```
brian-brazil, line 72: The = 0 is redundant, however it'd be cleaner to set it to -1 here.

prober/icmp.go
```diff
@@ -243,13 +279,50 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
 	for {
-		n, peer, err := socket.ReadFrom(rb)
+		var n int
+		var peer net.Addr
+		var err error
+
+		if dstIPAddr.IP.To4() == nil {
+			var cm *ipv6.ControlMessage
+			n, cm, peer, err = icmpConn.IPv6PacketConn().ReadFrom(rb)
+			if cm != nil && hopLimit >= 0 {
```
brian-brazil, line 300: Why the check on hopLimit here?

prober/icmp.go
luizluca: It can be either -1 or 0. I just need to set it to -1 if SetControlMessage fails or 0 if it works. As I don't use the "else" when I check SetControlMessage , I opted to use 0 by default and set it to -1 inside the existing block.

What would be the better option?

prober/icmp.go
luizluca: hopLimit==-1 would tell me if SetControlMessage failed. In this cade, TTL/HopLimit info might exist (when there is any OOB data) but it is trash. It would be much easier if hopLimit == 0 wasn't a valid state.

prober/icmp.go
brian-brazil: A separate bool would be cleaner I think, rather than overloading this variable.

luizluca: It was the first approach. Updated.

prober/icmp.go
brian-brazil, line 296: This is still messy, as there's now two hasHopLimit indicators doing different things. It'd be better to name the bool what is actually, namely whether the CM was set.

luizluca: The real issue is that go does not tell me if hopLimit is missing or zero. I opened an issue golang/go#41820.

I renamed the bool to ipv6HopLimitFlagSet and now I use it only for IPv6. If golang fixes the issue, it can be removed in the future to adapt to whatever go uses to inform hopLimit is missing.

Another solution is to assume that hopLimit will never be 0 in a echo-reply. I assumed that hop limit could be zero when it is using the last permitted hop. However, I was wrong as a packet with hopLimit==1 will not get forwarded. The IPv6 echo reply will only be 0 if the target OS is explicitly setting the answer to be 0 (maybe a way to avoid its traffic to be routed?). As it is a very very special case, we might ignore it.

What do you think?

luizluca:
- Changed the debug message when TTL/HopLimit is missing
- Using ipv6HopLimitFlagSet instead of hasHopLimit

prober/icmp.go
brian-brazil: Probably best not to make assumptions about the TTL in a tool like this, as if that weird situation does happen you wouldn't want to lead users down the wrong path. It's probably best to use the flag for both protocols.

prober/icmp.go
```diff
@@ -243,13 +279,53 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
+				h, p, cm, err = v4RawConn.ReadFrom(rb)
+				if err == nil {
+					copy(rb, p)
+					n = len(p)
+					peer = &net.IPAddr{IP: h.Src}
+				}
+			}
+			// TTL == 0 is not valid for IPv4
+			if cm != nil && cm.TTL > 0 {
+				/* Not really hopLimit, but it is in practice */
```
brian-brazil, line 323: Be consistent in comment style, and all comments should end with a full stop.

prober/icmp.go
luizluca: OK. Let's play safe. I reintroduced the bool check for both.

prober/icmp.go
luizluca: ok. I also switched from /*x*/ to //x

brian-brazil:
Thanks!

luizluca:
> Thanks!

Thank YOU for Prometheus!

PR #697, opened 2020-10-05
Add option to hide configuration

victorvarias:
I've added the option to hide the configuration as in some cases can be required.


brian-brazil:
Thanks for your PR. The config file doesn't contain any secrets, so I see no need to make it harder to debug things by hiding it. If you really need to hide it for some reason, I'd suggest using a reverse proxy.

victorvarias:
Hi Brian, thanks for your comments.
I'm going to review my use case and find a proper solution.

PR #703, opened 2020-10-27
CrLf option for newline

fischerman:
Most internet protocols require `CRLF` for a newline. For example, for SMTP see https://tools.ietf.org/html/rfc2821#section-4.1.1

The blackbox exporter is only sending `LF`, which is accepted by some servers (e.g. Postfix) but not by others (e.g. Exchange). This PR adds an option to send `CRLF`. I've set it to false by default for backwards compatibility (although it should probably be enabled in most cases).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,7 @@ var (
 	// DefaultTCPProbe set default value for TCPProbe
 	DefaultTCPProbe = TCPProbe{
 		IPProtocolFallback: true,
+		CrLf:               false,
 	}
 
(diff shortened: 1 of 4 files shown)


brian-brazil:
Hmm, couldn't you put a `\r` on the end of the string in your config?

fischerman:
To my surprise that works. Should we at least change the SMTP examples to make them compliant to spec?

brian-brazil:
That sounds like a good idea.

fischerman:
Closing in favor of #706

PR #707, opened 2020-11-03
Do not connect via IP when behind proxy

thiagoarrais:
# Problem

My proxy only allows connections to some domains. It therefore needs to see the domain name of the connection in order to let it through. This is specially true for SSL connections, since the `Host` header gets sent inside the encrypted tunnel.

# Proposal

This PR avoids connecting via IP when using proxies. It modifies the HTTP prober to send the domain name in the clear in case any proxy URL is set.

This is an initial approach, of course. I will be happy to iterate this out with the prometheus/blackbox team until we get to an usable version.

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -393,14 +393,16 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 
 	// Replace the host field in the URL with the IP we resolved.
 	origHost := targetURL.Host
-	if targetPort == "" {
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. This has previously been discussed in issues such as #264 and the short version is that DNS resolution of the target must stay within the blackbox exporter as otherwise you're not testing the target - you're testing the proxy. However proposals how how to test HTTP proxies are welcome.

PR #714, opened 2020-11-27
change Histogram bucket

chensunny:
* change Histogram bucket  to improve p99 status

### DIFF
--- a/main.go
+++ b/main.go
@@ -92,7 +92,7 @@ func probeHandler(w http.ResponseWriter, r *http.Request, c *config.Config, logg
 		Name: "probe_success",
 		Help: "Displays whether or not the probe was a success",
 	})
-	probeDurationGauge := prometheus.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. This doesn't make sense, as this would be on the /metrics which is metrics about the blackbox exporter itself. This would those not be useful as it'd be on the wrong target, plus could pose cardinality issues as there's no bound there. proce_duration_seconds should already provide everything you need, without having to use any buckets (or figure out what buckets would apply to every probe for every user in existence).

SuperQ:
To be specific, you can use query like [`quantile_over_time(0.99, probe_duration_seconds[5m])`](https://prometheus.demo.do.prometheus.io/graph?g0.range_input=1d&g0.step_input=300&g0.expr=quantile_over_time(0.99%2C%20probe_duration_seconds%5B5m%5D)&g0.tab=0&g1.range_input=1d&g1.step_input=15&g1.expr=probe_duration_seconds&g1.tab=0) to get the same results as this would provide.

brian-brazil:
Closing per above.

PR #716, opened 2020-11-27
Resolve #715 Metric for unknown probe

### LINKED ISSUE #715: [Feature Request] > Metric for unknown module
Currently this is hidden in the response
https://github.com/prometheus/blackbox_exporter/blob/9fd70341277528877d55bfeffd46593646bca347/main.go#L77

Would be great to add two info points to this
- Log Error message
- Metric

Reason we ask for a metric is so that we(SRE) team can alert on misconfiguration of the module by teams. Thanks

I might get time to do this myself...seems easy enough 🤞

gavk34:
Resolve #715

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ _test
 _cgo_defun.c
 _cgo_gotypes.go
 _cgo_export.*
+.idea
 
 _testmain.go
 
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ _test
 _cgo_defun.c
 _cgo_gotypes.go
 _cgo_export.*
-.idea
 
 _testmain.go
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ will return metrics for a HTTP probe against google.com. The `probe_success`
 metric indicates if the probe succeeded. Adding a `debug=true` parameter
 will return debug information for that probe.
 
-Note the `probe_module_unknown` metric indicates if the module used is unknown.
+Note the `probe_unknown_module` metric indicates if the module used is unknown.
 
 ## Building the software
 
--- a/README.md
+++ b/README.md
@@ -31,8 +31,6 @@ will return metrics for a HTTP probe against google.com. The `probe_success`
 metric indicates if the probe succeeded. Adding a `debug=true` parameter
 will return debug information for that probe.
 
-Note the `probe_unknown_module` metric indicates if the module used is unknown.
-
 ## Building the software
 
 ### Local Build
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ will return metrics for a HTTP probe against google.com. The `probe_success`
 metric indicates if the probe succeeded. Adding a `debug=true` parameter
 will return debug information for that probe.
 
+Note the `probe_unknown_module` metric indicates if the module used is unknown.
+
 ## Building the software
 
 ### Local Build
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ will return metrics for a HTTP probe against google.com. The `probe_success`
 metric indicates if the probe succeeded. Adding a `debug=true` parameter
 will return debug information for that probe.
 
-Note the `probe_unknown_module` metric indicates if the module used is unknown.
+Note the `probe_module_unknown` metric indicates if the module used is unknown.
 
 ## Building the software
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -16,6 +16,7 @@ _test
 _cgo_defun.c
 _cgo_gotypes.go
 _cgo_export.*
+.idea
```
brian-brazil, line 19: This belongs in your global gitignore.

README.md
```diff
@@ -31,6 +31,8 @@ will return metrics for a HTTP probe against google.com. The `probe_success`
 metric indicates if the probe succeeded. Adding a `debug=true` parameter
 will return debug information for that probe.
 
+Note the `probe_module_unknown` metric indicates if the module used is unknown.
```
brian-brazil, line 34: This doesn't belong here, please remove it.

main.go
```diff
@@ -68,13 +68,22 @@ var (
 )
 
 func probeHandler(w http.ResponseWriter, r *http.Request, c *config.Config, logger log.Logger, rh *resultHistory) {
+	probeModuleUnknownGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_module_unknown",
+		Help: "Displays whether or not probe module is unknown",
+	})
+	registry := prometheus.NewRegistry()
```
brian-brazil, line 75: This isn't right. This could be a counter using the default registry, and defined at file level.

main.go
```diff
@@ -68,13 +68,22 @@ var (
...
+	registry.MustRegister(probeModuleUnknownGauge)
+
 	moduleName := r.URL.Query().Get("module")
 	if moduleName == "" {
 		moduleName = "http_2xx"
 	}
 	module, ok := c.Modules[moduleName]
 	if !ok {
 		http.Error(w, fmt.Sprintf("Unknown module %q", moduleName), http.StatusBadRequest)
+		level.Error(logger).Log("msg", "Error ", "err", fmt.Sprintf("unknown module %s", moduleName))
```
brian-brazil, line 85: No need to spam the logs when someone has incorrect usage.

gavk34: Remove or switch to `debug`?

.gitignore
gavk34: Removed 👍

README.md
gavk34: Done 👍

main.go
gavk34: Think this is done now, please review

main.go
gavk34: I have switched to `debug`

main.go
```diff
@@ -65,6 +65,11 @@ var (
 		"icmp": prober.ProbeICMP,
 		"dns":  prober.ProbeDNS,
 	}
+
+	probeModuleUnknownCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "probe_module_unknown",
```
brian-brazil, line 70: Counters should end with _total.

This is not a metric from a probe, it is a metric about the blackbox exporter. Thus it should be blackbox_

main.go
```diff
@@ -75,6 +80,8 @@ func probeHandler(w http.ResponseWriter, r *http.Request, c *config.Config, logg
 	module, ok := c.Modules[moduleName]
 	if !ok {
 		http.Error(w, fmt.Sprintf("Unknown module %q", moduleName), http.StatusBadRequest)
+		level.Debug(logger).Log("msg", "Error ", "err", fmt.Sprintf("unknown module %s", moduleName))
```
brian-brazil, line 83: The message should be a fixed string about what's going on, and then the module a pair on its own.

main.go
```diff
@@ -80,7 +80,7 @@ func probeHandler(w http.ResponseWriter, r *http.Request, c *config.Config, logg
 	module, ok := c.Modules[moduleName]
 	if !ok {
 		http.Error(w, fmt.Sprintf("Unknown module %q", moduleName), http.StatusBadRequest)
-		level.Debug(logger).Log("msg", "Error ", "err", "unknown module name", "module", moduleName)
+		level.Debug(logger).Log("msg", "Error finding module", "module", moduleName)
```
brian-brazil, line 83: It's not about finding, keep it consistent with the error we send back over HTTP - so Unknown module.

gavk34: Done, just in time for the Toy Show 🇮🇪 😄

brian-brazil:
Thanks!

PR #720, opened 2020-12-04
Changes in .travis.yml

nageshlop:
This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.
The build and test results are available at the below location.
https://travis-ci.com/github/nageshlop/blackbox_exporter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
-
+arch:
+  - amd64
+  - ppc64le
+  
 language: go
 
 # Whenever the Go version is updated here, .circleci/config.yml and .promu.yml


SuperQ:
Unfortunately, we're moving our CI pipelines away from Travis, as the service quality has degraded to the point where it's not useful. This is one of the last instances used across all of the Prometheus projects.

nageshlop:
Hi @SuperQ 
Thanks for the update. Please refer to below blog which details further on this change.
https://blog.travis-ci.com/oss-announcement

SuperQ:
Thanks for the idea, but we're no longer using Travis here.

Issue #713, opened 2020-11-26
Error while building blackbox_exporter in Docker builder image


devops-42:
### Host operating system: output of `uname -a`
Builder image for go (Docker): golang:1.15.5-buster

### blackbox_exporter version: output of `blackbox_exporter -version`

No success, `make` failed with error

### What is the blackbox.yml module config.
Not relevant

### What is the prometheus.yml scrape config.
Not relevant

### What logging output did you get from adding `&debug=true` to the probe URL?
Not relevant

### What did you do that produced an error?
I tried to build blackbox_exporter with Docker using a builder image. My Dockerfile looks like:

```
FROM golang:1.15.5-buster

ARG ARCH="amd64"
ARG OS="linux"

USER root

RUN apt-get update && apt-get upgrade && \
      git \
      gcc \
      curl

RUN git clone https://github.com/prometheus/blackbox_exporter.git /go/src/github.com/prometheus/blackbox_exporter

ENV GOPATH=/go

WORKDIR /go/src/github.com/prometheus/blackbox_exporter

RUN make
[...]
```

When starting the Docker build the last step aborts with a failure (most likely in testing DNS). 

### What did you expect to see?
Issuing the `make` command finished successfully. Below an example of a successful build on a VM:

```
[...]
>> running check for unused/missing packages in go.mod
GO111MODULE=on go mod tidy
>> running check for unused packages in vendor/
GO111MODULE=on go mod vendor
>> building binaries
GO111MODULE=on /root/go/bin/promu build --prefix /root/blackbox_exporter 
 >   blackbox_exporter
>> running all tests
GO111MODULE=on go test -race  -mod=vendor ./...
ok  	github.com/prometheus/blackbox_exporter	(cached)
ok  	github.com/prometheus/blackbox_exporter/config	(cached)
ok  	github.com/prometheus/blackbox_exporter/prober	(cached)
```

### What did you see instead?
```
[...]
>> running check for unused packages in vendor/
GO111MODULE=on go mod vendor
curl -s -L https://github.com/prometheus/promu/releases/download/v0.5.0/promu-0.5.0.linux-amd64.tar.gz | tar -xvzf - -C /tmp/tmp.rwLEuU3EZH
promu-0.5.0.linux-amd64/
promu-0.5.0.linux-amd64/promu
promu-0.5.0.linux-amd64/NOTICE
promu-0.5.0.linux-amd64/LICENSE
mkdir -p /go/bin
[...]

PR #722, opened 2020-12-04
Update build

SuperQ:
* Move make test from Travis to CircleCI.
* Drop Travis testing.
* Bump prometheus orb to latest.


SuperQ:
Hmm, the gotestsum test seems to stall. Maybe IPv6 related?

brian-brazil:
It's not immediately obvious from the various retries I've seen/kicked off, it could be related to #713 but I've only seen that failure on v4-only.

SuperQ:
I figured out a couple of the issues.
* We had a TRAVIS specific IPv6 test bypass.
* CircleCI includes `::1 localhost` in the container /etc/hosts file.

brian-brazil:
:+1:

PR #728, opened 2021-01-03
Use preferred protocol first when resolving hostname

### LINKED ISSUE #724: icmp: Both A and AAAA records are queried regardless of preferred_ip_protocol/ip_protocol_fallback
My local DNS cache server doesn't cache results for record that doesn't exist. E.g. if a domain don't have a configured AAAA record, a DNS query to the AAAA record will not be cached. I have no idea if this is the expected behavior (The server is https://github.com/AdguardTeam/AdGuardHome FWIW).

This leads to a higher icmp probe latency for domain names that doesn't have an AAAA record.

Reading the code, it looks like the prober is always sending A/AAAA DNS queries regardless of the preferred_ip_protocol and ip_protocol_fallback by using `resolver.LookupIPAddr(...)`.

https://github.com/prometheus/blackbox_exporter/blob/998037b5b40c1de5fee348ffdea8820509d85171/prober/utils.go#L67

Would it make more sense to call `resolver.LookupIP` with the `preferred_ip_protocol`, and then make follow up DNS queries only if `ip_protocol_fallback` is true?

Linux nuc 4.19.0-9-amd64 #1 SMP Debian 4.19.118-2+deb10u1 (2020-06-07) x86_64 GNU/Linux

### blackbox_exporter version: output of `blackbox_exporter --version`
blackbox_exporter, version 0.18.0 (branch: HEAD, revision: 60c86e6ce5a1111f7958b06ae7a08222bb6ec839)
  build user:       root@53d72328d93f
  build date:       20201012-09:46:31
  go version:       go1.15.2

```
modules:
  icmp:
    prober: icmp
    timeout: 60s
    icmp:
      preferred_ip_protocol: "ip4"
```

```
- job_name: internet-ping
  metrics_path: /probe
  params:
    module: [icmp]
  static_configs:
  - targets:
    - your-dns.run
    - google.com
 ```

```
Logs for the probe:
ts=2020-12-21T18:01:21.831656621Z caller=main.go:304 module=icmp target=your-dns.run level=info msg="Beginning probe" probe=icmp timeout_seconds=60
ts=2020-12-21T18:01:21.831869869Z caller=icmp.go:91 module=icmp target=your-dns.run level=info msg="Resolving target address" ip_protocol=ip4
ts=2020-12-21T18:01:21.875056949Z caller=icmp.go:91 module=icmp target=your-dns.run level=info msg="Resolved target address" ip=35.247.39.128
ts=2020-12-21T18:01:21.875134792Z caller=main.go:119 module=icmp target=your-dns.run level=info msg="Creating socket"
ts=2020-12-21T18:01:21.875204367Z caller=main.go:119 module=icmp target=your-dns.run level=debug msg="Unable to do unprivileged listen on socket, will attempt privileged" err="socket: permission denied"
ts=2020-12-21T18:01:21.875373799Z caller=main.go:119 module=icmp target=your-dns.run level=info msg="Creating ICMP packet" seq=52293 id=50814
ts=2020-12-21T18:01:21.875425358Z caller=main.go:119 module=icmp target=your-dns.run level=info msg="Writing out packet"
ts=2020-12-21T18:01:21.875576684Z caller=main.go:119 module=icmp target=your-dns.run level=info msg="Waiting for reply packets"
ts=2020-12-21T18:01:21.900723776Z caller=main.go:119 module=icmp target=your-dns.run level=info msg="Found matching reply packet"
ts=2020-12-21T18:01:21.900897095Z caller=main.go:304 module=icmp target=your-dns.run level=info msg="Probe succeeded" duration_seconds=0.069158382



probe_dns_lookup_time_seconds 0.043219132
probe_duration_seconds 0.069158382
# HELP probe_icmp_duration_seconds Duration of icmp request by phase
# TYPE probe_icmp_duration_seconds gauge
probe_icmp_duration_seconds{phase="resolve"} 0.043219132
probe_icmp_duration_seconds{phase="rtt"} 0.025207606
probe_icmp_duration_seconds{phase="setup"} 0.000290841
# HELP probe_icmp_reply_hop_limit Replied packet hop limit (TTL for ipv4)
# TYPE probe_icmp_reply_hop_limit gauge
probe_icmp_reply_hop_limit 57
probe_ip_addr_hash 7.89474712e+08
probe_ip_protocol 4
# TYPE probe_success gauge
probe_success 1



Module configuration:
prober: icmp
timeout: 1m0s
http:
    ip_protocol_fallback: true
tcp:
    ip_protocol_fallback: true
icmp:
    preferred_ip_protocol: ip4
    ip_protocol_fallback: true
dns:
    ip_protocol_fallback: true
```

N/A (no error)

DNS resolve latency should be very low (using a local cached DNS server)

### What did you see instead?

DNS resolve latency is higher than expected:

probe_icmp_duration_seconds{phase="resolve"} 0.043219132

### PR DESCRIPTION
**Pull request by yegle**
Prior to this change, LookupIPAddr() is used to do the DNS query,
which sends two DNS queries for A/AAAA records, even if the config has
preferred protocol set to ip4 and does not allow fallback protocol.

This change will change to use LookupIP() with the preferred protocol,
and only try fallback protocol if it's set to true. In most cases doing
this will save one RTT of DNS query.

Fixes #724

### DIFF
--- a/prober/utils.go
+++ b/prober/utils.go
@@ -85,33 +85,31 @@ func chooseProtocol(ctx context.Context, IPProtocol string, fallbackIPProtocol b
 		probeDNSLookupTimeSeconds.Add(lookupTime)
 	}()
 
-	r := &resolver{
-		Resolver: net.Resolver{},
-	}
+	resolver := &net.Resolver{}
 
 	level.Info(logger).Log("msg", "Resolving target address", "ip_protocol", IPProtocol)
-	if ip, err := r.resolve(ctx, target, IPProtocol); err == nil {
-		level.Info(logger).Log("msg", "Resolved target address", "ip", ip.String())
+	if ips, err := resolver.LookupIP(ctx, IPProtocol, target); err == nil {
+		level.Info(logger).Log("msg", "Resolved target address", "ip", ips[0].String())
 		probeIPProtocolGauge.Set(protocolToGauge[IPProtocol])
-		probeIPAddrHash.Set(ipHash(ip.IP))
-		return ip, lookupTime, nil
+		probeIPAddrHash.Set(ipHash(ips[0]))
+		return &net.IPAddr{IP: ips[0]}, lookupTime, nil
 	} else if !fallbackIPProtocol {
 		level.Error(logger).Log("msg", "Resolution with IP protocol failed", "err", err)
 		return nil, 0.0, fmt.Errorf("unable to find ip; no fallback: %s", err)
 	}
 
 	level.Info(logger).Log("msg", "Resolving target address", "ip_protocol", fallbackProtocol)
-	ip, err = r.resolve(ctx, target, fallbackProtocol)
+	ips, err := resolver.LookupIP(ctx, fallbackProtocol, target)
 	if err != nil {
 		// This could happen when the domain don't have A and AAAA record (e.g.
 		// only have MX record).
 		level.Error(logger).Log("msg", "Resolution with IP protocol failed", "err", err)
 		return nil, 0.0, fmt.Errorf("unable to find ip; exhausted fallback: %s", err)
 	}
-	level.Info(logger).Log("msg", "Resolved target address", "ip", ip.String())
+	level.Info(logger).Log("msg", "Resolved target address", "ip", ips[0].String())
 	probeIPProtocolGauge.Set(protocolToGauge[fallbackProtocol])
-	probeIPAddrHash.Set(ipHash(ip.IP))
-	return ip, lookupTime, nil
+	probeIPAddrHash.Set(ipHash(ips[0]))
+	return &net.IPAddr{IP: ips[0]}, lookupTime, nil
 }
 
 func ipHash(ip net.IP) float64 {
--- a/prober/utils.go
+++ b/prober/utils.go
@@ -15,6 +15,7 @@ package prober
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"hash/fnv"
 	"net"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by yegle** [comment]
cc @brian-brazil

File: prober/utils_test.go [hunk 1]
```diff
@@ -162,14 +163,25 @@ func TestChooseProtocol(t *testing.T) {
...
 		t.Error(err)
 	} else if err == nil {
 		t.Error("should set error")
 	}
 	if ip != nil {
 		t.Error("without fallback it should not answer")
 	}
+
+	registry = prometheus.NewPedanticRegistry()
+	ip, _, err = chooseProtocol(ctx, "ip4", true, "does-not-exist.google.com", registry, logger)
```
brian-brazil, line 176: It's probably better to work from example.com, less chance of such a subdomain popping up.

prober/utils.go
```diff
@@ -26,6 +27,28 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+var protocolToGauge = map[string]float64{
+	"ip4": 4,
+	"ip6": 6,
+}
+
+type resolver struct {
```
brian-brazil, line 35: This seems unnecessary, we only call one method of it.

prober/utils.go
```diff
@@ -26,6 +27,28 @@ import (
...
+// A simple wrapper around resolver.LookupIP.
+func (r *resolver) resolve(ctx context.Context, target string, protocol string) (*net.IPAddr, error) {
+	ips, err := r.LookupIP(ctx, protocol, target)
+	if err != nil {
+		return nil, err
+	}
+	for _, ip := range ips {
+		return &net.IPAddr{IP: ip}, nil
+	}
+	// Go doc did not specify when this could happen, better be defensive.
```
brian-brazil, line 48: The comments are clear, there's at least one address when it's nil. So this is redundant.

It's be best to inline this function, given that it's simple.

prober/utils.go
yegle: Yes but the logic in `resolve` is used twice. I can have a function that takes ctx, resolver, target, protocol and eliminate this wrapper, but it seems like this is too much parameters.

Please let me know if keeping this wrapper is acceptable.

prober/utils.go
yegle: See the other comment, the function is called twice so inlining the function will complicate the code a bit.

yegle: Hmm, I also don't see in the doc indicating "when the err is nil, there's at least one address": https://golang.org/pkg/net/#Resolver.LookupIP

brian-brazil: You have to dig a bit, it's a comment on one of the underlying functions.

prober/utils.go
brian-brazil: This type appears unneeded. Also I don't see an issue with duplicating 4 lines of code.

brian-brazil:
Thanks!

yegle:
I wonder if a new version can cut with this PR included?

Or alternatively can you publish a nightly docker image?

crockk:
Does the docker image at prom/blackbox-exporter:master include this merge?

hferreira23:
@crockk It should be even though I'm actually using the master tag image and I see no change in the behavior.

Either way this PR will be on 0.19 which for the looks of it it's just around the corner.

james58899:
Also related to https://github.com/golang/go/issues/45024, need wait golang 1.17 release and upgrade.

Issue #717, opened 2020-12-02
Validating config should try to compile regexes


roidelapluie:
### blackbox_exporter version:

10b04fc9fd220343938233c5586556f7e84a2f45

### What is the blackbox.yml module config.

```
modules:
  http_proxy_sna_monitor:
    prober: http
    http:
      fail_if_body_not_matches_regexp: ['{"status":"ok","components":[']
```

### logs

```
ts=2020-12-02T09:22:07.826147753Z caller=http.go:456 module=http_proxy_sna_monitor target=https://X/status/monitor level=error msg="Could not compile regular expression" regexp="{\"status\":\"ok\",\"components\":[" err="error parsing regexp: missing closing ]: `[`"
```

I think we could compile regexes when reloading config to make this more obvious to users.

### What did you do that produced an error?

I wrote a bad config file with a wring regex, and got an error at runtime

### What did you expect to see?

An error while reloading config, not runtime.

### What did you see instead?

Error at probe time.

PR #729, opened 2021-01-05
Solve #717 Validating config should try to compile regexes

sshota0809:
Hello, This is my first contribution.

I solved issue #717.
I modified Validating config so that it includes trying to compile regexes and then output error if there are invalid regexes. Also I added test related to them.

Please check if my commit is correct. Thank you for your consideration.!


brian-brazil:
Thanks for your PR, could we use the same approach as in https://github.com/prometheus/prometheus/blob/master/pkg/relabel/relabel.go ?

sshota0809:
@brian-brazil 
Thank you for comment!
I have modified them based on your request. Please check if it seems what you intend.

prober/tcp.go
```diff
@@ -146,19 +145,14 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 	for i, qr := range module.TCP.QueryResponse {
 		level.Info(logger).Log("msg", "Processing query response entry", "entry_number", i)
 		send := qr.Send
-		if qr.Expect != "" {
-			re, err := regexp.Compile(qr.Expect)
-			if err != nil {
-				level.Error(logger).Log("msg", "Could not compile into regular expression", "regexp", qr.Expect, "err", err)
-				return false
-			}
+		if qr.Expect.Regexp != nil && qr.Expect.Regexp.String() != "" {
```
brian-brazil, line 148: Isn't Regexp always non-nil?

config/config.go
```diff
@@ -303,7 +355,7 @@ func (s *HeaderMatch) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return errors.New("header name must be set for HTTP header matchers")
 	}
 
-	if s.Regexp == "" {
+	if s.Regexp.Regexp == nil {
```
brian-brazil, line 358: Isn't this never nil, and you should be checking for the empty string?

I think you're being inconsistent in how you're treating this.

sshota0809: It's possible that `s.Regexp.Regexp` is nil when following case. (http.fail_if_header_not_matches.regexp isn't explicitly specified.)
```
modules:
  http_headers:
    prober: http
    timeout: 5s
    http:
      fail_if_header_not_matches:
        - header: Access-Control-Allow-Origin
          allow_missing: false
```
And as you said I have to check for the empty string, so I added its condition.

prober/tcp.go
sshota0809: > Isn't Regexp always non-nil?

There is the case that config.Module instance is created in `prober/tcp_test.go`.
https://github.com/prometheus/blackbox_exporter/blob/49ae740bd6f5626585cef994c294bba00a5ff02a/prober/tcp_test.go#L359-L378

And when `Expect` isn't explicitly specified as following line, it's possible that `qr.Expect.Rege` is nil.
https://github.com/prometheus/blackbox_exporter/blob/49ae740bd6f5626585cef994c294bba00a5ff02a/prober/tcp_test.go#L364

But that occors only in the test senario. In this case, should I modify `prober/tcp_test.go` as following.

```
   {Expect: config.MustNewRegexp(""), Send: "EHLO tls.prober"}, 
```

Could you give me an advice.

config/config.go
```diff
@@ -289,6 +337,10 @@ func (s *QueryResponse) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := unmarshal((*plain)(s)); err != nil {
 		return err
 	}
+	if s.Expect.Regexp == nil {
+		s.Expect = MustNewRegexp("")
```
brian-brazil, line 341: Not being set and being the empty string are different, if there's handling here it should be in the probe rather than here.

prober/tcp.go
brian-brazil: You just need the first predicate here, only check for nil. Expecting an empty response is valid after all.

config/config.go
sshota0809: I deleted this IF statement to accept nil.

prober/tcp.go
sshota0809: > You just need the first predicate here, only check for nil.
Thank you. I deleted the latter predicate.

sshota0809:
@brian-brazil 
I appreciate that you've spent much time to review code I wrote.

I'm done with modifying what you mentioned.
Would you check it again?

brian-brazil:
Thanks!

PR #749, opened 2021-02-18
Ability to randomly test multiple URL's

reasland:
Hello,

To preface I am very much a beginner at Go, so feel free to adapt my code if this functionality interests you

I created a little if function that will split URL's by ","
Basically for my own monitoring uses this functionality will be useful as I want to test an array of webpages, but I also don't want to hit these pages often enough where it will cause them to cache.

As an example, if i want to test multiple product pages, hitting any 1 page multiple times will cause the site to cache it, and through experience if there are currently issues the pages blackbox exporter is currently hitting will output probe_status 1

    - targets:
      - www.trekbikes.com/p/32565/,www.trekbikes.com/p/33136/,... 

Pretty dumb fix, but just wanted to throw in PR to see if there's any interest in this functionality

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -10,7 +10,6 @@ github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuy
 github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 h1:JYp7IbQjafoB+tBA3gMyHYHrpOtNuDiK/uB5uXxq5wM=
 github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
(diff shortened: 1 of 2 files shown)


roidelapluie:
Thank you, but I do not think this would benefit the majority our users, it seems pretty niche. I would also point that , is a valid character in a URL.

The best is probably for you to test both that you can reach the website behind the load balancer, additonally to the load balancer itself.

reasland:
Fair enough, probably would be best to do some type of array parsing so instead of just splitting by comma it's an array

["www.trekbikes.com/p/32565","...",...]

Something I can figure out.

But yeah you're right that this is most likely a niche use-case, so no issues from me if you don't want me tainting this repo 😄

roidelapluie:
What about you directly querying the backend(s), in addition to the load balancer? That is how I would personally address this.

reasland:
definitely a thought I've been toying with on my side, but I believe there is some level of cache sharing behind nodes, i won't go too far in-depth with my environment as i struggle to understand it myself, but i can definitely discuss with my team better solutions like that

i just quick made this code as it seemed like an easy solution, but i guarantee there's much much better ways of implementation than what i am doing, i implemented blackbox as a whim

roidelapluie:
there is also a possibility that your cache would not cache if there is a GET parameter, like https://www.trekbikes.com/p/32565/?blackbox

reasland:
that's actually a good idea, thanks for your help haha

reasland:
I think I found a solution, some components on my site aren't cached, so I can just check the existence of them, I'll close the PR

Thanks!

roidelapluie:
Thanks!!

PR #752, opened 2021-02-25
Add health endpoint

adebasi:
The existing endpoints expose potentially sensitive data. So they cannot be used without authentication to check from the internet if the services is running.

### DIFF
--- a/main.go
+++ b/main.go
@@ -309,6 +309,10 @@ func run() int {
 			}
 		})
 	http.Handle(path.Join(*routePrefix, "/metrics"), promhttp.Handler())
+	http.HandleFunc(path.Join(*routePrefix, "/-/healthy"), func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("Healthy"))
+	})
 	http.HandleFunc(path.Join(*routePrefix, "/probe"), func(w http.ResponseWriter, r *http.Request) {
 		sc.Lock()
 		conf := sc.C


SuperQ:
We should add this to the exporter toolkit as part of the landing page boilerplate functions.

Cc @roidelapluie

roidelapluie:
Yes, it would go to exporter-toolkit. Also note that Prometheus uses /-/healthy.

adebasi:
Hey folks, thanks for your quick reply. Unfortunately, I don't fully understand what you mean. 

Would you like to move the endpoint definition to the exporter-toolkit or maybe only exclude it there from auth mechanisms?
@roidelapluie Is your comment about the naming/path for the endpoint?

main.go
```diff
@@ -309,6 +309,10 @@ func run() int {
 			}
 		})
 	http.Handle(path.Join(*routePrefix, "/metrics"), promhttp.Handler())
+	http.HandleFunc(path.Join(*routePrefix, "/health"), func(w http.ResponseWriter, r *http.Request) {
```
roidelapluie, line 312: Please use /-/healthy

main.go
```diff
@@ -309,6 +309,10 @@ func run() int {
...
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("OK"))
```
roidelapluie, line 314: Can we use "Healthy" ?

roidelapluie (commented):
Let's move forward here for now :)

we were refering to the exporter-toolkit repository but it is unclear when we will have the landing page here.

main.go
adebasi: Done

adebasi:
@SuperQ or @roidelapluie, can you please merge the PR?

PR #758, opened 2021-03-14
http probe - fix how the tls phase is calculated

kgersen:
I have issue with some _tls_ phase with the http probe. It appears it is caused by the way it's calculated (`trace.gotConn - trace.dnsDone` ) which isn't always reliable.
This proposed fix uses `TLSHandshakeStart` and `TLSHandshakeDone` hooks.

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -549,11 +549,12 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			continue
 		}
 		if trace.tls {
+			// dnsDone must be set if gotConn was set.
+			durationGaugeVec.WithLabelValues("connect").Add(trace.connectDone.Sub(trace.dnsDone).Seconds())
 			durationGaugeVec.WithLabelValues("tls").Add(trace.tlsDone.Sub(trace.tlsStart).Seconds())
+		} else {
+			durationGaugeVec.WithLabelValues("connect").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
 		}
-
-		// actual connection - we could add a new phase between connectDone and gotConn
-		durationGaugeVec.WithLabelValues("connect").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
 
 		// Continue here if we never got a response from the server.
 		if trace.responseStart.IsZero() {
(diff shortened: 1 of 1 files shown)


prober/http.go
```diff
@@ -411,6 +424,8 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		ConnectDone:          tt.ConnectDone,
 		GotConn:              tt.GotConn,
 		GotFirstResponseByte: tt.GotFirstResponseByte,
+		TLSHandshakeStart:    tt.TLSHandshakeStart,
```
mem, line 427: this got introduced in Go 1.8, which is fine, as we are already requiring at least Go 1.13

prober/http.go
```diff
@@ -532,13 +549,12 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			continue
 		}
 		if trace.tls {
-			// dnsDone must be set if gotConn was set.
-			durationGaugeVec.WithLabelValues("connect").Add(trace.connectDone.Sub(trace.dnsDone).Seconds())
-			durationGaugeVec.WithLabelValues("tls").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
-		} else {
-			durationGaugeVec.WithLabelValues("connect").Add(trace.gotConn.Sub(trace.dnsDone).Seconds())
+			durationGaugeVec.WithLabelValues("tls").Add(trace.tlsDone.Sub(trace.tlsStart).Seconds())
```
mem, line 552: this is subtly changing the meaning of the tls phase, but I think it's OK since this reflects more accurately the duration of the TLS handshake. The existing method is measuring the handshake plus whatever time it takes to finish setting the connection up plus whatever time it takes to get to the point where the TLS handshake starts after DNS is done.

mem (approved):

kgersen:
with this fix I ran in a new issue: the global "probe_duration_seconds" metric doesn't match the sum of the phases anymore like before.
The tls phase duration is now counted approximately twice since it's also part of 'connect' phase.
A fix would be to subtract "tls" from "connect" or this will introduces a breaking change (breaking stacked graphs for instance).

dswarbrick:
Out of curiosity, is this somehow related to #485 ?

PR #764, opened 2021-03-16
Add support for decompression of HTTP responses

### LINKED ISSUE #684: Handling of compressed HTTP responses

Linux clara 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

```
blackbox_exporter, version 0.17.0 (branch: fix_content_length, revision: 58e365f15172f93709d2e4de4863bc12a8aef368)
  build user:       marcelo@clara
  build date:       20200825-16:25:42
  go version:       go1.14.7
```

```
modules:
  http:
    prober: http
    http:
      method: GET
      headers:
        Accept-Encoding: gzip
      fail_if_body_matches_regexp:
        - "<!DOCTYPE html>"
      preferred_ip_protocol: "ip4"
      ip_protocol_fallback: false
```

None.

```
Logs for the probe:
ts=2020-08-25T16:28:51.176627Z caller=main.go:304 module=http target=https://prometheus.io/ level=info msg="Beginning probe" probe=http timeout_seconds=119.5
ts=2020-08-25T16:28:51.177415706Z caller=http.go:342 module=http target=https://prometheus.io/ level=info msg="Resolving target address" ip_protocol=ip4
ts=2020-08-25T16:28:51.255273005Z caller=http.go:342 module=http target=https://prometheus.io/ level=info msg="Resolved target address" ip=172.67.141.23
ts=2020-08-25T16:28:51.255468626Z caller=client.go:252 module=http target=https://prometheus.io/ level=info msg="Making HTTP request" url=https://172.67.141.23/ host=prometheus.io
ts=2020-08-25T16:28:51.51694598Z caller=main.go:119 module=http target=https://prometheus.io/ level=info msg="Received HTTP response" status_code=200
ts=2020-08-25T16:28:51.517282451Z caller=main.go:119 module=http target=https://prometheus.io/ level=info msg="Response timings for roundtrip" roundtrip=0 start=2020-08-25T10:28:51.255713943-06:00 dnsDone=2020-08-25T10:28:51.255713943-06:00 connectDone=2020-08-25T10:28:51.308865333-06:00 gotConn=2020-08-25T10:28:51.371621155-06:00 responseStart=2020-08-25T10:28:51.516747842-06:00 end=2020-08-25T10:28:51.517265582-06:00
ts=2020-08-25T16:28:51.51759115Z caller=main.go:304 module=http target=https://prometheus.io/ level=info msg="Probe succeeded" duration_seconds=0.340772252



probe_dns_lookup_time_seconds 0.077859237
probe_duration_seconds 0.340772252
# HELP probe_failed_due_to_regex Indicates if probe failed due to regex
# TYPE probe_failed_due_to_regex gauge
probe_failed_due_to_regex 0
# HELP probe_http_content_length Length of http content response
# TYPE probe_http_content_length gauge
probe_http_content_length -1
# HELP probe_http_duration_seconds Duration of http request by phase, summed over all redirects
# TYPE probe_http_duration_seconds gauge
probe_http_duration_seconds{phase="connect"} 0.053151324
probe_http_duration_seconds{phase="processing"} 0.145126765
probe_http_duration_seconds{phase="resolve"} 0.077859237
probe_http_duration_seconds{phase="tls"} 0.115907188
probe_http_duration_seconds{phase="transfer"} 0.000517688
# HELP probe_http_redirects The number of redirects
# TYPE probe_http_redirects gauge
probe_http_redirects 0
# HELP probe_http_ssl Indicates if SSL was used for the final redirect
# TYPE probe_http_ssl gauge
probe_http_ssl 1
# HELP probe_http_status_code Response HTTP status code
# TYPE probe_http_status_code gauge
probe_http_status_code 200
# HELP probe_http_uncompressed_body_length Length of uncompressed response body
# TYPE probe_http_uncompressed_body_length gauge
probe_http_uncompressed_body_length 4585
# HELP probe_http_version Returns the version of HTTP of the probe response
# TYPE probe_http_version gauge
probe_http_version 2
probe_ip_addr_hash 2.601109704e+09
probe_ip_protocol 4
# HELP probe_ssl_earliest_cert_expiry Returns earliest SSL cert expiry in unixtime
# TYPE probe_ssl_earliest_cert_expiry gauge
probe_ssl_earliest_cert_expiry 1.628424e+09
# HELP probe_ssl_last_chain_expiry_timestamp_seconds Returns last SSL chain expiry in timestamp seconds
# TYPE probe_ssl_last_chain_expiry_timestamp_seconds gauge
probe_ssl_last_chain_expiry_timestamp_seconds 1.628424e+09
# HELP probe_ssl_last_chain_info Contains SSL leaf certificate information
# TYPE probe_ssl_last_chain_info gauge
probe_ssl_last_chain_info{fingerprint_sha256="44c9e62838db98e79918c841e4b72529849e2e7e6654e337a32a74ec502edaa7"} 1
# TYPE probe_success gauge
probe_success 1
# HELP probe_tls_version_info Contains the TLS version used



Module configuration:
prober: http
http:
    preferred_ip_protocol: ip4
    method: GET
    headers:
        Accept-Encoding: gzip
    fail_if_body_matches_regexp:
        - <!DOCTYPE html>
tcp:
    ip_protocol_fallback: true
icmp:
    ip_protocol_fallback: true
dns:
    ip_protocol_fallback: true
```

See configuration above. Target is https://prometheus.io/

probe_success 0

### What did you see instead?

probe_success 1

By adding:

```
    headers:
        Accept-Encoding: gzip
```

the server is correctly returning a compressed response. The _uncompressed_ body matches the specified regular expression, and my expectation is that the probe fails.

BBE is not decompressing the response and the regular expression is being applied to the compressed response, and therefore does not match.

mem:
If the module configuration specifies the "compression" option
blackbox_exporter will try to decompress the response using the
specified algorithm. If the response is not compressed using that
algorithm, the probe will fail.

It validates that the "Accept-Encoding" header is either absent, or that
it specifies the same algorithm as the "compression" option. If the
"Accept-Encoding" header is present but it specifies a different
algorithm, the probe will fail.

If the compression option is *not* used, probe_http_content_length and
probe_http_uncompressed_body_length will have the same value
corresponding to the original content length. If the compression option
is used and the content can be decompressed, probe_http_content_length
will report the original content length as it currently does, and
probe_http_uncompressed_body_length will report the length of the body
after decompression as expected.

Fixes #684

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -14,6 +14,8 @@
 package prober
 
 import (
+	"compress/flate"
+	"compress/gzip"
 	"context"
 	"crypto/tls"
 	"errors"
@@ -31,6 +33,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/andybalholm/brotli"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
@@ -414,6 +417,7 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			request.Host = value
 			continue
 		}
+
 		request.Header.Set(key, value)
 	}
 
@@ -470,6 +474,31 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			}
 		}
 
+		// Since the configuration specifies a compression algorithm, blindly treat the response body as a
+		// compressed payload; if we cannot decompress it it's a failure because the configuration says we
+		// should expect the response to be compressed in that way.
+		if httpConfig.Compression != "" {
+			dec, err := getDecompressionReader(httpConfig.Compression, resp.Body)
+			if err != nil {
+				level.Info(logger).Log("msg", "Failed to get decompressor for HTTP response body", "err", err)
+				success = false
+			} else if dec != nil {
+				// Since we are replacing the original resp.Body with the decoder, we need to make sure
+				// we close the original body. We cannot close it right away because the decompressor
+				// might not have read it yet.
+				defer func(c io.Closer) {
+					err := c.Close()
+					if err != nil {
+						// At this point we cannot really do anything with this error, but log
+						// it in case it contains useful information as to what's the problem.
+						level.Info(logger).Log("msg", "Error while closing response from server", "err", err)
+					}
+				}(resp.Body)
+
+				resp.Body = dec
+			}
+		}
+
 		byteCounter := &byteCounter{ReadCloser: resp.Body}
 
 		if success && (len(httpConfig.FailIfBodyMatchesRegexp) > 0 || len(httpConfig.FailIfBodyNotMatchesRegexp) > 0) {
@@ -491,8 +520,9 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			respBodyBytes = byteCounter.n
 
 			if err := byteCounter.Close(); err != nil {
-				// We have already read everything we could from the server. The error here might be a
-				// TCP error. Log it in case it contains useful information as to what's the problem.
+				// We have already read everything we could from the server, maybe even uncompressed the
+				// body. The error here might be either a decompression error or a TCP error. Log it in
+				// case it contains useful information as to what's the problem.
 				level.Info(logger).Log("msg", "Error while closing response from server", "error", err.Error())
 			}
 		}
@@ -594,4 +624,23 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	bodyUncompressedLengthGauge.Set(float64(respBodyBytes))
 	redirectsGauge.Set(float64(redirects))
 	return
+}
+
+func getDecompressionReader(algorithm string, origBody io.ReadCloser) (io.ReadCloser, error) {
+	switch strings.ToLower(algorithm) {
+	case "br":
+		return ioutil.NopCloser(brotli.NewReader(origBody)), nil
+
+	case "deflate":
+		return flate.NewReader(origBody), nil
+
+	case "gzip":
+		return gzip.NewReader(origBody)
+
+	case "identity", "":
+		return origBody, nil
+
+	default:
+		return nil, errors.New("unsupported compression algorithm")
+	}
 }
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -49,6 +49,18 @@ The other placeholders are specified separately.
   headers:
     [ <string>: <string> ... ]
 
+  # The compression algorithm to use to decompress the response (gzip, br, deflate, identity).
+  #
+  # If an "Accept-Encoding" header is specified, it MUST be such that the compression algorithm
+  # indicated using this option is acceptable. For example, you can use `compression: gzip` and
+  # `Accept-Encoding: br, gzip` or `Accept-Encoding: br;q=1.0, gzip;q=0.9`. The fact that gzip is
+  # acceptable with a lower quality than br does not invalidate the configuration, as you might
+  # be testing that the server does not return br-encoded content even if it's requested. On the
+  # other hand, `compression: gzip` and `Accept-Encoding: br, identity` is NOT a valid
+  # configuration, because you are asking for gzip to NOT be returned, and trying to decompress
+  # whatever the server returns is likely going to fail.
+  [ compression: <string> | default = "" ]
+
   # Whether or not the probe will follow any redirects.
   [ no_follow_redirects: <boolean> | default = false ]
 
--- a/config/config.go
+++ b/config/config.go
@@ -16,9 +16,13 @@ package config
 import (
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"regexp"
 	"runtime"
+	"sort"
+	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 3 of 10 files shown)


prober/http.go
```diff
@@ -397,10 +400,23 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
 			request.Host = value
 			continue
 		}
+
+		// If there's a compression setting, and there's also an
+		// accept-encoding header, they MUST match, otherwise we
+		// end up requesting one encoding and trying to process
+		// a different one, which is more likely than not going
+		// to fail.
+		if httpConfig.Compression != "" && normalizedKey == "Accept-Encoding" && !strings.EqualFold(value, httpConfig.Compression) {
```
roidelapluie, line 415: Not correct: `Accept-Encoding: br;q=1.0, gzip;q=0.8, *;q=0.1`

prober/http.go
```diff
@@ -455,6 +471,31 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			}
 		}
 
+		// Since the configuration specifies a compression algorithm, blindly treat the response body as a
+		// compressed payload; if we cannot decompress it it's a failure because the configuration says we
+		// should expect the response to be compressed in that way.
+		if httpConfig.Compression != "" {
+			dec, err := getDecompressionReader(httpConfig.Compression, resp.Body)
+			if err != nil {
+				level.Info(logger).Log("msg", "Failed to get decompressor for HTTP response body", "err", err.Error())
```
roidelapluie, line 480:
```suggestion
				level.Info(logger).Log("msg", "Failed to get decompressor for HTTP response body", "err", err)
```

prober/http.go
```diff
@@ -455,6 +471,31 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+			} else if dec != nil {
+				// Since we are replacing the original resp.Body with the decoder, we need to make sure
+				// we close the original body. We cannot close it right away because the decompressor
+				// might not have read it yet.
+				defer func(c io.Closer) {
+					err := c.Close()
+					if err != nil {
+						// At this point we cannot really do anything with this error, but log
+						// it in case it contains useful information as to what's the problem.
+						level.Info(logger).Log("msg", "Error while closing response from server", "error", err.Error())
```
roidelapluie, line 491:
```suggestion
						level.Info(logger).Log("msg", "Error while closing response from server", "error", err)
```

roidelapluie:
Thanks for this! I have left a few comments.

prober/http.go
mem: I was wondering about that case.

The intention of the change is to be very "do what I say". If I allow for "br;q=1.0, gzip;q=0.8" in principle the server can select whatever it wants from the acceptable choices. How would that case work? Or better said, what would the check be checking? For example, this configuration:

```
    compression: gzip
    headers:
        Accept-Encoding: br;q=1.0, gzip;q=0.8
```

if the server responds with br-encoded content, this configuration would still fail because it will try to decompress it as gzip. You can argue that the user is trying to verify that the server selects gzip even if the user says br is preferred.

I think that's an acceptable use case. I can remove the strict checking here, allow the user to specify any Accept-Encoding and still fail in the case described above. I should update the documentation to spell that out.

Would that be OK?

mem: I took a stab at this. I added code to process the Accept-Encoding header, so that the compression algorithm must be acceptable, so this configuration is valid:

```
    compression: gzip
    headers:
        Accept-Encoding: br;q=1.0, gzip;q=0.8
```

and this is invalid:

```
    compression: gzip
    headers:
        Accept-Encoding: br;q=1.0, deflate;q=0.8
```

the second configuration would cause the server to return something that will fail decompression.

prober/http.go
```diff
@@ -397,10 +400,27 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+		// If there's a compression setting, and there's also an
+		// accept-encoding header, they MUST match, otherwise we
+		// end up requesting something that doesn't include the
+		// specified compression, and that's likely to fail,
+		// depending on how the server is configured. Testing
+		// that the server _ignores_ Accept-Encoding, e.g. by
+		// not including a particular compression in the header
+		// but expecting it in the response falls out of the
+		// scope of the tests we perform.
+		if httpConfig.Compression != "" && normalizedKey == "Accept-Encoding" && !isEncodingAcceptable(httpConfig.Compression, value) {
```
roidelapluie, line 419: Can we check that at unmarshalling?

roidelapluie:
Thanks, I have added one last comment.

prober/http.go
mem: Sure! I have moved some of the code around to check the config is valid at unmarshalling.

mem:
> Thanks, I have added one last comment.

Thank you! I think I have addressed everything. Can you take another look?

prober/http.go
```diff
@@ -397,10 +400,13 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 	request = request.WithContext(ctx)
 
 	for key, value := range httpConfig.Headers {
-		if strings.Title(key) == "Host" {
+		normalizedKey := strings.Title(key)
```
roidelapluie, line 403: Why this indirection?

roidelapluie (approved):
LGTM. I do not even think we should check the params like you did because there may be non-standard servers implementations that would use other formats, but for which users might still want to use with this exporter, but let's keep them and gather users feedback.

I have added a minor comment.

prober/http.go
mem: ah, that's a leftover from a different implementation, thanks for catching it

mem:
> LGTM. I do not even think we should check the params like you did because there may be non-standard servers implementations that would use other formats, but for which users might still want to use with this exporter, but let's keep them and gather users feedback.
> 
> I have added a minor comment.

Cool, thank you!

PR #765, opened 2021-03-17
Add source_ip_address support to http probes

bobrik:
This is a logical continuation of #262. It requires https://github.com/prometheus/common/pull/284.

Signed-off-by: Ivan Babrou <github@ivan.computer>

PR #766, opened 2021-03-17
Add source_ip_address support to http probes

bobrik:
This is a logical continuation of #262.

It's the same as #765 but without changing `prometheus/common` as @roidelapluie suggested.

Signed-off-by: Ivan Babrou <github@ivan.computer>

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -97,6 +97,9 @@ The other placeholders are specified separately.
   [ preferred_ip_protocol: <string> | default = "ip6" ]
   [ ip_protocol_fallback: <boolean> | default = true ]
 
+  # The source IP address.
+  [ source_ip_address: <string> ]
+
   # The body of the HTTP request used in probe.
   body: [ <string> ]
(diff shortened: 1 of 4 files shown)


bobrik:
@roidelapluie, this approach doesn't seem to work very well, given that there may be multiple layers of transport.

roidelapluie:
> @roidelapluie, this approach doesn't seem to work very well, given that there may be multiple layers of transport.

Can you be more explicit?

bobrik:
Look at the failed test:

```
panic: interface conversion: http.RoundTripper is *config.basicAuthRoundTripper, not *http.Transport
```

I don't see a good way to update `DialContext` in `http.RoundTripper` that is wrapped by `config.basicAuthRoundTripper`.

roidelapluie:
Hello @bobrik . Now the HTTP client supports passing special dialcontext funcs so we could implement all this in the blackbox exporter: https://github.com/prometheus/common/pull/291

PR #769, opened 2021-03-30
Add test to verify published metrics do not change

mem:
The new checkRegistryMetrics verifies that the metrics produced by a particular probe match the expected metrics. Use this function in tests so that if a change adds or deletes metrics, the test has to be modified accordingly and it doesn't slip thru.

The probes involving TLS are tricky because they contain the fingerprint of the certificate. For the HTTP prober, since we are using the httptest package, the localhost certificate is actually known and fixed, so it's OK to embed the fingerprint in the test. For TCP, we are generating the certificate using random data, so the fingerprint changes each time. Change this to follow a similar strategy as the httptest package. The `prober/internal/testcert.go` file is actually copied from the Go source and it's the same key and certificate used in the httptest.

### DIFF
--- a/prober/icmp_test.go
+++ b/prober/icmp_test.go
@@ -1,0 +1,77 @@
+// Copyright 2016 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


prober/icmp_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright 2016 The Prometheus Authors
```
SuperQ, line 1: Nit, 2021

SuperQ (commented):
I like it. Just needs a rebase.

electron0zero:
+1, would like to merge. needs rebase.

ArthurSens:
Hello hellooo, I'm taking the liberty of closing this PR due to the lack of activity. That said, the idea here is great!

We probably don't need to build and maintain the test helper ourselves, though, since the latest prometheus/client_golang provides most (if not all) of the utilities we need in this package https://github.com/prometheus/client_golang/tree/main/prometheus/testutil

The remaining work is basically using that package in our tests :)

PR #783, opened 2021-05-05
Fix example ssh_banner check: Send identification string

Cytrian:
This is to reduce log spam:

"sshd[3750]: Did not receive identification string from ..:"

There will still be a "Connection closed by" message.


mem (approved):

SuperQ (approved):
LGTM, thanks!

Issue #787, opened 2021-05-11
prom/blackbox-exporter runs as root


030:
```
WARN	- CIS-DI-0001: Create a user for the container
	* Last user should not be root
```

roidelapluie:
I don't think there is an effective way to setcap in containers so I think this is the reason we run as root (for ICMP)

PR #789, opened 2021-05-12
[GH-787] Add nonroot user.

030:


roidelapluie:
Did you try this? I suspect that ping might not work without root, or with more difficulties, which could be a breaking change.

030:
@roidelapluie I tested it locally using docker:

```bash
/ $ cat /root/
cat: can't open '/root/': Permission denied
/ $ ping
BusyBox v1.33.1 (2021-05-03 21:56:41 UTC) multi-call binary.

Usage: ping [OPTIONS] HOST

Send ICMP ECHO_REQUESTs to HOST

	-4,-6		Force IP or IPv6 name resolution
	-c CNT		Send only CNT pings
	-s SIZE		Send SIZE data bytes in packets (default 56)
	-i SECS		Interval
	-A		Ping as soon as reply is recevied
	-t TTL		Set TTL
	-I IFACE/IP	Source interface or IP address
	-W SEC		Seconds to wait for the first response (default 10)
			(after all -c CNT packets are sent)
	-w SEC		Seconds until ping exits (default:infinite)
			(can exit earlier with -c CNT)
	-q		Quiet, only display output at start/finish
	-p HEXBYTE	Payload pattern
```

and on OpenShift3, but not on k8s yet.

roidelapluie:
I mean, the  "icmp" probe of the blackbox exporter is what I would test, not the ping command.

PR #795, opened 2021-05-27
Add http endpoint support

mblaschke:
Add support for custom endpoint (IP/Port or full hostname) to make probes possible in specific and complex environments and to test the service chain:

- support for checking services in a service chain (app gateway -> loadbalancer -> ingress controller -> app) to test http response of eg. the kubernetes ingress (from inside the cluster by using `xxx.xxx.svc.cluster.local` as `endpoint` where you cannot reach the loadbalancer due to cloud limitations) and get the ability to setup alerts on probe failure and TLS certificate expiry
- support for checking of "hidden" services which are not directly reachable (eg where there is a NAT in between cloud and on premise and DNS resolves to on premise IPs)
- easier use of prometheus-operator for such checks/environments

example module configuration:
```
  http_2xx_example_with_endpoint:
    prober: http
    timeout: 5s
    http:
      endpoint: "1.1.1.1:443"
      valid_http_versions: ["HTTP/1.1", "HTTP/2.0"]
      method: GET
      ip_protocol_fallback: false  # no fallback to "ip6"

  http_2xx_ingress_internal:
    prober: http
    timeout: 5s
    http:
      endpoint: "controller.ingress.svc.cluster.local:80"
      valid_http_versions: ["HTTP/1.1", "HTTP/2.0"]
      method: GET
      ip_protocol_fallback: false  # no fallback to "ip6"
```

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -45,6 +45,9 @@ The other placeholders are specified separately.
   # The HTTP method the probe will use.
   [ method: <string> | default = "GET" ]
 
+  # Endpint IP (IP or IP:Portand port) to send HTTP requests, skip DNS lookup.
+  [ endpoint: <string> ]
+
   # The HTTP headers set for the probe.
   headers:
     [ <string>: <string> ... ]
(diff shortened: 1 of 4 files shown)


2mf:
I really would love the `hostname` parameter

roidelapluie:
instead of endpoint, you can use Headers: Host in the current blackbox exporter.

I have  a design doc to enable this per probe: https://docs.google.com/document/d/1VwqXi2TOb5KXaZY6Iio7411x64pJao3GusX8MqYsJ2g/edit

mblaschke:
The extra `hostname` parameter would be also ok. I just want to avoid creating one configuration entry for only one probe.

roidelapluie:
#823 is merged.

PR #796, opened 2021-05-28
Simplify prefered protocol DNS probe

### LINKED ISSUE #791: DNS probing performance regression in 0.19.0
I run blackbox-exporter in docker probing a DNS server in another docker container.

With cache on DNS server end, probing latency is usually ~120us at 0.18.0. After updating to 0.19.0 it jumps to ~450us.

Knowing my patch in https://github.com/prometheus/blackbox_exporter/issues/724 I updated my config with `preferred_ip_protocol: "ip4"` since I don't configure ipv6 in my docker network, but it still result in ~200us latency.

Reverting back to 0.18.0 and now I have consistent ~120us latency again.

roidelapluie:
This change is an attempt to fix #791

### DIFF
--- a/prober/utils.go
+++ b/prober/utils.go
@@ -26,6 +26,11 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+var protocolToGauge = map[string]float64{
+	"ip4": 4,
+	"ip6": 6,
+}
+
 // Returns the IP for the IPProtocol and lookup time.
 func chooseProtocol(ctx context.Context, IPProtocol string, fallbackIPProtocol bool, target string, registry *prometheus.Registry, logger log.Logger) (ip *net.IPAddr, lookupTime float64, err error) {
 	var fallbackProtocol string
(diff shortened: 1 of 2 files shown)


prober/utils.go
```diff
@@ -32,7 +32,7 @@ var protocolToGauge = map[string]float64{
 }
 
 // Returns the IP for the IPProtocol and lookup time.
-func chooseProtocol(ctx context.Context, IPProtocol string, fallbackIPProtocol bool, target string, registry *prometheus.Registry, logger log.Logger) (ip *net.IPAddr, lookupTime float64, returnerr error) {
+func chooseProtocol(ctx context.Context, IPProtocol string, fallbackIPProtocol bool, target string, registry *prometheus.Registry, logger log.Logger) (ip *net.IPAddr, lookupTime float64, err error) {
```
SuperQ, line 35: How about we switch this to un-named returns?

roidelapluie: We set lookupTime in a defer, once for all:

https://github.com/prometheus/blackbox_exporter/pull/796/files#diff-b99c121217776765037cbc7e4dfca853b9ef8243cb2f42f4a95f92c7afbb0361L66-L68

Since we use this 3 or 4 times, I think that the named returns are a better solution that repeating this multiple times.

SuperQ: ok

SuperQ (approved):
LGTM

mem (approved):
LGTM, too!

PR #822, opened 2021-09-02
Update dependencies

mem:


mem:
This is the result of running `go get -u`. I did some sanity checking locally and it seems to be working fine.

roidelapluie:
Can you use `make update-go-deps` instead? that gets less intermediate dependencies.

mem:
> Can you use `make update-go-deps` instead? that gets less intermediate dependencies.

Updated, sorry I missed your reply.

PR #825, opened 2021-09-13
Add support for resolving the default gateway.

tommie:
The magic hostname "default-gateway.internal" resolves to the host's
default gateway.

Currently, the gateway library only supports IPv4. Supported OSs are
FreeBSD, Linux, OS X (Darwin), Solaris, Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Additionally, an [example configuration](example.yml) is also available.
 HTTP, HTTPS (via the `http` prober), DNS, TCP socket and ICMP (see permissions section) are currently supported.
 Additional modules can be defined to meet your needs.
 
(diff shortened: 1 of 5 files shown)


roidelapluie:
This is out of scope for the blackbox exporter.

The node exporter can expose all the routes in a system.

tommie:
Thanks for the quick reply.

The route config itself isn't what I'm looking for. In my case, it's to monitor laptops to see if latencies are due to uplink or WiFi, and the Prometheus server doesn't know the node's gateway. (I.e. I want to ping the gateway and `8.8.8.8`.)

For anyone looking to do the same: I have two other options: (1) fork one of the ping exporters instead, or (2) add a script to maintain an `/etc/hosts` record.

tommie:
And (3) build a dedicated connectivity exporter, of course. :)

roidelapluie:
What does your network topology look like? If it's a lot of /24 and routers in .1, you could use relabeling.

tommie:
They might be anywhere on Earth, not tied to an internal network. Thanks again.

Issue #665, opened 2020-07-16
Feature request: gRPC support


juan-martinez-hernandez:
Hi there,

Do you guys have plans to add support to perform probe checks to gRPC services? 

I've searched for solutions with Blackbox exporter to achieve this but no success.

brian-brazil:
That's not planned, the blackbox exporter scope is more basic protocols like plain HTTP. I'd suggest writing your own exporter if you need this, I'm not aware of one currently existing.

andbuitra:
Couldn't you just query it via HTTP? gRPC basically uses HTTP/2 for communication. I'm totally talking out ignorance, tho, haven't dealt with that technology too much.

kmcminn:
grpc is primarily used for backends, this tool seems designed to probe frontends. You probably dont want to be calling your backend endpoints/rpcs with your monitoring system but instead emitting stats from within them (statsd or similar) and exposing a health endpoint for things like a deploy system or k8s liveness/readyness

kotyara85:
@brian-brazil I can send a PR with grpc support

brian-brazil:
That's for the offer, but I'm not willing to expand the scope of this exporter to cover that. I'd suggest starting a new exporter so that it can handle all the various gRPC use cases as they come up, which can then also be listed on the prometheus website.

Issue #603, opened 2020-04-22
Add gRPC support/probe


cpitstick-argo:
The [existing probes](https://github.com/prometheus/blackbox_exporter/tree/master/prober) probes do not include gRPC; yet it's a [Kubernetes standard](https://github.com/grpc-ecosystem/grpc-health-probe/).

How hard would it be to add a gRPC probe?

brian-brazil:
The blackbox exporter supports the usual standard network protocols, if you want to probe something like gRPC I'd suggest writing a custom exporter.

cpitstick-argo:
Would this custom exporter be checked into master?

Otherwise, what is the proper mechanism for including a custom exporter that I've written myself if I'm installing blackbox with the helm chart?

brian-brazil:
This isn't something that'd be accepted into this codebase, there's just too many protocols out there for us to include everything.

That sounds like a Helm usage question, which I'm afraid I can't help you with.

cpitstick-argo:
Ok, rewind a bit then. If I wasn't using helm, how might I get blackbox to use a custom exporter? From a brief scan of the code it looks like the exporters are built into the docker image, which would require a fork of the code base, a rebuild of the image, and a place to store the custom image, which is a lot for simply one extra protocol.

brian-brazil:
You don't, the blackbox exporter supports what it supports. You'd have to create a new exporter, or hope that someone had already created one.

cpitstick-argo:
Alright :/

PR #835, opened 2021-10-13
Add support for grpc health check

averzicco:
Hi, This PR is to add support for [gRPC health check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) to blackbox exporter.
I've read the discussion on issues #665 and #603, it seems that gRPC is out of scope for blackbox_exporter, if this is still the case just close the PR.


averzicco:
great, I've checked the dev summit video and realized didn't implement the logic to specify the service name for the health check. I've just pushed the code which takes care of that

prober/grpc.go
```diff
@@ -0,0 +1,220 @@
...
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+	"net"
+	"net/url"
+	"strings"
+	"time"
+)
+
+// Health defines a health-check connection.
+type Health interface {
```
roidelapluie, line 36:
```suggestion
type GRPCHealthCheck interface {
```

I'd like to have GRPC in the name.

roidelapluie (approved):
LGTM besides name change.

prober/grpc.go
averzicco: makes sense, changed

roidelapluie:
Thanks!

autumnw:
Appreciate to all the people who joined conversation and contribute code here! This is so awesome! 
Kudos to @averzicco!

PR #836, opened 2021-10-15
Add body_size_limit option to http module

mem:
This option limits the maximum body length that will be read from the
HTTP server. It's meant to prevent misconfigured servers from causing
the probe to use too many resources, even if temporarily. It's not an
additional check on the response, for that, use the resulting metrics
(probe_http_content_length, probe_http_uncompressed_body_length, etc).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -290,7 +290,7 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 
-	if s.BodySizeLimit < 0 {
+	if s.BodySizeLimit <= 0 {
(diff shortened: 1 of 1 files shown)


mem:
> Can we match Prometheus behaviour? Prometheus looks at the uncompressed response.
> 
> ```
> # An uncompressed response body larger than this many bytes will cause the
> # scrape to fail. 0 means no limit. Example: 100MB.
> # This is an experimental feature, this behaviour could
> # change or be removed in the future.
> [ body_size_limit: <size> | default = 0 ]
> ```

OK, I added that for consistency. I used the same name and syntax, too.

roidelapluie:
Does it actually work with decompressed data?

Issue #837, opened 2021-10-19
Unset body_size_limit option causes panic 


vorotech:
### Host operating system
```
Linux ip-xxx-xxx-xxx-xxx 5.11.0-1019-aws #20~20.04.1-Ubuntu SMP Tue Sep 21 10:40:39 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
### blackbox_exporter version: output of `blackbox_exporter --version`
`master` with the merged PR https://github.com/prometheus/blackbox_exporter/pull/836

### What is the blackbox.yml module config.

```yml
modules:
  http_2xx:
    prober: http
    http:
      preferred_ip_protocol: ip4
  http_post_2xx:
    prober: http
    http:
      method: POST
```

### Docker-compose service config.

```yml
version: "3"
services:
  blackbox_exporter:
    image: prom/blackbox-exporter:master
    container_name: blackbox_exporter
    volumes:
      - ./blackbox_exporter_config.yml:/config/blackbox.yml
    command: --config.file=/config/blackbox.yml
    ports:
      - 9115:9115
    restart: always
```

### What did you do that produced an error?

Container is crashed on docker start with the panic.

If `body_size_limit` is set, than service starts without a crash.

### What did you expect to see?

```
level=info ts=2021-10-19T19:48:05.267Z caller=main.go:224 msg="Starting blackbox_exporter" version="(version=0.19.0, branch=master, revision=e86980961f8e463288e31cac03255701cb1a11d1)"
level=info ts=2021-10-19T19:48:05.267Z caller=main.go:225 build_context="(go=go1.16.9, user=root@34250ccfd885, date=20211019-16:51:42)"
level=info ts=2021-10-19T19:48:05.267Z caller=main.go:237 msg="Loaded config file"
level=info ts=2021-10-19T19:48:05.268Z caller=main.go:385 msg="Listening on address" address=:9115
level=info ts=2021-10-19T19:48:05.268Z caller=tls_config.go:191 msg="TLS is disabled." http2=false
```

### What did you see instead?

```
2021/10/19 19:29:27 http: panic serving 127.0.0.1:44030: runtime error: slice bounds out of range [:-9223372036854775808]
goroutine 11233 [running]:
net/http.(*conn).serve.func1(0xc00044b0e0)
  /usr/local/go/src/net/http/server.go:1804 +0x153
panic(0xb616a0, 0xc000036ff0)
[...]

PR #838, opened 2021-10-19
Fix body_size_limit negative handling

mem:
When body_size_limit is 0 (either explicitly or implicitly because it
wasn't specified) or less, we are setting it to math.MaxInt64. It turns
out that the implementation in http.MaxBytesReader tries to add 1 to the
specified value, and it wraps around. After that, it tries to use the
result to index an slice, causing it to panic.

Work around this by setting the limit to math.MaxInt64 - 1.

Also, if body_size_limit is exactly 0, leave it like that. That causes
the code to avoid setting up the limiter, saving some extra processing.

Closes: #837 

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -290,8 +290,14 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 
-	if s.BodySizeLimit <= 0 {
-		s.BodySizeLimit = math.MaxInt64
+	// BodySizeLimit == 0 means no limit. By leaving it at 0 we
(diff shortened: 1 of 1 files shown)


mem:
> Merge without review?

Sorry, I did not want to leave a panic in the code hanging around like that. I'm more than glad to make any changes.

mem:
For the record, I'm looking at why the existing tests did not catch this, as the path that was failing is the default one. I in particular missed it when testing by hand because I did not connect the dots in my head that when I changed `<` to `<=` in the original review I was changing the behavior for the default case, and I felt safe seeing that the tests were all passing.

Issue #474, opened 2019-05-07
Feature request: DNS probes should return enough metrics to tell servers that don't answer from servers that refuse our query


siebenmann:
Right now, DNS probes do not return enough information to tell the difference between a DNS server not responding at all and a DNS server that responds with query refusals (eg status REFUSED). These are potentially fairly different failures, especially if you're querying an external DNS server to verify that it can resolve your domains, and it would be nice if you could tell the difference through the blackbox exporter.

Perhaps the easiest feature addition would be to expose a probe_dns_rcode metric and give it a value of -1 if the probe received no response.

brian-brazil:
Wouldn't probe_dns_answer_rrs and whether the probe succeeded cover this?

siebenmann:
Unfortunately they don't, at least currently. If the probe fails because of either a REFUSED return code from the server or because the server does not respond, probe success is 0 and probe_dns_answer_rrs and the other _rrs metrics are 0, so you can't tell the two cases apart. Perhaps the probe_dns rrs metrics should have some other value if there's no reply at all, but 0 still feels natural in that case and I think it would be better to have a different metric to be explicit.

(If the server is down, you can try to look at the probe duration to guess if there was a timeout. But if the server is up but rejecting UDP DNS packets, for example because it's not currently running a DNS server, the probe duration is still short.)

candlerb:
Being able to set or clear RD in requests would be useful too (for correctly testing recursive and authoritative servers)

akshatgit:
If nobody is working on this, I'd like to contribute to this issue.

Issue #799, opened 2021-06-09
Metric for rcode returned from DNS probe


bennetty:
This is a feature request.

Currently the for the DNS probe, the metric `probe_success` returned 0/1 based on the success of the probe but i was wondering if it would be of value to return a metric which counts the number of each rcode for example?

this could give greater detail into the failed DNS probe without having to look through the logs.

Fo example:
rate(probe_dns_rcode{rcode="SERVFAIL"}[5m]) > 0.5
vs 
rate(probe_dns_rcode{rcode="REFUSED"}[5m]) > 0.5
could give better alerts

If this is a feature that is worth doing, i dont mind trying to implement it. Please let me know what you think :)

dgl:
This is covered by #474.

Note using labels would be nice for readability, but is also adding a lot of labels per probe, so I'm looking into doing this with a metric that has the value of the rcode instead.

PR #858, opened 2022-01-02
Add probe_dns_rcode metric

dgl:
This metric has the value of the response code from the DNS server; 0 is
NOERROR, per
https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6

It can be used along with probe_success in order to diagnose why
failures occurred.

For #474.

Signed-off-by: David Leadbeater <dgl@dgl.cx>

### DIFF
--- a/prober/dns.go
+++ b/prober/dns.go
@@ -142,6 +142,10 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_dns_additional_rrs",
 		Help: "Returns number of entries in the additional resource record list",
 	})
+	probeDNSRcodeGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_dns_rcode",
+		Help: "Returns the raw DNS rcode, 0 is NOERROR",
+	})
 
 	for _, lv := range []string{"resolve", "connect", "request"} {
 		probeDNSDurationGaugeVec.WithLabelValues(lv)
@@ -151,6 +155,7 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 	registry.MustRegister(probeDNSAnswerRRSGauge)
 	registry.MustRegister(probeDNSAuthorityRRSGauge)
 	registry.MustRegister(probeDNSAdditionalRRSGauge)
+	registry.MustRegister(probeDNSRcodeGauge)
 
 	qc := uint16(dns.ClassINET)
 	if module.DNS.QueryClass != "" {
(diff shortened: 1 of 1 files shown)


SuperQ (commented):
Hmm, I wonder if it'd be worth it to have this as a ENUM style metric.

```
probe_dns_rcode{rcode="NOERROR"} 1
probe_dns_rcode{rcode="SRVFAIL"} 0
probe_dns_rcode{rcode="NXDOMAIN"} 0
```

dgl:
It was mentioned in #799 -- it would be nice, but per https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6 there's quite a few possibilities, maybe we could pick the ones we think are likely...

Say NOERROR, SERVFAIL, NXDOMAIN, REFUSED and then have an OTHER for the rest would cover it, but we do then lose what the actual rcode is (unless we use the value of OTHER to be the rcode, but that is wrong for an ENUM style metric...)

dgl:
Adding some reviewers to get any thoughts on labels vs numbers...

mem:
> Adding some reviewers to get any thoughts on labels vs numbers...

Sorry, I was sure I had replied to this...

I would prefer @SuperQ's proposal. That makes it easier to build a counter out of this if desired (e.g. using a recording rule).

IIRC there are ~ 40 possible response codes. As long as the returned one is always captured, I don't have a strong preference for including a few select ones all the time.

SuperQ:
We could provide a list of default "always expodsed" rcodes as a command line flag. Something like `--probe.dns.rcodes=NOERROR,SRVFAIL,NXDOMAIN"`. If the rcode is some other code, we could either assign it to an `OTHER` bucket, or since it's reasonably limited, we could directly expose the "other" code.

Opinions?

mem:
> We could provide a list of default "always expodsed" rcodes as a command line flag. Something like `--probe.dns.rcodes=NOERROR,SRVFAIL,NXDOMAIN"`. If the rcode is some other code, we could either assign it to an `OTHER` bucket, or since it's reasonably limited, we could directly expose the "other" code.
> 
> Opinions?

I would be OK with that (either option: exposing the "new" code, or using an "OTHER" bucket, maybe leaning slightly towards the latter).

ArthurSens:
Hey hey 👋, it looks like #990 already solved the issue you're trying to solve here, unless the extra response details are really important.

I'm closing the PR since it's gotten quite stale, but please reopen it if you still think we should add the response details as a label!

PR #865, opened 2022-01-22
Delay init of http phase values

SuperQ:
Don't init http probe phase labels until we're ready to send the
request.

Avoids exposing 0 value samples for things we haven't started to
measure.

Fix http resolve timing to report lookup time even if it's an error
(ie not found).

Fixes: https://github.com/prometheus/blackbox_exporter/issues/579

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -307,10 +307,6 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		})
 	)
 
-	for _, lv := range []string{"resolve", "connect", "tls", "processing", "transfer"} {
-		durationGaugeVec.WithLabelValues(lv)
-	}
-
 	registry.MustRegister(durationGaugeVec)
 	registry.MustRegister(contentLengthGauge)
 	registry.MustRegister(bodyUncompressedLengthGauge)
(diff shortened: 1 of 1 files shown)


dgl (approved):

electron0zero (approved):

roidelapluie:
Mmmmh that's an interesting PR, it means that averages etc will only work when a phase is executed, I am not sure we should do this.

SuperQ:
@roidelapluie It's a difficult choice. The problem is we emit 0 for things we have explicitly not measured, which is what the original issue is about.

roidelapluie:
`Why this is a problem: if you perform any analytics over those metrics (eg percentiles, or min) it would be skewed.`

But if probe_success == 0 you should not run analytics over them either.

zerkms:
@roidelapluie 

> But if probe_success == 0 you should not run analytics over them either.

how would you exclude those series that were obtained with an error

roidelapluie:
I've been thinking this further.

When HAProxy Exporter can not scrape HAProxy, it does not return zero-ed metrics. We should do the same here.

roidelapluie (approved):

SuperQ:
I also added a fix for the "resolve" phase. It now accounts for the resolve phase time even if the resolve result is a failure.

roidelapluie:
Thanks!

PR #869, opened 2022-01-25
Negative maxTimeoutSeconds

### LINKED ISSUE #868: Negative maxTimeoutSeconds
Setting a small timeout value in a scraper (Prometheus) without changing the default offset leads to negative timeout values.
Prometheus configured to scrape every 200ms with default Blackbox offset of 500ms leads to timeout of -0.3s.

The possibility of a negative value of maxTimeoutSeconds is not checked here.
https://github.com/prometheus/blackbox_exporter/blob/f4a5173e9d4a722f0a8356aa60c961bd96b68e33/main.go#L445

zhdkirill:
Setting a small timeout value in a scraper (Prometheus) without changing the default offset leads to negative timeout values.
Fixes #868

Signed-off-by: Kirill Zhdanov <kzhdanov@mirantis.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -442,7 +442,7 @@ func getTimeout(r *http.Request, module config.Module, offset float64) (timeoutS
 	}
 
 	var maxTimeoutSeconds = timeoutSeconds - offset
-	if module.Timeout.Seconds() < maxTimeoutSeconds && module.Timeout.Seconds() > 0 {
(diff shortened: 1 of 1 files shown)


mem (approved):
Sorry about the delay, I missed this PR among a whole bunch of other stuff!

Thanks for the fix!

electron0zero (approved):

PR #884, opened 2022-02-22
Added log line for received html body.

pgmgb:
We had some troubles identifying why a specific HTTP probe with a `fail_if_body_matches_regexp` did not work for a specific website. In the end, the website started to deliver compressed data, so the regex was not working. For identifying such cases, printing the received HTML body would be handy.

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -50,6 +50,10 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
 		level.Error(logger).Log("msg", "Error reading HTTP body", "err", err)
 		return false
 	}
+
+	// Display received html body
+	level.Info(logger).Log("msg", "Received html body", "body", body)
+
 	for _, expression := range httpConfig.FailIfBodyMatchesRegexp {
 		if expression.Regexp.Match(body) {
 			level.Error(logger).Log("msg", "Body matched regular expression", "regexp", expression)


brian-brazil:
Historically I purposefully never logged this or returned headers as they often will contain secrets, plus it would be very spammy as it could be MB of data.

pgmgb:
Or would it be possible / an option to add another "debug" switch like `log_html_body`? It was quite a pain to debug this issue because the website was showing fine in the browser, but the check was not working...

roidelapluie:
Additionally to @brian-brazil's concerns about secrets, this would make the blackbox exporter an open proxy, creating security hazard for the majority of users.

pgmgb:
@brian-brazil & @roidelapluie
You would generally not add a possibility to debug the HTML body?
For me as a consumer of blackbox exporter, it would be handy to have this option.
So I could debug why something does not work as expected.

roidelapluie:
Would it be an option to have a command line flag to simulate a probe? so we do not expose the feature over the network.

Develoman:
are there any news about this MR?
In my case, i would like to check that an iframe is well loaded by verifing her html centent.
The URL of the iframe is well secured, i mean that securities aspect are maintained by the application of the iframe.

soasme:
@brian-brazil's concern totally makes sense. But the body of some ping target responses are simple text without secrets and in just a few bytes, such as "OK", "READY", "HEALTHY", etc. 

I would agree with @roidelapluie's opinion. It is great to have the ability letting end users decide whether to enable this feature.

To reduce the security impact to most users, this feature is turned off by default.

pgmgb:
@roidelapluie I would be happy with any debug option (plus documentation), which does not need any modification of the source code (which I had to do in my case) :thumbsup:

electron0zero:

This PR is quite old so I am going to close this, please feel free to reopen if you want to update it, and discuss it further.

PR #888, opened 2022-03-06
add OCSP

jrwren:


jrwren:
Apologies for not including unit tests with this. I took a look at the tests and since the TLS tests are around self signed CA, I'm not sure how OCSP could be tested. If anyone could advise, I'll gladly add tests.

prober/http.go
```diff
@@ -260,6 +260,11 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Help: "Indicates if SSL was used for the final redirect",
 		})
 
+		hasOCSPGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_http_ssl_has_ocsp",
```
mem, line 264: since it's possible to have multiple OCSP responses in a single TLS handshake, wouldn't it be better to name this `probe_http_ssl_ocsp_total`? So the value would be 0 if it's not present, instead of simply not having the metric?

mem:
> Apologies for not including unit tests with this. I took a look at the tests and since the TLS tests are around self signed CA, I'm not sure how OCSP could be tested. If anyone could advise, I'll gladly add tests.

I couldn't come up with a way to test this.

prober/http.go
jrwren: I don't know. I'm not very experienced with creating metrics. total of what? I think of this like a boolean value. 1 if there is OCSP, 0 if not. IMO it doesn't make sense to total them or add them up.

mem: Let me take a look at the code and figure out what it's able to report...

ArthurSens: I believe `probe_http_ssl_ocsp_total` would not be a good idea since `_total` suffixes are reserved for counters, following the [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/).

Could it be `probe_http_ssl_ocsp_stapled` instead? That indeed sounds like a boolean to me

jrwren:
I'm no longer interested in this feature.
https://letsencrypt.org/2024/12/05/ending-ocsp

ArthurSens (commented):
Hi, sorry for the super delayed review!

I'm reading [docs](https://pkg.go.dev/crypto/tls#ConnectionState:~:text=//%20OCSPStaple%20contains%20an%20optional%20OCSP%20response%20which%20will%20be%20served%0A%09//%20to%20clients%20that%20request%20it.%0A%09OCSPStaple%20%5B%5Dbyte) and thinking that we may be able to test this by passing a dummy OCSP staple

```go
certificate.OCSPStaple = []byte("dummy ocsp")
```

and then asserting on the exposed metric?

ArthurSens:
Lol, what a great timing hahahaha, that's alright! Thanks for your time anyway!

electron0zero:
> I'm no longer interested in this feature. [letsencrypt.org/2024/12/05/ending-ocsp](https://letsencrypt.org/2024/12/05/ending-ocsp)

cool, I think I merged a PR that added a flag to check CRLS, i think the flag is called check_revoked, it's on master, not released yet.

PR #903, opened 2022-04-21
fix CVE-2022-27191 in blackbox exporter service

pgvishnuram:
trivy scan result

image


SuperQ:
Fixed in https://github.com/prometheus/blackbox_exporter/pull/907

PR #919, opened 2022-05-30
Allow TTL to be explicitly set on ICMP packets from icmp prober

fhltang:
This adds plumbing to allow TTL to be explicitly set on ICMP packets from the icmp prober.

A specific use case is to allow me to detect when my broadband fails over to the backup ISP. I configure a module to send ICMP packets to a host (e.g. a DNS server) which is reachable in a short number of hops when connected via one ISP but not the other.


roidelapluie:
Can we set the default to 64?

fhltang:
> Can we set the default to 64?

I can but in its current form, this is the more conservative option. If TTL isn't set, it should be behave as it does now.

The code already uses a TTL of 64 in the raw connection code path. I'm not sure what default TTL is used otherwise.

fhltang:
TTL now defaults to 64 for ICMProbe. PTAL.

prober/icmp.go
```diff
@@ -234,15 +234,31 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
+				c4.SetTTL(ttl)
+			}
+			if c6 := icmpConn.IPv6PacketConn(); c6 != nil {
+				level.Debug(logger).Log("msg", "Setting TTL (IPv6 unprivileged)", "ttl", ttl)
+				c6.SetHopLimit(ttl)
+			}
+		}
 		_, err = icmpConn.WriteTo(wb, dst)
 	} else {
+		ttl := 64
```
electron0zero, line 250: bit uneasy with the magic numner here, can we use value of TTL from `config/config.go:82`.

config/config.go
```diff
@@ -405,6 +407,10 @@ func (s *ICMPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if runtime.GOOS == "windows" && s.DontFragment {
 		return errors.New("\"dont_fragment\" is not supported on windows platforms")
 	}
+
+	if s.TTL < 0 {
+		return errors.New("\"ttl\" cannot be negative")
+	}
```
mem, line 413: Can you add a check for `<= 255`?

CONFIGURATION.md
```diff
@@ -262,6 +262,11 @@ validate_additional_rrs:
 # The size of the payload.
 [ payload_size: <int> ]
 
+# TTL of outbound packets. Can be used to test reachability of a target
+# within a given number of hops, for example, to determine when network
+# routing has changed.
```
mem, line 267: Please add a note about the valid range being [0, 255], and that 0 means use the default value of 64... but that bit doesn't seem to be true. It looks like there's a path where the default is 64 and another one where it defaults to the current setting in the OS.

prober/icmp.go
```diff
@@ -234,15 +234,31 @@ func ProbeICMP(ctx context.Context, target string, module config.Module, registr
...
+			if c4 := icmpConn.IPv4PacketConn(); c4 != nil {
+				level.Debug(logger).Log("msg", "Setting TTL (IPv4 unprivileged)", "ttl", ttl)
+				c4.SetTTL(ttl)
+			}
+			if c6 := icmpConn.IPv6PacketConn(); c6 != nil {
+				level.Debug(logger).Log("msg", "Setting TTL (IPv6 unprivileged)", "ttl", ttl)
+				c6.SetHopLimit(ttl)
+			}
+		}
 		_, err = icmpConn.WriteTo(wb, dst)
```
mem, line 248: This one is the one that has me scratching my head...

What happens here if module.ICMP.TTL is zero? It uses the operating system's current setting. On Linux that's usually 64 (but it's distribution dependent). I have no idea what it is elsewhere.

mem (changes requested):

prober/icmp.go
fhltang: Done in commit 2393796

config/config.go
```diff
@@ -79,6 +79,7 @@ var (
 	// DefaultICMPProbe set default value for ICMPProbe
 	DefaultICMPProbe = ICMPProbe{
 		IPProtocolFallback: true,
+		TTL:                64,
```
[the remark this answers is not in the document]
fhltang: Done in commit 2393796

config/config.go
fhltang: Good idea. Done in commit 6c29e82.

CONFIGURATION.md
fhltang: I've included this in commit 6c29e82.

See next comment about discussion of using the OS's default versus a default baked into blackbox_exporter.

prober/icmp.go
fhltang: Yeah, this had me scratching my head too.

Unless I screwed up, this should preserve the existing behaviour:
- If we're using the unprivileged mode, it defaults to the OS's TTL, I believe, because it doesn't explicitly set the TTL: https://github.com/prometheus/blackbox_exporter/blob/f0d1414f1fa6fd0b0c4052e4dfae9b9e721060bd/prober/icmp.go#L237
- If we're using the IPv4 raw mode, it uses the default baked into the blackbox_exporter (i.e. 64): https://github.com/prometheus/blackbox_exporter/blob/f0d1414f1fa6fd0b0c4052e4dfae9b9e721060bd/prober/icmp.go#L245

I'm not sure if this difference in behaviour was by design. I am okay to make the changes in this PR so that the behaviour is more consistent, e.g. use the default value baked into blackbox_exporter.

With apologies, it's a new experience for me to contribute to this project so I'm not sure who can/should make these decisions (or even who can accept this PR). I am happy to make appropriate changes if someone points me in the right direction.

mem: That's alright.

Because of the way this is implemented, I doubt many people are relying on falling back to the OS setting.

Let's go forward with this as is.

mem (approved):

electron0zero (approved):
lgtm 👍🏼

Issue #922, opened 2022-06-06
Data race in ProbeHTTP


ozon2:
Hi, with blackbox_exporter 0.21.0 I get data races in `ProbeHTTP`, caused by the new `caser` used here https://github.com/prometheus/blackbox_exporter/blob/master/prober/http.go#L355.

```
==================
WARNING: DATA RACE
Write at 0x00c000468c80 by goroutine 28:
  golang.org/x/text/cases.(*titleCaser).Transform()
      /home/tpillot/dev/pkg/mod/golang.org/x/text@v0.3.7/cases/map.go:336 +0x144
  golang.org/x/text/transform.String()
      /home/tpillot/dev/pkg/mod/golang.org/x/text@v0.3.7/transform/transform.go:601 +0x1f8
  golang.org/x/text/cases.Caser.String()
      /home/tpillot/dev/pkg/mod/golang.org/x/text@v0.3.7/cases/cases.go:51 +0x8475
  github.com/prometheus/blackbox_exporter/prober.ProbeHTTP()
      /home/tpillot/dev/pkg/mod/github.com/prometheus/blackbox_exporter@v0.21.0/prober/http.go:355 +0x842b

Previous write at 0x00c000468c80 by goroutine 29:
  golang.org/x/text/cases.(*titleCaser).Transform()
      /home/tpillot/dev/pkg/mod/golang.org/x/text@v0.3.7/cases/map.go:336 +0x144
  golang.org/x/text/transform.String()
      /home/tpillot/dev/pkg/mod/golang.org/x/text@v0.3.7/transform/transform.go:601 +0x1f8
  golang.org/x/text/cases.Caser.String()
      /home/tpillot/dev/pkg/mod/golang.org/x/text@v0.3.7/cases/cases.go:51 +0x3075
  github.com/prometheus/blackbox_exporter/prober.ProbeHTTP()
      /home/tpillot/dev/pkg/mod/github.com/prometheus/blackbox_exporter@v0.21.0/prober/http.go:428 +0x302b
```

Here is an example to reproduce the issue:
```go
import (
	"sync"

	"github.com/go-kit/log"
	"github.com/prometheus/blackbox_exporter/config"
        "github.com/prometheus/blackbox_exporter/prober"
)

func TestBlackBox(t *testing.T) {
	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
	mod := config.DefaultModule
[...]

PR #927, opened 2022-06-13
Make caser a local varialbe, not a global one.

mem:
The cases.Caser returned by calling cases.Title *cannot* be shared among
goroutines. This might happen when Prometheus tries to scrape multiple
targets at the same time. From the docs:

A Caser may be stateful and should therefore not be shared between
goroutines.

Fixes: #922

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -46,8 +46,6 @@ import (
 	"github.com/prometheus/blackbox_exporter/config"
 )
 
-var caser = cases.Title(language.Und)
-
 func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logger log.Logger) bool {
 	body, err := ioutil.ReadAll(reader)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


electron0zero (approved):
lgtm :+1:

prober/http.go
```diff
@@ -342,6 +340,16 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		return false
 	}
 
+	// Do not move the following variable to global scope. The cases.Caser returned by
+	// calling cases.Title *cannot* be shared among goroutines. This might happen when
+	// Prometheus tries to scrape multiple targets at the same time. From the docs:
+	//
+	// A Caser may be stateful and should therefore not be shared between goroutines.
+	//
+	// Issue: https://github.com/prometheus/blackbox_exporter/issues/922
```
roidelapluie, line 349: I don't think it is necessary to point to the gh issue here

roidelapluie (approved):

prober/http.go
mem: I felt it was necessary to keep the reference close to the code because the original version is what I would have written, too. The wording in the docs, "a Caser may", sounds like if that were the actual case, the individual Caser implementations would spell that out. Given that in this particular instance there no such note, _I_ would have assumed that this Caser is not stateful.

Thanks for the review and merging this!

PR #935, opened 2022-06-22
Adding unix sockets support

issu-02:
Hello,
I'm currently using Blackbox exporter and I need to expose endpoints though Unix Sockets. This could be useful for many people because we can set file ACLs on Unix Sockets and so preventing unauthorized users from accessing the service, and it requires only adding few lines since UNIX socket support is a built-in in Golang. I have already tested the code, and it works well.

Signed-off-by: Ismail <benrakia.ismail@gmail.com>


roidelapluie:
This has been discussed before and we did not reach a consensus in Prometheus team whether we want it or not.

tiix95:
Bump this one. This will really help if this code is merged upstream. I complied it using the proposed patch and it's working well.

Why is this subject to discussion ? This does not implies any security issue, nor big code changes. This and can be really helpful, especially when working with apparmor.

Nginx team did that in their exporter (https://github.com/nginxinc/nginx-prometheus-exporter) and I managed to use it exactly the way I want.

I can fix failed checks if you're okay to merge this one. If you have more questions, I would be happy to answer them !

electron0zero:

This PR is quite old so I am going to close this, please feel free to reopen if you want to update it, and discuss it further.

PR #938, opened 2022-06-23
Use DialContext to deal with pre-resolving for HTTP 

dgl:
This replaces the method of resolving the IP and then munging the URL,
with essentially the other way around. It puts the desired hostname in
the URL, then uses a DialContext function to override the particular
host that is specified in target.

To try to explain that better, this behaves more like curl does when:
`curl --resolve example.com:443:192.0.2.1 http://example.com` is used.
That is, any request to the host specified (such as via redirects) will
continue to use the resolved IP.

This solves a very hard to debug problem -- if a target sends a
redirect, which has its own hostname in it, then blackbox_exporter will
look up that hostname via DNS and connect to it. This means the second
redirect could hit a different server.

An additional upside of this is the TLS server name logic is not needed
-- we are giving the Go HTTP client the real hostname, so it can check
SNIs for us (e.g. this solves some issues around port numbers in SNIs,
which blackbox_exporter didn't strip from the Host header).

Further, this deliberately makes no assumptions about other ports, so
for example a HTTP -> HTTPS redirect will result in a DNS resolution. It
also adds a log for the actual IP and port being connected to, making
debugging easier, and removing the confusing log message where the URL
is replaced with the IP.

Opening this as a pull request early for comments, there could be some more tests around some of this behaviour for sure (port handling at least, maybe HTTP/2). In general I feel this should remove a bunch of special cases and defer to Go's http client, which usually does the right thing. This maybe needs some thought around proxies, I think in general it will remove some weird edge cases, but I suspect someone is relying on those now...


prober/http.go
```diff
@@ -216,10 +219,13 @@ func (t *transport) TLSHandshakeStart() {
 	defer t.mu.Unlock()
 	t.current.tlsStart = time.Now()
 }
-func (t *transport) TLSHandshakeDone(_ tls.ConnectionState, _ error) {
+func (t *transport) TLSHandshakeDone(cs tls.ConnectionState, err error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	t.current.tlsDone = time.Now()
+	if err == nil {
+		level.Info(t.logger).Log("msg", "TLS handshake done", "alpn", cs.NegotiatedProtocol, "sni", cs.ServerName)
```
roidelapluie, line 227: Indo seems to be too much ?

prober/http.go
```diff
@@ -201,10 +203,11 @@ func (t *transport) ConnectDone(net, addr string, err error) {
 	defer t.mu.Unlock()
 	t.current.connectDone = time.Now()
 }
-func (t *transport) GotConn(_ httptrace.GotConnInfo) {
+func (t *transport) GotConn(conn httptrace.GotConnInfo) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	t.current.gotConn = time.Now()
+	level.Info(t.logger).Log("msg", "Connected to target", "addr", conn.Conn.RemoteAddr())
```
roidelapluie, line 210: debug?

prober/http.go
```diff
@@ -340,39 +349,43 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
-		// if targetHost is an IP address.
-		for name, value := range httpConfig.Headers {
-			if caser.String(name) == "Host" {
-				httpClientConfig.TLSConfig.ServerName = value
-			}
+	hostDialContext := func(ctx context.Context, network, address string) (net.Conn, error) {
+		var d net.Dialer
+		// If the address matches the address that the HTTP request specifies,
+		// ensure it goes to the IP already resolved from the target.
+		if network == "tcp" && address == net.JoinHostPort(requestHost, requestPort) {
```
roidelapluie, line 370: This part seems fragile to me, how well is it covered by existing tests?

prober/http.go
```diff
@@ -340,39 +349,43 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+			return d.DialContext(ctx, network, net.JoinHostPort(ip.String(), targetPort))
 		}
+		return d.DialContext(ctx, network, address)
```
roidelapluie, line 373: Maybe we want to log when we hit this case

roidelapluie:
How does this work with HTTP proxies?

jordant:
would like to see this get merged. subscribed

Issue #892, opened 2022-03-10
[RFE] Subject of TLS certificate and of Issuer certificate


lapo-luchini:
### Host operating system: output of `uname -a`
`FreeBSD $hostname 13.0-RELEASE-p4 FreeBSD 13.0-RELEASE-p4 #0: Tue Aug 24 07:33:27 UTC 2021     root@amd64-builder.daemonology.net:/usr/obj/usr/src/amd64.amd64/sys/GENERIC  amd64`

### blackbox_exporter version: output of `blackbox_exporter --version`
```
% pkg info -x blackbox
blackbox_exporter-0.17.0_1
% blackbox_exporter --version
blackbox_exporter, version  (branch: , revision: )
  build user:
  build date:
  go version:       go1.16.7
```

### What did you expect to see?
```
probe_ssl_earliest_cert_expiry 1.650999936e+09
probe_tls_version_info{version="TLS 1.3"} 1
probe_tls_certificate_info{subject="www.domain.test",issuer="R3",subjectalternative="*.domain.test,192.168.0.1,test.domain.test"} 1
```

### What did you see instead?
```
probe_ssl_earliest_cert_expiry 1.650999936e+09
probe_tls_version_info{version="TLS 1.3"} 1
```

electron0zero:
we build BBE on each merge in master, you can find most recent build at: https://app.circleci.com/pipelines/github/prometheus/blackbox_exporter/769/workflows/291b9405-ac0e-438c-9db2-64660660b57e/jobs/2531/artifacts

docker images for master builds can be found at: https://quay.io/repository/prometheus/blackbox-exporter?tab=tags&tag=master

another alternative is build locally: https://github.com/prometheus/blackbox_exporter#building-the-software

Till we get around cutting a new release, you can use these :)

PR #943, opened 2022-07-06
Added probe_tls_certificate_info metric with basic certificate details

djcode:
I tried implementing the RFE mentioned in #892 as I also need something like this for a project.


prober/tls.go
```diff
@@ -36,6 +37,21 @@ func getFingerprint(state *tls.ConnectionState) string {
...
+	return cert.Subject.CommonName
+}
+
+func getIssuer(state *tls.ConnectionState) string {
+	cert := state.PeerCertificates[0]
+	return cert.Issuer.CommonName
+}
+
+func getDNSNames(state *tls.ConnectionState) string {
+	cert := state.PeerCertificates[0]
```
electron0zero, line 51: if I understand correctly, `PeerCertificates[0]` will return leaf certificate.

do we want to only export details of leaf certificate?

djcode: Great question! The details of this certificate in my opinion are the most important, if we were to add others, that wil require rethink in how these values are extracted.

I reused the logic behind the "probe_ssl_last_chain_info" metric and in hindsight, that metric name is more succinct and applicable for the info the RFE requested so maybe it would be better to just add these labels there?

electron0zero: yes, `probe_ssl_last_chain_info` makes it explicit that this is info of `last` (i.e leaf) certificate with the metric name.

I think we can proceed with one of the following way:
1. keep the new metric, and update the PR to add details of all certificates in the chain.
2. rename the new metric to include `last` in the metric name, and only add last cert. info.
3. skip this new metric, and add the labels into `probe_ssl_last_chain_info` metric.

I am fine with any of the above solution.

@SuperQ @roidelapluie @mem what do you folks think?

djcode: I felt option 1 would need a pretty significant rewrite of the logic to iterate through all the certificates.
Option 2 might cause confusion for end users between the metric names so I opted to try option 3 and moved these labels into the `probe_ssl_last_chain_info` metric.
I also expanded the subject and issuer labels to return all their respective information, not just the CN of both.
Example

```
$ curl -s 'http://172.17.0.2:9115/probe?target=twitter.com' | grep probe_ssl_last_chain_info
# HELP probe_ssl_last_chain_info Contains SSL leaf certificate information
# TYPE probe_ssl_last_chain_info gauge
probe_ssl_last_chain_info{fingerprint_sha256="1d0020d1d0b632e78625219e9e2fdcf3c3a0f641240c4ecc46768cd0644884e3",issuer="CN=DigiCert TLS RSA SHA256 2020 CA1,O=DigiCert Inc,C=US",subject="CN=twitter.com,O=Twitter\\, Inc.,L=San Francisco,ST=California,C=US",subjectalternative="twitter.com,www.twitter.com"} 1
```

wwalters12:
@electron0zero @djcode Is there anything stopping this from being merged in? Would be great to see this feature get released!

electron0zero:
> @electron0zero @djcode Is there anything stopping this from being merged in? Would be great to see this feature get released!

merged,  @djcode thanks for the PR 🎉

PR #944, opened 2022-07-11
HTTP: Add Skip Resolve Phase With Proxy option

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@o11y.eu>

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -114,7 +114,7 @@ The other placeholders are specified separately.
   # HTTP proxy server to use to connect to the targets.
   [ proxy_url: <string> ]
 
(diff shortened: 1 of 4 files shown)


CONFIGURATION.md
```diff
@@ -114,6 +114,9 @@ The other placeholders are specified separately.
   # HTTP proxy server to use to connect to the targets.
   [ proxy_url: <string> ]
 
+  # Skip DNS resolution and URL change when an HTTP proxy is set.
```
electron0zero, line 117: nit: making it explicit that we are talking about `proxy_url` config option here.

```suggestion
  # Skip DNS resolution and URL change when an HTTP proxy (proxy_url) is set.
```

electron0zero (approved):

PR #947, opened 2022-07-14
Add hostname param support for tls servername

BoringCat:
### Modified for this case
- need to monitor multiple domain which is dynamic resolve to different instance
- instance handle multiple certificate and distinguish with SNI

To monitor more than thousands domain. I can not just create config for prometheus and blackbox-exporter dynamiced. If each domain need it own `job` and `module`. The config file will to large and may be affect prometheus performance.

So it will be perfect if single `module` can handle all request just with hostname param like http probe 👍

### DIFF
--- a/main.go
+++ b/main.go
@@ -126,11 +126,20 @@ func probeHandler(w http.ResponseWriter, r *http.Request, c *config.Config, logg
 	}
 
 	hostname := params.Get("hostname")
-	if module.Prober == "http" && hostname != "" {
-		err = setHTTPHost(hostname, &module)
-		if err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
-			return
+	if hostname != "" {
+		switch module.Prober {
+		case "http":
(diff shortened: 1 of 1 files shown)


roidelapluie:
Thanks. This would require tests.

electron0zero:
fyi: https://github.com/prometheus/blackbox_exporter/pull/981 is also fixing the same issue.

BoringCat:
> fyi: #981 is also fixing the same issue.

Well, after i make this PR. My target was changed. I need to known more infomation about the certificate. I'm now creating my own exporter. So I have no time work for this. If #981 can fix it and pass test. Please close this.

electron0zero:
> > fyi: #981 is also fixing the same issue.
> 
> Well, after i make this PR. My target was changed. I need to known more infomation about the certificate. I'm now creating my own exporter. So I have no time work for this. If #981 can fix it and pass test. Please close this.

closing in favor of #981

PR #952, opened 2022-07-27
Remove ioutil

inosato:
`ioutil` is deprecated from go1.16 (ref: https://go.dev/doc/go1.16#ioutil).
So removed them.

Signed-off-by: inosato <si17_21@yahoo.co.jp>


SuperQ (approved):
Again, thanks for all these cleanups.

PR #957, opened 2022-08-09
Refactor proberHandler as an exported function

marctc:
We are in the process to add blackbox exporter in grafana agent
as integration (https://github.com/grafana/agent/pull/1997). With the current
layout is not possible to reuse any of the code that makes the exporter run.

This PR refactors `probeHandler` to be an exporter function of the `probe`
package. There are not code changes, just moved things around.


marctc:
@mem @roidelapluie friendly nudge to get a review here :pray:

mem (approved):
Thanks for this. I have done this myself, so I know it's possible without modifications to the existing code, but I also know it's very brittle. This way is cleaner and I think it's a valid use case.

I went thru the changes, and I couldn't find any change in behavior. I tested it and it's behaving the same way, so I think we are good.

roidelapluie (approved):

marctc:
Thanks for the review @mem @roidelapluie! Can I get this merged? Thanks!

PR #961, opened 2022-08-24
add boilerplate code for running as a windows service

terra-nova:
this PR adds support for running blackbox_exporter as a Windows service

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,6 +46,8 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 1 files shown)


SuperQ:
Doing this as a drive-by PR to one exporter is not the way to do this. Please start with a proposal to the [Prometheus Developers community](https://prometheus.io/community).

If we decide we want to support this kind of boilerplate, it would need to be part of the [exporter-toolkit](https://github.com/prometheus/exporter-toolkit) first.

ArthurSens:
Closing this PR due to inactivity.

PR #970, opened 2022-09-20
consistent help message of gauge metrics across different probes (http/tcp)

mvnarsing:
We are running otelcollector and blackbox exporter probes (http/tcp) on same instances and noticing gaps in metrics, after analyzing logs we think it might be due to inconsistent HELP message for the metrics listed in this PR.

Screen Shot 2022-09-20 at 4 46 24 PM

Here is the snippet of errors in logs:
```
label:<name:\"type\" value:\"htp\" > gauge:<value:1.6e+09 >  has help \"Returns earliest SSL cert expiry date\" but should have \"Returns earliest SSL cert expiry in unixtime\"\n*
1label:<name:\"version\" value:\"TLS 1.3\" > gauge:<value:1 >  has help \"Returns the TLS version used, or NaN when unknown\" but should have \"Contains the TLS version used\"\n* collected metric
```

Additional details on the check that is failing in prometheus:
the source code of the **otelcollector prometheusexporter** ((https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/prometheus.go#L24) ) that we are using for scraping metrics, is using client_golang/prometheus/promhttp, which is the prometheus go client library.
check that is failing:
https://github.com/prometheus/client_golang/blob/83d56b1144a0c2eb10d399e7abbae3333bebc463/prometheus/registry.go#L953


prober/tcp.go
```diff
@@ -91,11 +91,11 @@ func dialTCP(ctx context.Context, target string, module config.Module, registry
 func ProbeTCP(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) bool {
 	probeSSLEarliestCertExpiry := prometheus.NewGauge(prometheus.GaugeOpts{
 		Name: "probe_ssl_earliest_cert_expiry",
-		Help: "Returns earliest SSL cert expiry date",
+		Help: "Returns earliest SSL cert expiry in unixtime",
```
SuperQ, line 94: If we want these to be consistent, we should move these help strings to `const` values in `prober.go`.

Or maybe move the whole `prometheus.GaugeOpts{}` to be a package level var.

mvnarsing: pushed new commit with `const` values in `prober.go`

electron0zero (approved):

prober/prober.go
```diff
@@ -23,3 +23,9 @@ import (
 )
 
 type ProbeFn func(ctx context.Context, target string, config config.Module, registry *prometheus.Registry, logger log.Logger) bool
+
+const (
+	helpSSLEarliestCertExpiry     = "Returns last SSL chain expiry in unixtime"
+	helpSSLChainExpiryInTimeStamp = "Returns last SSL chain expiry in timestamp"
+	helpProbeTLSInfo              = "Contains the TLS version used"
```
electron0zero, line 30: nit: I am leaning on the side of using `Returns the TLS version used, or NaN when unknown` for the helpProbeTLSInfo. it's bit more explict. 

feel free to ignore this if you feel otherwise.

electron0zero, line 30:
```suggestion
	helpProbeTLSInfo              = "Returns the TLS version used or NaN when unknown"
```

electron0zero:
@mvnarsing thank you for the contribution

mvnarsing:
when can we expect next release?

Issue #695, opened 2020-09-29
Installation on Windows 


herland:
I created the service with the following parameters:

C:\Data\blackbox_exporter\blackbox_exporter.exe --config.file=C:\Data\blackbox_exporter\blackbox.yml 

I have the following error at startup:: The service did not respond to the start or control request in a timely fashion.

has someone installed the service on windows?

brian-brazil:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

herland:
grazie mile

Issue #559, opened 2019-12-16
How do I start up in the background in Windows


for-mat:
I created the service on Windows, but the startup times were wrong
error 1053 The service did not respond to startup or control requests in a timely manner

brian-brazil:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

nadme92:
> I created the service on Windows, but the startup times were wrong
> error 1053 The service did not respond to startup or control requests in a timely manner

I have exactly the same problem have you find a solution ? thanks

PR #971, opened 2022-09-27
chore: wrap blackbox exporter with kardianos/service to support Windows controls

locmai:
## Intent

An attempt to wrap the the blackbox exporter with kardianos/service to be able to run as a Windows background service while still be able to run as a normal Linux service.

I'm open for change request and suggestions!

## Issues

A few issues opened in the past:
- https://github.com/prometheus/blackbox_exporter/issues/695
- https://github.com/prometheus/blackbox_exporter/issues/559

A similar (intent) PR I found here using `golang.org/x/sys/windows/svc` (underlying `kardianos/service` is using the same).
- https://github.com/prometheus/blackbox_exporter/pull/961

## Prometheus Users group

Also opened up a discussion here as well https://groups.google.com/g/prometheus-users/c/p1DUqyqT05o


locmai:
> How does [windows_exporter](https://github.com/prometheus-community/windows_exporter) deal with this problem? It seems to wrap it in an .msi installer.

Hmm let me take a look at the windoes_exporter to see their solution. We tried a different MSI packager but it ended up with the following error:

> error 1053 The service did not respond to startup or control requests in a timely manner

locmai:
@candlerb I just took a look at the windows_exporter and figured out that we approach the similar way to handle the windows service control https://github.com/prometheus-community/windows_exporter/blob/677a7c8d67deb99b92f4f24b8c890e0a4c152b0c/exporter.go#L7-L9 - from that we have the https://github.com/prometheus-community/windows_exporter/blob/677a7c8d67deb99b92f4f24b8c890e0a4c152b0c/initiate/initiate.go#L8  `golang.org/x/sys/windows/svc` usage.

This seems like the right way to make it.

> I think it should be opt-in

And I also could make a flag to opt-in to use that.

candlerb:
I guess windows_exporter is only ever going to run under Windows, so it's reasonable for it to code directly to Windows service APIs.

I wonder if there is a more general solution: some sort of wrapper which can run as a windows service and in turn executes the application in question?  A quick google turns up https://github.com/winsw/winsw - could that be used to run blackbox_exporter?

PR #979, opened 2022-10-21
Update to exporter-toolkit v0.8.1

DaAwesomeP:
This pull updates the blackbox_exporter to use the latest version of exporter-toolkit and includes the new flag methods.

Associated exporter-toolkit pull: https://github.com/prometheus/exporter-toolkit/pull/95

Similar node_exporter pull: https://github.com/prometheus/node_exporter/pull/2393

cc @SuperQ @roidelapluie


DaAwesomeP:
Build is failing because `go mod tidy` prompts if Go 1.16 compatibility is required (this is reproduceable on local machines). Running `go mod tidy -compat=1.17` does not prompt and passes. [client_golang v1.13.0](https://github.com/prometheus/client_golang/releases/tag/v1.13.0) requires a minimum of Go 1.17.

Please advise on the preferred solution.

SuperQ:
Bumping the indirect dependencies is the preferred method. For example `go get golang.org/x/xerrors`.

SuperQ:
Maybe we should move the external URL and route prefix features to the toolkit next.

DaAwesomeP:
Fixed dependency issues! Now passes tests.

> Maybe we should move the external URL and route prefix features to the toolkit next.

Yeah, I can look into that next possibly. Though no promises as my non-work open source timeline is often sporadic. The route prefix option seems pretty popular, though I guess in my limited experience so far I have not encountered external URL as much.

I think if we were to formalize getting data out of the flags then we should probably implement helper methods instead of encouraging struct access to avoid possible later breaking changes. This is more a discussion for the toolkit repo so maybe I will open an issue there.

For v0.8.0 though, this should be good to go (pun intended)!

go.mod
```diff
@@ -6,12 +6,12 @@ require (
...
 	github.com/miekg/dns v1.1.50
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.12.2
+	github.com/prometheus/client_golang v1.13.0
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.37.0
-	github.com/prometheus/exporter-toolkit v0.7.1
-	golang.org/x/net v0.0.0-20220728211354-c7608f3a8462
-	google.golang.org/grpc v1.48.0
+	github.com/prometheus/exporter-toolkit v0.8.0
```
SuperQ, line 12: We need to bump this to `v0.8.1` to pick up small bug fix.

https://github.com/prometheus/exporter-toolkit/pull/118

SuperQ (changes requested):
Need to pull in the bug fix, otherwise LGTM.

DaAwesomeP:
Bumped exporter-toolkit to v0.8.1!

DaAwesomeP:
@SuperQ just checking in on this!

main.go
```diff
@@ -92,7 +91,14 @@ func run() int {
 	level.Info(logger).Log("msg", "Loaded config file")
 
 	// Infer or set Blackbox exporter externalURL
-	beURL, err := computeExternalURL(*externalURL, *listenAddress)
+	listenAddrs := toolkitFlags.WebListenAddresses
+	if *externalURL == "" && *toolkitFlags.WebSystemdSocket {
```
SuperQ, line 95: Shouldn't this be `*externalURL != ""`

DaAwesomeP: I believe it is correct, and I tested it. This if statement intentionally throws an error if an external URL is not explicitly provided when using a systemd socket listener. This is because the external URL can never be inferred from the systemd socket.

SuperQ: But, what if you don't care about the external URL? It's an optional setting in the normal case. Wouldn't this for the user to set an external URL in systemd mode? I don't think that's desired.

DaAwesomeP: External URL is an optional setting because it infers it from the port number or hostname provided when no external URL is provided. The default from the optional flag is an empty string, but then the function called on it later finds the real fallback value.

If you're saying the external URL is not actually needed for the exporter to function, then I think that goes past the scope of this pull.

roidelapluie (approved):

roidelapluie:
Thanks!

PR #823, opened 2021-09-11
Add hostname parameter

anemyte:
This is refactoring of https://github.com/prometheus/blackbox_exporter/pull/816
As per [the design doc](https://docs.google.com/document/d/1VwqXi2TOb5KXaZY6Iio7411x64pJao3GusX8MqYsJ2g), the exporter now accept `hostname` parameter, which sets `Host` header for HTTP probes. It also sets TLS server name (if it is not already defined by module configuration) to avoid TLS handshake problems when `target` is an IP address. Setting `Host` with both module configuration and `hostname` parameter results in 400 Bad Request.

PR #981, opened 2022-10-25
Add hostname parameter for TCP probe

lspiehler:
This adds the ability to set the TLS server name for TCP probes using the hostname parameter just like #823 did for HTTP probes. It was mentioned by @Nothing4You in #624, and its something I need for monitoring in my environment as well.


prober/handler.go
```diff
@@ -107,6 +107,14 @@ func Handler(w http.ResponseWriter, r *http.Request, c *config.Config, logger lo
 		}
 	}
 
+	if module.Prober == "tcp" && hostname != "" {
```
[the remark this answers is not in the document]
lspiehler: Test has been created. Thanks!

electron0zero (approved):
lgtm :+1:

electron0zero:
looking for comments from @mem @roidelapluie, will merge if there are no comments :)

gitanuj:
This seems like a missing feature for tcp probes (since http probes already expose hostname). Would love to see this ship soon!

prober/handler.go
```diff
@@ -137,6 +145,15 @@ func Handler(w http.ResponseWriter, r *http.Request, c *config.Config, logger lo
 	h.ServeHTTP(w, r)
 }
 
+func setTLSServerName(hostname string, module *config.Module) error {
```
electron0zero, line 148: error is always nil? so doesn't make sense to return error in this case.

also, can we not inline this assignment?

lspiehler: I removed the function and made the assignment inline in d207f64. Thanks!

electron0zero (approved):

electron0zero:
merged, thanks @lspiehler :)

PR #982, opened 2022-10-26
Add ip_filter support to block certain IP addresses

### LINKED ISSUE #905: Allow blocking of specified IP ranges
We are using blackbox exporter as a library, in a kubernetes cluster, and some of our targets are specified by other teams, or even our customers. We would like to ensure there is no way that probes can interact with internal cluster components, or certain sensitive services in this network. 

Firewall rules, NetworkPolicy, and so forth have been explored extensively, and have proven difficult if not impossible to implement the levels of granularity we need. The solution we would like to explore next is implementing this blocking in-app, by having bbe check a blocklist before dialing out. It needs to be done after dns resolution, so we can't simply filter our targets before calling the probers.

I have implemented this [in my fork](https://github.com/prometheus/blackbox_exporter/compare/master...captncraig:cmp_dialer), by providing a custom dialer function that can be overridden by a user of the library. This would allow us to inject a custom dialer that respects the cidr blocks we want to deny any access to.

Is this something that this project would consider accepting? I am not sure if this is the best way to go about it, or if it would make sense to put full IP blocking functionality into blackbox exporter itself.

adriansr:
IPFilter is used to filter network traffic so that probes are limited in the destination addresses they can probe.

The configuration looks like:
```
      ip_filter:
               default: allow|deny
               allowed:
                 [... list of networks in CIDR format ...]
               denied:
                 [... list of networks in CIDR format ...]
```

This can be specified at the top level or inside each module definition. All keys are optional: an empty undefined `ip_filter` allows all traffic. When just `allowed` or `denied` is defined, `default` takes the implicit reasonable value:

```
    ip_filter:
      # implicit default: allow
      denied:
        - 127.0.0.0/8
        - 10.0.0.0/8
        - 192.168.0.0/16
        - ::1/128
        - fd00::/8
        - fc00::/7    
```

The longest prefix match takes precedence:

```
    ip_filter:
      default: deny
      allowed:
        - 10.0.0.0/8
      denied:
        - 10.0.1.0/24
```

Only traffic to 10.0.0.0/8 is allowed, except the 10.0.1.0/24 subnet which is blocked.

Closes #905

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -95,7 +95,8 @@ func init() {
 }
 
 type Config struct {
-	Modules map[string]Module `yaml:"modules"`
+	Modules  map[string]Module `yaml:"modules"`
+	IPFilter *IPFilter         `yaml:"ip_filter,omitempty"`
 }
 
 type SafeConfig struct {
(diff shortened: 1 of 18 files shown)


roidelapluie:
Thanks.

I think this would requires a design document first, do address several questions:

- Can we do it low-level, at the net.Conn level?
- Do we need both allow & block list? Can we deal with just allow list?
- How does this work with hostnames? Do we hard fail if the first resolved addresses is not allowed or do we iterate?
- What do do with http proxies and skip resolution?
- How do we make a standard logic that applies to every probe?

roidelapluie:
one of the issue with the current approach is that we do not verify redirects.

adriansr:
> one of the issue with the current approach is that we do not verify redirects.

Can you explain this one with a little bit more detail? not sure I understand what you mean.

roidelapluie:
If http://127.0.0.1 replies with a 302 redirect to http://192.168.10.200, we would follow that http redirect even if it is not in the whitelist

adriansr:
@roidelapluie unless I'm missing something, redirects are taken into account in this implementation. There's even a test for it:
https://github.com/prometheus/blackbox_exporter/blob/9ad49c07fa4dbc353663ce6dfcf46438d0b55b4c/prober/http_test.go#L1289

PR #990, opened 2022-11-25
Add probe_dns_query_succeeded metric

### LINKED ISSUE #474: Feature request: DNS probes should return enough metrics to tell servers that don't answer from servers that refuse our query

danteu:
Currently, one may not be able to differentiate between a query refusal
by the target and the target not responding to a query. These two cases
can have very different ramifications.

This patch adds the probe_dns_query_succeeded gauge which is 1 if and
only if the DNS query was executed correctly, i.e., that the target
host has sent a response, and 0 otherwise.

Resolves #474

Signed-off-by: Daniel Teunis <daniel@teunis.cc>

### DIFF
--- a/prober/dns.go
+++ b/prober/dns.go
@@ -142,6 +142,10 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_dns_additional_rrs",
 		Help: "Returns number of entries in the additional resource record list",
 	})
+	probeDNSQuerySucceeded := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_dns_query_succeeded",
+		Help: "Displays whether or not the query was executed successfully",
+	})
 
 	for _, lv := range []string{"resolve", "connect", "request"} {
 		probeDNSDurationGaugeVec.WithLabelValues(lv)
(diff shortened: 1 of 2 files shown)


prober/dns.go
```diff
@@ -142,6 +142,10 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_dns_additional_rrs",
 		Help: "Returns number of entries in the additional resource record list",
 	})
+	probeDNSTargetReachable := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_dns_target_reachable",
```
SuperQ, line 146: Based on the issue, and what this represents, this isn't about the target being reachable. It's more about the query being answered or refused.

How about:
```suggestion
		Name: "probe_dns_query_succeeded",
```

SuperQ (changes requested):
The metric name / help doesn't seem to match what this is really about.

prober/dns.go
danteu: Thanks for the quick reply! 'probe_dns_query_succeeded' works for me.

SuperQ (approved):

PR #1009, opened 2022-12-21
Add root (`modules`) key to CONFIGURATION.md

DazWilkin:
The Exporter configuration is complicated by the omission of the root (`modules`) key from the configuration document.

Not only must configuration begin with `modules` but each module is a map of module names to module values.

I think I've correctly reflected this here using the document's schema.

Signed-off-by: Daz Wilkin <DazWilkin@users.noreply.github.com>

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -16,6 +16,8 @@ Generic placeholders are defined as follows:
 
 The other placeholders are specified separately.
 
+See [example.yml](example.yml) for configuration examples.
+
 ```yml
 
 modules:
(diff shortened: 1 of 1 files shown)


CONFIGURATION.md
```diff
@@ -16,7 +16,15 @@ Generic placeholders are defined as follows:
 
 The other placeholders are specified separately.
```
electron0zero, line 17: unrelated to this change but let's link to `example.yml` in this file so users can find it easily :)

```suggestion
The other placeholders are specified separately.

See [example.yml](example.yml) for configuration examples.
```

electron0zero (approved):

PR #1056, opened 2023-04-05
Remove initialization of metrics from init functions

marctc:
When blackbox exporter is imported as package (i.e in Grafana Agent) is exporting metrics event if the exporter is not enabled (see https://github.com/grafana/agent/issues/2627).

This PR moves the initialization of some of the internal metrics of the exporter out of `init` functions.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -34,17 +34,18 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/miekg/dns"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/config"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


SuperQ:
Maybe this would be better done with promauto?

config/config.go
```diff
@@ -103,6 +99,12 @@ type SafeConfig struct {
 	C *Config
 }
 
+func NewSafeConfig() *SafeConfig {
+	prometheus.Register(configReloadSuccess)
+	prometheus.Register(configReloadSeconds)
+	return &SafeConfig{C: &Config{}}
+}
```
SuperQ, line 106: With `promauto`, this shouldn't be necessary.

Also, if you're using `prometheus.Register()`, you need to check the error return.

SuperQ (approved):

electron0zero (approved):

PR #1059, opened 2023-04-14
http: Prevent logging confusing error message

adriansr:
When FailIfNotSSL is set, the error message:

> Final request was not over SSL

is logged even if a previous error has prevented the request to complete, for example when it fails due to a timeout or connection error.

This updates the logic around the log message so it's printed when an otherwise successful request is not performed over SSL and that causes the probe to fail.


prober/http.go
```diff
@@ -647,7 +647,7 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			level.Error(logger).Log("msg", "Final request was over SSL")
 			success = false
 		}
-	} else if httpConfig.FailIfNotSSL {
+	} else if httpConfig.FailIfNotSSL && success {
```
electron0zero, line 650: can we add a test case for it please :)

adriansr: @electron0zero can you check again please?

electron0zero: lgtm, approved.

PR #1060, opened 2023-04-19
Explicit registration of internal exporter metrics

marctc:
- The first attempt to de-register internal metrics in #1056 was not sucessful.
- Added a `NewSafeConfig` function that gets a `prometheus.Registerer` as parameter and register config reloading metrics.
- Moved `moduleUnknownCounter` again to main and now it's passed (optionally) to `Handler` function.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -34,18 +34,17 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/miekg/dns"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/config"
 )
 
 var (
-	configReloadSuccess = promauto.NewGauge(prometheus.GaugeOpts{
+	configReloadSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: "blackbox_exporter",
 		Name:      "config_last_reload_successful",
 		Help:      "Blackbox exporter config loaded successfully.",
 	})
 
-	configReloadSeconds = promauto.NewGauge(prometheus.GaugeOpts{
+	configReloadSeconds = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: "blackbox_exporter",
 		Name:      "config_last_reload_success_timestamp_seconds",
 		Help:      "Timestamp of the last successful configuration reload.",
@@ -97,6 +96,12 @@ type Config struct {
 type SafeConfig struct {
 	sync.RWMutex
 	C *Config
+}
+
+func NewSafeConfig(reg prometheus.Registerer) *SafeConfig {
+	reg.MustRegister(configReloadSuccess)
+	reg.MustRegister(configReloadSeconds)
+	return &SafeConfig{C: &Config{}}
 }
 
 func (sc *SafeConfig) ReloadConfig(confFile string, logger log.Logger) (err error) {
--- a/config/config.go
+++ b/config/config.go
@@ -38,18 +38,6 @@ import (
 )
 
 var (
-	configReloadSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
-		Namespace: "blackbox_exporter",
-		Name:      "config_last_reload_successful",
-		Help:      "Blackbox exporter config loaded successfully.",
-	})
-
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -34,18 +34,17 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/miekg/dns"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/config"
 )
 
 var (
-	configReloadSuccess = promauto.NewGauge(prometheus.GaugeOpts{
+	configReloadSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
```
SuperQ, line 41: Rather than switch away from promauto, we should move these vars inside the `SafeConfig` struct. Then we can create them in `NewSafeConfig()`.

marctc: I moved to `NewSafeConfig` but we can't use `promauto` as it always uses the default registerer (`config_test.go` panics due to duplicated registration)

SuperQ: You can use promauto with an explicit registry.

config/config.go
```diff
@@ -96,17 +83,36 @@ type Config struct {
...
 type SafeConfig struct {
 	sync.RWMutex
-	C *Config
+	C                   *Config
+	configReloadSuccess prometheus.Gauge
+	configReloadSeconds prometheus.Gauge
+}
+
+func NewSafeConfig(reg prometheus.Registerer) *SafeConfig {
+	configReloadSuccess := prometheus.NewGauge(prometheus.GaugeOpts{
```
SuperQ, line 92:
```suggestion
	configReloadSuccess := promauto.With(reg).NewGauge(prometheus.GaugeOpts{
```

main.go
```diff
@@ -55,10 +53,16 @@ var (
 	externalURL   = kingpin.Flag("web.external-url", "The URL under which Blackbox exporter is externally reachable (for example, if Blackbox exporter is served via a reverse proxy). Used for generating relative and absolute links back to Blackbox exporter itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Blackbox exporter. If omitted, relevant URL components will be derived automatically.").PlaceHolder("<url>").String()
 	routePrefix   = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").PlaceHolder("<path>").String()
 	toolkitFlags  = webflag.AddFlags(kingpin.CommandLine, ":9115")
+
+	moduleUnknownCounter = prometheus.NewCounter(prometheus.CounterOpts{
```
SuperQ, line 57: Similar here, we can use `promauto`, since this is part of our main.

main.go
```diff
@@ -178,7 +182,7 @@ func run() int {
 		sc.Lock()
 		conf := sc.C
 		sc.Unlock()
-		prober.Handler(w, r, conf, logger, rh, *timeoutOffset, nil)
+		prober.Handler(w, r, conf, logger, rh, *timeoutOffset, nil, &moduleUnknownCounter)
```
SuperQ, line 185: `prometheus.Counter` is an interface, so I'm pretty sure we don't need to pass a reference.

CC @roidelapluie

prober/handler.go
```diff
@@ -60,7 +56,9 @@ func Handler(w http.ResponseWriter, r *http.Request, c *config.Config, logger lo
 	if !ok {
 		http.Error(w, fmt.Sprintf("Unknown module %q", moduleName), http.StatusBadRequest)
 		level.Debug(logger).Log("msg", "Unknown module", "module", moduleName)
-		moduleUnknownCounter.Add(1)
+		if moduleUnknownCounter != nil {
```
SuperQ, line 59: I kinda feel like we should enforce that a non-nil `prometheus.Counter` is passed to this.

SuperQ (changes requested):

prober/handler.go
marctc: I did this way because we don't need it in the agent...but I can change it if is problematic.

SuperQ (approved):
LGTM

prober/handler.go
SuperQ: Makes sense.

PR #1066, opened 2023-05-10
Build blackbox_exporter with go 1.19.9 

electron0zero:
Go came out with new secruity release with [fixes for bunch of CVEs](https://go.dev/doc/devel/release#go1.20), Update go to build blackbox_exporter.

Also see #1070

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ executors:
   # Whenever the Go version is updated here, .promu.yml should also be updated.
   golang:
     docker:
-      - image: cimg/go:1.19
+      - image: cimg/go:1.19.9
 jobs:
   test:
     executor: golang
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/prometheus/blackbox_exporter
 
-go 1.18
+go 1.19
 
 require (
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -1,44 +1,46 @@
 module github.com/prometheus/blackbox_exporter
 
-go 1.18
+go 1.20
```
SuperQ, line 3: We only update the Go minimum version if required.

go.mod
```diff
@@ -1,44 +1,46 @@
...
-	github.com/miekg/dns v1.1.50
+	github.com/miekg/dns v1.1.54
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.14.0
-	github.com/prometheus/client_model v0.3.0
-	github.com/prometheus/common v0.39.0
-	github.com/prometheus/exporter-toolkit v0.8.2
-	golang.org/x/net v0.7.0
-	google.golang.org/grpc v1.52.3
+	github.com/prometheus/client_golang v1.15.1
```
SuperQ, line 11: We typically let dependabot take care of these things.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ executors:
   # Whenever the Go version is updated here, .promu.yml should also be updated.
   golang:
     docker:
-      - image: cimg/go:1.19
+      - image: cimg/go:1.19.9
```
SuperQ, line 8: We don't ever specify a specific version here, this tag always uses the latest point release. No need to change it.

SuperQ (changes requested):
These changes don't actually do anything, as the final binary is produced with the Prometheus `golang-builder` docker image. This is controlled via promu.

electron0zero:
closing this in favor of https://github.com/prometheus/blackbox_exporter/pull/1072

PR #1091, opened 2023-07-01
Log errors from probe

jkroepke:
This PR adds a flag `--log.prober` which can be one of: [debug, info, warn, error, none] to investigate probe failures port-mortem. this flag defaults to none.

At the moment, this analyse problems is only possible via Web UI with an large enough history. Access the unprotected web ui can be difficult and the probe history is only available until restart.

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,7 +50,7 @@ var (
 	configFile     = kingpin.Flag("config.file", "Blackbox exporter configuration file.").Default("blackbox.yml").String()
 	timeoutOffset  = kingpin.Flag("timeout-offset", "Offset to subtract from timeout in seconds.").Default("0.5").Float64()
 	configCheck    = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
-	logProbeErrors = kingpin.Flag("log.probe-errors", "Log error from probe requests").Default().Bool()
+	logLevelProber = kingpin.Flag("log.prober", "Log level from probe requests. One of: [debug, info, warn, error, none]").Default("none").String()
 	historyLimit   = kingpin.Flag("history.limit", "The maximum amount of items to keep in the history.").Default("100").Uint()
 	externalURL    = kingpin.Flag("web.external-url", "The URL under which Blackbox exporter is externally reachable (for example, if Blackbox exporter is served via a reverse proxy). Used for generating relative and absolute links back to Blackbox exporter itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Blackbox exporter. If omitted, relevant URL components will be derived automatically.").PlaceHolder("<url>").String()
 	routePrefix    = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").PlaceHolder("<path>").String()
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -47,13 +47,14 @@ import (
...
-	externalURL   = kingpin.Flag("web.external-url", "The URL under which Blackbox exporter is externally reachable (for example, if Blackbox exporter is served via a reverse proxy). Used for generating relative and absolute links back to Blackbox exporter itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Blackbox exporter. If omitted, relevant URL components will be derived automatically.").PlaceHolder("<url>").String()
-	routePrefix   = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").PlaceHolder("<path>").String()
-	toolkitFlags  = webflag.AddFlags(kingpin.CommandLine, ":9115")
+	configFile     = kingpin.Flag("config.file", "Blackbox exporter configuration file.").Default("blackbox.yml").String()
+	timeoutOffset  = kingpin.Flag("timeout-offset", "Offset to subtract from timeout in seconds.").Default("0.5").Float64()
+	configCheck    = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
+	logProbeErrors = kingpin.Flag("log.probes-errors", "Log error from probe requests").Default().Bool()
```
SuperQ, line 53:
```suggestion
	logProbeErrors = kingpin.Flag("log.probe-errors", "Log error from probe requests").Default().Bool()
```

SuperQ (changes requested):

jkroepke:
@electron0zero Could you please also take a look here?

prober/handler.go
```diff
@@ -176,6 +179,15 @@ func (sl scrapeLogger) Log(keyvals ...interface{}) error {
 	sl.bufferLogger.Log(keyvals...)
 	kvs := make([]interface{}, len(keyvals))
 	copy(kvs, keyvals)
+
+	if sl.logErrors {
+		for i := 0; i < len(kvs); i += 2 {
+			if kvs[i] == level.Key() && kvs[i+1] == level.ErrorValue() {
```
electron0zero, line 185: this will only log errors from the checks? maybe we can make this configurable where users can configure the level they want to log? 

this would make this flag bit generic and more useful. we can rename the flag to something like `--log.prober=info`, and you can configure it to `--log.prober=error` to log only error, and replicate what we have in current state.

IMO this feature will be useful to users who are running BBE with central log collection in place.

we can set it to `error` by default or disable it by default, I am fine with both options.

cc @SuperQ what do you think?

jkroepke: What if `--log.prober=info`, but  `--log.level=warn` if set? 

Should info logs still visible on stdout?

jkroepke: @electron0zero @SuperQ how we can proceed here?

electron0zero (approved):
lgtm, will wait for @roidelapluie or @mem to take a look and approve :)

jkroepke:
@electron0zero @roidelapluie or @mem How we can proceed here?

electron0zero:
assuming that there are no comments or feedback from other maintainers, I am going to go ahead and merge it.

PR #1097, opened 2023-07-08
allow specification of source address for probes (http, grpc)

FOWind:
This adds the configuration option source_ip_address to modules HTTP and GRPC which changes the
prober's local address.


asokolov365:
Hi @FOWind
What's the status of the PR? I see the DCO check's failed. Is it the only reason why the PR is not merged yet?
Thank you!

FOWind:
I think it's not a necessary feature for the blackbox_exporter, so there is no response to this PR from maintainers.

PR #1099, opened 2023-07-18
Allow source_ip as a parameter

NPellet:
In our use cases, we have machines that have potentially hundreds of virtual interfaces, sitting in different networks
We'd like to set the source IP as a parameter instead of having to create ICMP and TCP probers for each interface we define.


NPellet:
I guess concerns are similar as in #1098, in the sense that this could enable probing a private network (which is exactly what we want in this case), so I would suggest also a config flag to allow this.

What is the fate of #1098 ? If this gets accepted then I would rather rebase on it and add the source ip there.

PR #1108, opened 2023-08-03
Add flag for debug probes and config hiding

### LINKED ISSUE #1107: Exposed sensitive data in debug probe, recent probe logs, and config endpoint
I have sensitive data in my TCP probe like this:
```
    my_tcp_probe:
      prober: tcp
      timeout: 5s
      tcp:
        query_response:
        - expect: "220"
        - send: "USER my_username\r"
        - expect: "331"
        - send: "PASS sensitive_data_here"
        - expect: "230"
        - send: "QUIT\r"
        - expect: "221"
 ```
I set `--log.level=error`, but someone can still access my sensitive data by visiting `/config` or `/probe?debug=true...`.

I would like to hide my sensitive data from these endpoints. Some ideas:
- flag `config.hide` will turn off `/config`
- flag `probe.no-debug` will not allow query parameter `debug=true` to be used
- flag `secure` will run blackbox-exporter in mode with only `/probe` endpoint without option `debug=true`

EDIT: Setting `history.limit=0` solves the problem with `/logs`

Maros112358:
fixes #1107

### DIFF
--- a/main.go
+++ b/main.go
@@ -188,7 +188,7 @@ func run() int {
 		sc.Lock()
 		conf := sc.C
 		sc.Unlock()
-		prober.Handler(w, r, conf, logger, rh, *timeoutOffset, nil, moduleUnknownCounter, logLevelProber, probeDebug)
+		prober.Handler(w, r, conf, logger, rh, *timeoutOffset, nil, moduleUnknownCounter, logLevelProber, *probeDebug)
 	})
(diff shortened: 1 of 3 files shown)


electron0zero:
going to close this, see https://github.com/prometheus/blackbox_exporter/issues/1107#issuecomment-1736145254 for details :)

PR #1115, opened 2023-08-27
Support multipart/form-data

hatamiarash7:
## Description

After this PR, we can send POST requests in `multipart/form-data` format. For example, users can test the upload process quickly.

```yaml
http_post_body_multipart:
    prober: http
    timeout: 5s
    http:
      method: POST
      body_multipart:
        - type: "file"
          key: "file"
          value: "/files/body.txt"
        - type: "text"
          key: "name"
          value: "arash"
```

We can define `text` or `file` fields here.

---

## Changes

- Support `multipart` for HTTP probe
- Update example file
- Update configuration document

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -158,9 +158,13 @@ modules:
   [ body: <string> ]
 
   # Read the HTTP request body from from a file.
-  # It is mutually exclusive with `body`.
+  # It is mutually exclusive with `body` and `body_multipart`.
   [ body_file: <filename> ]
 
+  # Read the HTTP request body from from a file.
+  # It is mutually exclusive with `body` and `body_file`.
+  [ body_multipart: <filename> ]
+    [ - <body_multipart_spec>, ... ]
 ```
 
 #### `<http_header_match_spec>`
@@ -169,6 +173,17 @@ modules:
 header: <string>,
 regexp: <regex>,
 [ allow_missing: <boolean> | default = false ]
+```
+
+#### `<body_multipart_spec>`
+
+```yml
+# The type of the multipart body part. Valid values are "file" and "text".
+type: <string>
+# The name of the multipart body part.
+key: <string>
+# The value of the multipart body part. It can be a file path or a string.
+value: <string>
 ```
 
 ### `<tcp_probe>`
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -25,8 +25,8 @@ modules:
 
 ```
 
-
 ### `<module>`
+
 ```yml
 
   # The protocol over which the probe will take place (http, tcp, dns, icmp, grpc).
@@ -45,6 +45,7 @@ modules:
 ```
 
 ### `<http_probe>`
+
 ```yml
 
   # Accepted status codes for this probe. Defaults to 2xx.
--- a/prober/http.go
+++ b/prober/http.go
@@ -14,6 +14,7 @@
 package prober
 
 import (
+	"bytes"
 	"compress/flate"
 	"compress/gzip"
 	"context"
@@ -21,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime/multipart"
 	"net"
 	"net/http"
 	"net/http/cookiejar"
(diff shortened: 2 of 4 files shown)


hatamiarash7:
any update?

hatamiarash7:
Hello

Do we have an update for this PR?

The latest changes have been synced and new tests have been added for multipart.

Unfortunately, the IPv6 tests did not pass, and https://github.com/prometheus/blackbox_exporter/pull/1342 will solve the problem. Otherwise, a separate update is needed.

prober/http.go
```diff
@@ -424,6 +426,60 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		body = body_file
 	}
 
+	// If a multipart form is configured, add it to the request.
```
electron0zero, line 429: this implementation and configuration feels a little awkward, can we make it simple and implement this with just one bool flag called `multipart`?

If the flag is set, you can just make the check multi-part. for text and data, you can use the existing `Body` and `BodyFile` config values?

this would make it easy to configure and simplify this implementation as well?

prober/http_test.go
```diff
@@ -1554,3 +1555,108 @@ func TestBody(t *testing.T) {
 		}
 	}
 }
+
+// Test that the HTTP probe module can handle a multipart form request
```
electron0zero, line 1559: I think this test can also be simplified after the implementation is updated.

electron0zero (changes requested):

electron0zero:
for CI failure: https://github.com/prometheus/blackbox_exporter/pull/1342 is merged, please rebase.

CONFIGURATION.md
```diff
@@ -25,8 +25,8 @@ modules:
 
 ```
 
-
 ### `<module>`
+
```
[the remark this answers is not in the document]
hatamiarash7: These blank lines are not extra. In a standard MD code, there should be a blank line before and after all headers. **The rest of this file follows this practice**.

Take a look at [MD022 rule](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md022---headers-should-be-surrounded-by-blank-lines).

prober/http.go
hatamiarash7: In a multipart request, submitting multiple files (not just one file or text) should be possible. This is not possible using `Body` and `BodyFile`.

prober/http_test.go
hatamiarash7: As explained in the previous comment, we need to be able to POST different files/text. For this purpose, the `bodyMultipart` variable has been defined and used in this test.

hatamiarash7:
> for CI failure: #1342 is merged, please rebase.

Rebased.

prober/http.go
electron0zero: so I have read up on the multipart, and I still don't understand this code.

```
type Multipart struct {
	Type  string `yaml:"type,omitempty"`
	Key   string `yaml:"key,omitempty"`
	Value string `yaml:"value,omitempty"`
} 
```

the part I am having hard time understanding is the why behind `Multipart` struct and the `Type`, and the `Key` and `Value` in it.

do you need multiple text fields? if yes, can you share more it?

I think it should be possible to have it something like. `MultipartFiles []string`, and pass multiple filenames for the multi-part check, and append them to the Body and use the `Body` or BodyText for the body text.

hatamiarash7: Imagine a scenario where two files and two texts are to be sent. Consider the following curl example:

```bash
curl --location 'https://domain.com' \
--form 'text-key-1="text-value-1"' \
--form 'text-key-2="text-value-2"' \
--form 'file-key-1=@"/files/file-1.jpeg"' \
--form 'file-key-2=@"/files/file-2.jpeg"'
```

If we want to implement something like this, we need a separate structure to set the `Type` (text/file), `Key`, and `Value`.

```yaml
http_post_body_multipart:
  prober: http
  timeout: 5s
  http:
    method: POST
    body_multipart:
      - type: "text-key-1"
        key: "name"
        value: "text-value-1"
      - type: "text-key-2"
        key: "name"
        value: "text-value-2"
      - type: "file"
        key: "file-key-1"
        value: "/files/file-1.jpeg"
      - type: "file"
        key: "file-key-2"
        value: "/files/file-2.jpeg"
```

hatamiarash7:
Any update?

Issue #1077, opened 2023-06-06
[feature request] probe (module scoped) failure reason metric


domcyrus:
## Current Situation
Currently, the `blackbox_exporter` provides some general metrics such as probe_success and probe_duration_seconds that apply universally to all modules. Additionally, specific modules like the http module / prober offer their own metrics like probe_http_status_code, which help monitor the availability and performance of http endpoints. However, when a probe fails, it can be challenging to pinpoint the exact cause of the failure without manually inspecting `blackbox_exporter` logs or attempting to reproduce the error, if possible.

## Proposal
To address this issue, we propose the addition of a new metric called probe_($MODULE)_failure_reason to the `blackbox_exporter`. This metric would provide more detailed information about the reasons behind probe failures. It would include a label named "reason" with descriptive and enumerable values such as "dns-resolution-error," "http-timeout," or "ssl-certificate-validation-failed," among others. Currently, these failures can only be inferred from the logged errors.

## Benefits
The introduction of the probe_($MODULE)_failure_reason metric would significantly enhance troubleshooting capabilities. In most cases users would be able to identify the root cause of a probe failure without the need for manual log inspection or additional testing. Moreover, this new metric would facilitate the setup of alerts and notifications tailored to specific failure scenarios.

## Contribution
We believe that incorporating the probe_($MODULE)_failure_reason metric would be a valuable enhancement for the `blackbox_exporter`, improving its usability and effectiveness. We would be happy to contribute to the development of this feature and provide feedback on its implementation.

Thank you for considering our proposal. If this is something that would be ok to go forward with we’d love to contribute the functionality to `blackbox_exporter`.

druanoor:
Relates to this: #1062

slrtbtfs:
Hi, I'm a bit confused as to why this is marked `closed as completed`, as it doesn't seem to have been merged.

That being said, I'd be very happy to see a feature like this in blackbox exporter and am thankful for the work you put into it!

beorn7:
> Hi, I'm a bit confused as to why this is marked closed as completed, 

I guess because that is the default if you simply hit the "close" button. :)

Maybe the actual meaning was "I won't have time to work on this anymore" or "I never got feedback from the maintainers, so I gave up."

domcyrus:
> > Hi, I'm a bit confused as to why this is marked closed as completed,
> 
> I guess because that is the default if you simply hit the "close" button. :)

Yes, sorry this is what I did.

> 
[...]

PR #1139, opened 2023-10-23
Add http failure reason metric #1077

domcyrus:
Adds `probe_http_failures_total` metric implementing feature for #1077 . This `couter` type metric provides the reason of http probe failures.

For example:
```
probe_http_failures_total{reason="dns_not_found"}
probe_http_failures_total{reason="request_creation_error"}
probe_http_failures_total{reason="request_get_timeout"}
...
```

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -324,7 +324,9 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		level.Error(logger).Log("msg", "Could not parse target URL", "err", err)
 		var ue *url.Error
 		if errors.As(err, &ue) {
-			probeFailureCounter.WithLabelValues(labelFromUrlError("parse", ue)).Inc()
+			probeFailureCounter.WithLabelValues(labelFromUrlError("urlparse", ue)).Inc()
+		} else {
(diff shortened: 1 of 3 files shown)


slrtbtfs:
Hi, first of all thanks for this PR!

May i ask why this PR introduces a global `probe_http_failures_total` metric instead of the per-probe `probe_http_failure_reason` metric proposed by #1077 ?

The latter seems like a better fit for the underlying issue to me.

(To avoid confusion: I'm not a maintainer, just someone who would like to use this feature.)

domcyrus:
> Hi, first of all thanks for this PR!
> 
> May i ask why this PR introduces a global `probe_http_failures_total` metric instead of the per-probe `probe_http_failure_reason` metric proposed by #1077 ?
> 
> The latter seems like a better fit for the underlying issue to me.
> 
> (To avoid confusion: I'm not a maintainer, just someone who would like to use this feature.)

@slrtbtfs Yes, looking at this again I also do think that we don't need to introduce a new global metric per-probe. Regarding the metric name in #1077 @SuperQ proposed to use `probe_failure_info`.

slrtbtfs:
Ok, if you don’t mind, I’ll try to come up with a new PR using probe_failure_info instead then.

PR #1145, opened 2023-11-03
fix: Correct tool existence checks

pulecp:
`command -v <cmd>` returns a path to a binary to stdout if it's found in PATH. The condition deciding if `<cmd>` is installed redirects stdout to `/dev/null` and so this check could never work.

This change redirects stderr to `/dev/null` instead of stdout to compare the output of `command -v <cmd>` properly with an empty string.

I unified the redirection to a file in whole file. The empty space behind redirection is not needed.

Signed-off-by: Pavel Pulec <pulecp@gmail.com>

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -49,7 +49,7 @@ endif
 GOTEST := $(GO) test
 GOTEST_DIR :=
 ifneq ($(CIRCLE_JOB),)
-ifneq ($(shell command -v gotestsum > /dev/null),)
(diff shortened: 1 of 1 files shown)


electron0zero:
hi @pulecp Makefile.common is managed in https://github.com/prometheus/prometheus repo, and synced here by automation.

please make this change in upstream. once merged in prometheus/prometheus, automation will sync it here.

electron0zero:
thanks for the PR but I am going to close this PR because we can't merge this change here.

PR #1162, opened 2023-11-28
security vulnerability fix - add new cmdline flags to disable/change …

123456890987654321:
…route for pprof debug endpoints

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,7 +19,7 @@ import (
 	"html"
 	"net"
 	"net/http"
-	_ "net/http/pprof"
+	"net/http/pprof"
 	"net/url"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


mem (approved):

SuperQ:
Sorry, we do not want to do this, as pprof is not a vulnerability.

mem:
> Sorry, we do not want to do this, as pprof is not a vulnerability.

pprof can be used to obtain the command line, the environment and the heap of the process, which is more than you can get by simply having shell access to the machine where blackbox exporter is running on. In the configuration you get by using a blank import, it's available without authentication to anyone who can connect to the port, and by default it listens on all interfaces.

SuperQ:
The exporter is also an open proxy. If security of the exporter's endpoint is an issue, we provide TLS and authentication.

Issue #1005, opened 2022-12-15
Breaking change in DNS Windows name resolution in last release due to change in [Resolver.PreferGo]


JDA88:
Some context first:

Most of our target are stored with their short name hostname01 instead of their full DNS name hostname01.site.domain
The main reason is that the target are retrieved from our CMDB via scripts and that in our environment the conversion from short name to FQDN is not fully predictable and depend on the site ; It is the “suffix list” in the LAN interface that usually take care of this.

### Host operating system: output of `uname -a`
Windows server
### blackbox_exporter version: output of `blackbox_exporter --version`
0.23.0

### What did you do that produced an error?
Using short DNS name in the target

### What did you expect to see?
Same behaviour as 0.22.0

### What did you see instead?
Failed to resolve the target

Same issues as in [prometheus #11480](https://github.com/prometheus/prometheus/issues/11480)

Root cause: a change from go1.19:
> 
> Resolver.PreferGo is now implemented on Windows and Plan 9. It previously only worked on Unix platforms. Combined with Dialer.Resolver and Resolver.Dial, it's now possible to write portable programs and be in control of all DNS name lookups when dialing.
> 
> The net package now has initial support for the netgo build tag on Windows. When used, the package uses the Go DNS client (as used by Resolver.PreferGo) instead of asking Windows for DNS results. The upstream DNS server it discovers from Windows may not yet be correct with complex system network configurations, however.
>

Wagnard:
I can confirm such behavior on my end with 0.23.0
For me it seems the probe fail because it is bypassing the windows hosts file. (where i configured some IPs and names)

PR #1183, opened 2024-01-15
Do not build with netgo on Windows

weidongkl:
#1005

### DIFF
--- a/.promu.yml
+++ b/.promu.yml
@@ -5,7 +5,11 @@ go:
 repository:
     path: github.com/prometheus/blackbox_exporter
 build:
-    flags: -a -tags netgo
+    tags:
+        all:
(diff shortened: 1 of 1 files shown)


.promu.yml
```diff
@@ -5,7 +5,11 @@ go:
 repository:
     path: github.com/prometheus/blackbox_exporter
 build:
-    flags: -a -tags netgo
+    tags:
+        all:
+            - netgo
+        windows:
+    flags: -a
```
SuperQ, line 12: This is actually an obsolete param setup. It can be fully removed as the defaults work correctly now.

```suggestion
```

weidongkl:
https://github.com/prometheus/blackbox_exporter/pull/1184

PR #1203, opened 2024-03-07
add metric to record tls ciphersuite negotiated during handshake

sding3:
Record name of ciphersuite negotiated during handshake in new `probe_tls_cipher_info` metric, useful for evaluating ciphers in use by monitored endpoints.

Example:

```shell
$ curl http://localhost:9115/probe?target=https://google.com&module=http_2xx

<...abbreviated...>

# HELP probe_tls_cipher_info Returns the TLS cipher negotiated during handshake or 0x000 when unknown
# TYPE probe_tls_cipher_info gauge
probe_tls_cipher_info{cipher="TLS_AES_128_GCM_SHA256"} 1
# HELP probe_tls_version_info Returns the TLS version used or NaN when unknown
```

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -282,6 +282,11 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			[]string{"version"},
 		)
 
+		probeTLSCipher = prometheus.NewGaugeVec(
+			probeTLSCipherGaugeOpts,
+			[]string{"cipher"},
+		)
+
 		probeHTTPVersionGauge = prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: "probe_http_version",
 			Help: "Returns the version of HTTP of the probe response",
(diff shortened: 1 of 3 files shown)


mem:
Looks good to me, it's inline with other information that's exported.

Reading thru the tls code I'm not 100% sure when `0x0000` will be used. Unless I'm missing something, the code cannot negotiate a suite that it doesn't know about, so it seems to me that 0 actually means "negotiation failed", is that correct?

sding3:
@mem , yes that's right, and in that case we wouldn't produce this new metric and some of the other TLS related metrics due to not being able to enter into the `if resp.TLS != nil { }` branch on line 644 in `prober/http.go`.

I've cleaned up the metric help text in b63e5ae791bccabbcd1bd2ba20fdc5e07996a86b to no longer mention `0x0000`.

mem:
> @mem , yes that's right, and in that case we wouldn't produce this new metric and some of the other TLS related metrics due to not being able to enter into the `if resp.TLS != nil { }` branch on line 644 in `prober/http.go`.
> 
> I've cleaned up the metric help text in [b63e5ae](https://github.com/prometheus/blackbox_exporter/commit/b63e5ae791bccabbcd1bd2ba20fdc5e07996a86b) to no longer mention `0x0000`.

Thanks, looks goood.

You will need to sign the DCO in order to get this merged.

sding3:
Oops, the DCO in the commit message should be fixed now.

sding3:
@mem - friendly ping.

mem (approved):
Oops, sorry, I missed this notification.

PR #1246, opened 2024-05-17
Fix generateCertificate helper function (prober tests)

santihernandezc:
All certificated created by `generateCertificate()` are self-signed. This PR fixes that by passing the correct argument for `parent`.


ArthurSens:
Hey, sorry for the super-delayed review. The codebase has changed quite a bit in the last few years, and the test you've changed is very different now.

I'll close this PR, but feel free to open new ones if you find more incorrect tests!

PR #1255, opened 2024-06-05
Support matching JSON body with CEL expressions

juho9000:
This PR implements parsing a JSON response from the body and checking the result using a CEL query. Looking for feedback on this.

Sample usage:

```
modules:
  http_2xx:
    prober: http
    http:
      fail_if_body_json_not_matches_cel: "body.foo.bar == 'baz'"
```

If response is `{"foo": { "bar": "qux" } }` the probe fails. fail_if_body_json_matches_cel is also implemented and should work as expected.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -89,6 +89,12 @@ modules:
   # Probe fails if SSL is not present.
   [ fail_if_not_ssl: <boolean> | default = false ]
 
+  # Probe fails if response body JSON matches CEL:
+  fail_if_body_json_matches_cel: <cel expression, root field is called body>
+
+  # Probe fails if response body JSON does not match CEL:
+  fail_if_body_json_not_matches_cel: <cel expression, root field is called body>
+
   # Probe fails if response body matches regex.
   fail_if_body_matches_regexp:
     [ - <regex>, ... ]
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -89,11 +89,11 @@ modules:
   # Probe fails if SSL is not present.
   [ fail_if_not_ssl: <boolean> | default = false ]
 
-  # Probe fails if response body JSON matches CEL:
-  fail_if_body_json_matches_cel: <cel expression, root field is called body>
+  # Probe fails if response body JSON matches the CEL expression or if response is not JSON. See: https://github.com/google/cel-spec:
+  fail_if_body_json_matches_cel: <string>
 
-  # Probe fails if response body JSON does not match CEL:
-  fail_if_body_json_not_matches_cel: <cel expression, root field is called body>
+  # Probe fails if response body JSON does not match CEL expression or if response is not JSON. See: https://github.com/google/cel-spec:
+  fail_if_body_json_not_matches_cel: <string>
 
   # Probe fails if response body matches regex.
   fail_if_body_matches_regexp:
--- a/config/config.go
+++ b/config/config.go
@@ -173,7 +173,7 @@ func NewCelProgram(s string) (CelProgram, error) {
 		return program, fmt.Errorf("error compiling CEL program: %s", issues.Err())
 	}
 
-	celProg, err := env.Program(ast)
+	celProg, err := env.Program(ast, cel.InterruptCheckFrequency(100))
 	if err != nil {
 		return program, fmt.Errorf("error creating CEL program: %s", err)
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -149,14 +149,14 @@ func (sc *SafeConfig) ReloadConfig(confFile string, logger *slog.Logger) (err er
 // CelProgram encapsulates a cel.Program and makes it YAML marshalable.
 type CelProgram struct {
 	cel.Program
-	expression string
+	Expression string
 }
 
 // NewCelProgram creates a new CEL Program and returns an error if the
 // passed-in CEL expression does not compile.
 func NewCelProgram(s string) (CelProgram, error) {
 	program := CelProgram{
-		expression: s,
+		Expression: s,
 	}
 
 	env, err := cel.NewEnv(
@@ -199,8 +199,8 @@ func (c *CelProgram) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 // MarshalYAML implements the yaml.Marshaler interface.
 func (c CelProgram) MarshalYAML() (interface{}, error) {
-	if c.expression != "" {
-		return c.expression, nil
+	if c.Expression != "" {
+		return c.Expression, nil
 	}
 	return nil, nil
 }
--- a/config/config.go
+++ b/config/config.go
@@ -28,6 +28,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/checker/decls"
 	yaml "gopkg.in/yaml.v3"
 
 	"github.com/alecthomas/units"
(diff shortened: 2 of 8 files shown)


juho9000:
@roidelapluie @mem ?

CONFIGURATION.md
```diff
@@ -89,6 +89,12 @@ modules:
   # Probe fails if SSL is not present.
   [ fail_if_not_ssl: <boolean> | default = false ]
 
+  # Probe fails if response body JSON matches CEL:
+  fail_if_body_matches_cel: <cel expression, root field is called body>
```
mem, line 93: this should be `fail_if_body_json_matches_cel`, right?

CONFIGURATION.md
```diff
@@ -89,6 +89,12 @@ modules:
...
+
+  # Probe fails if response body JSON does not match CEL:
+  fail_if_body_not_matches_cel: <cel expression, root field is called body>
```
mem, line 96: same here

prober/http.go
```diff
@@ -297,6 +351,11 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Help: "Indicates if probe failed due to regex",
 		})
 
+		probeFailedDueToCel = prometheus.NewGauge(prometheus.GaugeOpts{
```
mem, line 354: I'm thinking this should be added conditionally, if the configuration asks for validation using CEL expressions.

mem (commented):
A couple of minor comments.

mem:
Thank you for this, this seems useful!

Question, I have no experience with CEL... is it possible to use it to match HTML, too? I see in the code that Eval takes a `map[string]any`, but I couldn't spot how that `any` is handled.

The reason I'm asking is because I'm trying to figure out if it's possible to extend this to more inputs, or if it needs to be restricted to JSON.

prober/http.go
```diff
@@ -65,6 +67,58 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
+		level.Error(logger).Log("msg", "Error unmarshalling HTTP body", "err", err)
+		return false
+	}
+
+	evalPayload := map[string]interface{}{
+		"body": bodyJSON,
+	}
+
+	if httpConfig.FailIfBodyJSONMatchesCel != nil {
+		result, details, err := httpConfig.FailIfBodyJSONMatchesCel.Eval(evalPayload)
```
mem, line 88: Question: shouldn't this be EvalContext instead? I'm not clear about the guarantees that the evaluation will terminate, and the fact that EvalContext exists suggests that there's a benefit from imposing a deadline.

CONFIGURATION.md
juho9000: Correct

prober/http.go
juho9000: I was following the same logic that is applied to regex checks, following your logic that should also be conditionally added, no?

prober/http.go
juho9000: You're probably right, I've changed it to use ContextEval

juho9000:
> Thank you for this, this seems useful!
> 
> Question, I have no experience with CEL... is it possible to use it to match HTML, too? I see in the code that Eval takes a `map[string]any`, but I couldn't spot how that `any` is handled.
> 
> The reason I'm asking is because I'm trying to figure out if it's possible to extend this to more inputs, or if it needs to be restricted to JSON.

I atleast do not know how CEL could be used to match HTML in any sane way. You might want to check this out https://github.com/google/cel-spec/blob/master/doc/langdef.md#json-data-conversion

Thanks for your comments and suggestions so far. Really hoping we can get this merged. I think that having the ability to validate JSON responses would be an awesome addition to blackbox exporter. :)

zetaab:
@mem kindly ping :) we are intested about this feature and would like to see it as well

zetaab:
@mem @roidelapluie any comments to this PR?

juho9000:
@electron0zero @mem I haven't heard from any of the maintainers in a while. I fixed or clarified everything that was mentioned in the review comments. The initial message I got was that this seems useful, but it seems that it's forgotten now. Any update? I'd love to get this merged in order to avoid maintaining my own fork. :)

zetaab:
ping @electron0zero @mem could you give some opinion about this?

juho9000:
@mem I'm willing to donate you a sixpack in exchange for a review at this point. Please either review this or close it if it's not going to make it way into blackbox exporter.

juho9000:
bernie

CONFIGURATION.md
```diff
@@ -89,6 +89,12 @@ modules:
   # Probe fails if SSL is not present.
   [ fail_if_not_ssl: <boolean> | default = false ]
 
+  # Probe fails if response body JSON matches CEL:
+  fail_if_body_json_matches_cel: <cel expression, root field is called body>
```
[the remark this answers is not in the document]
electron0zero: I would prefer if made this explicit to users that the target needs to return json when they are using this config.

also, can you update examples.yaml to include few examples on how to use this? and maybe link to CEL docs so users know what CEL is, and how it can be used.

prober/http.go
```diff
@@ -64,6 +66,58 @@ func matchRegularExpressions(reader io.Reader, httpConfig config.HTTPProbe, logg
...
 }
 
+func matchCelExpressions(ctx context.Context, reader io.Reader, httpConfig config.HTTPProbe, logger *slog.Logger) bool {
+	body, err := io.ReadAll(reader)
+	if err != nil {
+		logger.Error("Error reading HTTP body", "err", err)
+		return false
+	}
+
+	bodyJSON := make(map[string]interface{})
```
dgl, line 76: What if the response is a JSON list or a string (or even just `0` or `null`)? It feels like this should be declared as `any` (which matches the `interface{}` where `body` is passed below). Probably worth adding a test for each type as well, but it looks like it should work, provided the CEL expression expects the right type.

```suggestion
	var bodyJSON any
```

prober/http.go
```diff
@@ -296,6 +350,11 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			Help: "Indicates if probe failed due to regex",
 		})
 
+		probeFailedDueToCel = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_failed_due_to_cel",
+			Help: "Indicates if probe failed due to CEL",
```
dgl, line 355: `due to CEL` is missing a qualifier, how about `due to CEL not matching`

config/config.go
```diff
@@ -144,6 +146,74 @@ func (sc *SafeConfig) ReloadConfig(confFile string, logger *slog.Logger) (err er
 	return nil
 }
 
+// CelProgram encapsulates a cel.Program and makes it YAML marshalable.
```
dgl, line 149: Across this PR it would be better to use `CEL` consistently in identifier names, rather than `Cel`, per https://go.dev/wiki/CodeReviewComments#initialisms

prober/http.go
juho9000: This works similarily to the regex one, so if the probe fails due to the CEL expression matching or CEL expression not matching it will be increased in both cases.

.idea/.gitignore
dgl, line 1: This seems to be adding some `.idea` files now, could you revert adding those?

config/config.go
```diff
@@ -144,6 +145,72 @@ func (sc *SafeConfig) ReloadConfig(confFile string, logger *slog.Logger) (err er
...
+
+// NewCELProgram creates a new CEL Program and returns an error if the
+// passed-in CEL expression does not compile.
+func NewCELProgram(s string) (CELProgram, error) {
+	program := CELProgram{
+		Expression: s,
+	}
+
+	env, err := cel.NewEnv(
+		cel.Variable("body", cel.MapType(cel.StringType, cel.DynType)),
```
dgl, line 162: To allow any type, this can be declared as a totally dynamic type:

```suggestion
		cel.Variable("body", cel.DynType),
```

prober/http_test.go
```diff
@@ -936,6 +936,188 @@ func TestFailIfNotSSLLogMsg(t *testing.T) {
...
+			respBody:       `{"foo": {"bar": "baz"}}`,
+			celExpression:  "body.foo.bar",
+			expectedResult: false,
+		},
+		"body is not json": {
+			respBody:       "hello world",
+			celExpression:  "body.foo.bar == 'baz'",
+			expectedResult: false,
+		},
+		"body is empty json": {
```
dgl, line 1062:
```suggestion
		"body is empty json object": {
```

prober/http_test.go
```diff
@@ -936,6 +936,188 @@ func TestFailIfNotSSLLogMsg(t *testing.T) {
...
+			respBody:       "{}",
+			celExpression:  "!has(body.foo)",
+			expectedResult: true,
+		},
```
dgl, line 1066: Add some tests that non-object types work:

```suggestion
		},
               "body is json string": {
                       respBody:       `"foo"`,
                       celExpression:  "body == 'foo'",
                       expectedResult: true,
               },
               "body is json list": {
                       respBody:       `["foo","bar","baz"]`,
                       celExpression:  "body[2] == 'baz'",
                       expectedResult: true,
               },
               "body is json boolean": {
                       respBody:       `true`,
                       celExpression:  "body",
                       expectedResult: true,
               },
```

prober/http_test.go
```diff
@@ -936,6 +936,188 @@ func TestFailIfNotSSLLogMsg(t *testing.T) {
...
+			respBody:       `{"foo": {"bar": "baz"}}`,
+			celExpression:  "body.foo.bar",
+			expectedResult: false,
+		},
+		"body is not json": {
+			respBody:       "hello world",
+			celExpression:  "body.foo.bar == 'baz'",
+			expectedResult: false,
+		},
+		"body is empty json": {
```
dgl, line 970:
```suggestion
		"body is empty json object": {
```

prober/http_test.go
```diff
@@ -936,6 +936,188 @@ func TestFailIfNotSSLLogMsg(t *testing.T) {
...
+			respBody:       "{}",
+			celExpression:  "body.foo.bar == 'baz'",
+			expectedResult: false,
+		},
```
dgl, line 974:
```suggestion
		},
               "body is json string": {
                       respBody:       `"foo"`,
                       celExpression:  "body == 'foo'",
                       expectedResult: false,
               },
               "body is json list": {
                       respBody:       `["foo","bar","baz"]`,
                       celExpression:  "body[2] == 'baz'",
                       expectedResult: false,
               },
               "body is json boolean": {
                       respBody:       `true`,
                       celExpression:  "body",
                       expectedResult: false,
               },
```

dgl (commented):
I started reviewing if JSON types other than objects would work, in the end I wrote some tests and made them work.

(This will probably need gofmt before committing, I should just have pushed a commit...)

dgl (approved):
LGTM, after comment resolved. Thanks! I'll leave final review to a maintainer though.

electron0zero (approved):

electron0zero:
thank you for your contribution , and patience with the review process :bow:

PR #1257, opened 2024-06-17
feat: Add optional `/logs?module=` filter parameter

EconomicTouristsArmLate:
This allows to uniquely identify a probe log by target and module.

Previously with the `?target=` selector only,  it was impossible to select the correct probe log, when multiple modules had probed a target.


EconomicTouristsArmLate:
not stale, just waiting for review...
just rebased onto latest master.

prober/history_test.go
```diff
@@ -127,17 +128,37 @@ func TestHistoryGetByTarget(t *testing.T) {
...
 
-	resultFalse := history.GetByTarget("target-1")
+	resultFalse := history.GetByTarget("target-1", "")
 	if resultFalse == nil {
 		t.Errorf("Error finding the result in history by target for target-1")
 	} else {
 		if resultFalse.Target != "target-1" {
 			t.Errorf("Error finding the result in history by target for target: expected \"%s\" and got \"%s\"", "target-1", resultFalse.Target)
 		}
+		if resultFalse.ModuleName != "module-1" {
```
electron0zero, line 140: we also have `"module-0", "target-1"` entry, so shouldn't we check for both?

prober/history.go
```diff
@@ -97,18 +97,18 @@ func (rh *ResultHistory) GetById(id int64) *Result {
 	return nil
 }
 
-// Get returns a given result by url.
-func (rh *ResultHistory) GetByTarget(target string) *Result {
+// GetByTarget returns a given result by url, optionally filtered by a module name.
```
electron0zero: nit: maybe we should rename the method to show that module is also supported.

electron0zero (commented):
few nits and a comment, rest lgtm.


can you also update the PR description as per the new PR template: https://github.com/prometheus/blackbox_exporter/blob/master/.github/PULL_REQUEST_TEMPLATE.md

prober/history_test.go
EconomicTouristsArmLate: the query would match both entries `module-0 target-1`, `module-1 target-1`, but the function only returns the first match.

you mean i should add another testcase, with a different order of history entries, to make the matching-by-insert-order more explicit? not sure it actually adds more coverage, but i guess it makes it more explicit

EconomicTouristsArmLate:
@electron0zero  thanks for the review.
i addressed your comments and rebased onto current master.
would be really happy to see a merge

prober/history_test.go
electron0zero: yeah, that would make it more clear.

electron0zero:
lgtm, I will for few weeks to allow other maintainers to comment on it. feel free to ping me to merge if we have comments from other maintainers.

PR #1273, opened 2024-08-01
pass params as a struct to ProbeFn

LaurenceChau:
I have a use case which is to send icmp ping and tcp ping via different network interfaces in a machine with multiple network cards. Although the source ip can be configured through blackbox.yaml, I think it would be more fixable if we can control the source ip of the probeFn through remote parameters of Prometheus request. 

To do so, I suggest we create a probeOpts struct. Instead of passing the target to probeFn, we can pass the struct with target value inside to probeFn. This change also allow different ProbeFn to have different parameters in the future.

### DIFF
--- a/prober/dns.go
+++ b/prober/dns.go
@@ -124,7 +124,7 @@ func validRcode(rcode int, valid []string, logger log.Logger) bool {
 	return false
 }
 
-func ProbeDNS(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger log.Logger) bool {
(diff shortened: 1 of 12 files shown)


SuperQ (changes requested):
For the same reasons as previous requests to add more query parameters to the exporter we can not accept this.

LaurenceChau:
Since the source IP is allowed to be configured in the config file, it is reasonable that the value should also be able to be configured through remote parameters.

SuperQ:
For the same reasons as #1098, we do not want a proliferation of module configurations as query params.

PR #1276, opened 2024-08-06
Add `fail_if_body_too_large` http probe option

hashworks:
This can be used in conjunction with `body_size_limit` when it is expected that endpoints return a body larger than the specified limit, e.g. streaming endpoints that potentially return infinite data.


prober/http.go
```diff
@@ -550,7 +550,7 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 
 		if !requestErrored {
 			_, err = io.Copy(io.Discard, byteCounter)
-			if err != nil {
+			if err != nil && (errors.Is(err, &http.MaxBytesError{}) || httpConfig.GetFailIfBodyTooLarge()) {
```
hashworks, line 553: Please check if this is okay or if I should move this to another `if` so it's clearer.

config/config.go
```diff
@@ -450,6 +451,13 @@ func (s *HeaderMatch) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+func (h HTTPProbe) GetFailIfBodyTooLarge() bool {
```
hashworks, line 454: Alternatively I could do the `nil` check in `UnmarshalYAML`, but then it could panic if someone changes the variable under the hood later on.

hashworks:
Do you want any changes here? Or is this a feature you don't want merged?

electron0zero:
can you share more on the use-case? and what problem this flag will solve?

once I better understand the "why?" behind the change, I will find some time to review.

CONFIGURATION.md
```diff
@@ -89,6 +89,9 @@ modules:
   # Probe fails if SSL is not present.
   [ fail_if_not_ssl: <boolean> | default = false ]
 
+  # Probe fails if a defined body_size_limit is exceeded.
+  [ fail_if_body_too_large: <boolean> | default = true ]
```
electron0zero, line 93: can you also add an example in the example.yml?

hashworks: Done in [081a0e3](https://github.com/prometheus/blackbox_exporter/pull/1276/commits/081a0e33ad59664be888e1644dbd2b7668e4925c)

hashworks:
> can you share more on the use-case? and what problem this flag will solve?

At work, we need to probe endpoints that endlessly stream data. Without this flag every probe will fail after some time. We want to measure successful requests and "time to X bytes".

config/config.go
electron0zero: To preserve the current behaviour, `FailIfBodyTooLarge` should be set to `true` by default.

you should be able to set it to `true` by default in DefaultHTTPProbe at https://github.com/streamonkey/blackbox_exporter/blob/master/config/config.go#L51


if we make this bool instead of *bool, we do't have to worry about `nil`, and can remove this whole function.

prober/http.go
electron0zero: I would prefer another if for clarity

hashworks: Changed in 39729e0

config/config.go
```diff
@@ -211,6 +211,7 @@ type HTTPProbe struct {
 	NoFollowRedirects            *bool                   `yaml:"no_follow_redirects,omitempty"`
 	FailIfSSL                    bool                    `yaml:"fail_if_ssl,omitempty"`
 	FailIfNotSSL                 bool                    `yaml:"fail_if_not_ssl,omitempty"`
+	FailIfBodyTooLarge           *bool                   `yaml:"fail_if_body_too_large,omitempty"`
```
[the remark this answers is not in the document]
hashworks: I used *bool with `omitempty` to allow a `true` default, but this is fixed by 0ad871e.

prober/http_test.go
```diff
@@ -838,6 +839,32 @@ func TestFailIfNotSSL(t *testing.T) {
...
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		resp := bytes.Repeat([]byte{'A'}, int(bodySizeLimit)+1)
+
+		w.Header().Set("Content-Length", strconv.Itoa(len(resp)))
+		w.WriteHeader(http.StatusOK)
+		w.Write(resp)
+	}))
+	defer ts.Close()
+
+	for _, failIfBodyTooLarge := range []bool{true, false} {
```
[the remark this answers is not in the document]
hashworks: Fixed in 137226a

hashworks:
Sorry for the delay, added the requested changes.

hashworks:
Fixed the DCO CI failures.

hashworks:
I don't think the test failures are related to my changes. Are they green on master?

electron0zero:
> Are they green on master?

yes, try rebasing on the latest main.

hashworks:
Hm... the branch ist up2date.

Issue #1081, opened 2023-06-23
Request to add support for WebSocket


thevops:
There is an old issue about WebSocket probe: https://github.com/prometheus/blackbox_exporter/issues/536

It'd be nice to reconsider adding WebSocket support to Blackbox. The popularity of this protocol is growing.
What do you think?

SuperQ:
Seems like a good idea. Can you write out a proposal for what this would look like? What additional parameters / features would be needed.

thevops:
# Proposal

The popularity of the WebSocket protocol is growing. Having support for this protocol would be very useful.

## Features

The WebSocket probe could have the same attributes as the HTTP probe, like support for Basic Auth or TLS checking. Although a Minimum Viable Product can support only essential functions: sending requests with query data, receiving answers, and comparing received data to expected data.

## Feedback

I’m ready to test the WebSocket probe implementation, provide feedback and contribute as much as I can.

muety:
Any changes to have blackbox exporter support websocket any time soon?

adnull:
@muety as the PR is still not accepted you can try using websocket support version directly from my repo. 
https://github.com/adnull/blackbox_exporter/tree/websocket-support

electron0zero:
left some review comments on https://github.com/prometheus/blackbox_exporter/pull/1278#pullrequestreview-3397863502, will review it again once comments are addressed.

ivankatliarchuk:
I think this one is resolved https://github.com/prometheus/blackbox_exporter/pull/1278#pullrequestreview-3397863502. Maybe worth to close or something else left?

electron0zero:
We are missing `probe_websocket_ssl` metric for for websocket prober: @adnull said they will send a follow up PR to add it https://github.com/prometheus/blackbox_exporter/pull/1278#discussion_r2757984786

so I am looking for PR to add probe_websocket_ssl and then websocket will be in paririty with other probers.

adnull:
Yeah, I'm still lack of time to manage that. But still hoping to implement it.

adnull:
@electron0zero added the metric and unified SSL metrics registry in the https://github.com/prometheus/blackbox_exporter/pull/1574 PR

PR #1278, opened 2024-08-11
feat: Add websocket connection prober support

adnull:
According to #1081 this change adds websocket target support to monitor availability and functionality of websocket servers. 

This probe provides features as follows:
- Basic HTTP authentification
- Bearer token
- Arbitrary HTTP headers
- TLS configuration
- Send/Receive sequences with regexp matching to ensure the server is working correctly ( similar to the TCP probe)

New metrics:
- probe_websocket_status_code: resulting http status code on the connection
- probe_websocket_connection_upgraded: is the connection successfuly upgraded to a websocket connection
- probe_websocket_failed_due_to_regex: 1 when send/receive sequence wasn't matched to server messages

CHANGES:

```
[FEATURE] Add websocket connection prober support
```

- [x] Tests updated
- [x] Documentation added
- [ ] CHANGELOG added in release-notes section of PR Desc.


sanya2022:
Any updates here?

adnull:
> Any updates here?

If only.. althrough the year isn't passed yet :)

bilkoua:
Could you either approve this merge or let us know what needs to be revised?
@electron0zero @SuperQ

config/config.go
```diff
@@ -292,6 +294,27 @@ type DNSRRValidator struct {
 	FailIfNoneMatchesRegexp []string `yaml:"fail_if_none_matches_regexp,omitempty"`
 }
 
+type WebsocketProbe struct {
```
electron0zero, line 409: can you add tests for the probe config.

CONFIGURATION.md
```diff
@@ -327,9 +328,105 @@ tls_config:
...
+query_response:
+  [ - [ [ expect: <string> ],
+        [ send: <string> ],
+        [ starttls: <boolean | default = false> ]
+      ], ...
+  ]
+
+```
+
+### `<tls_config>`
```
electron0zero, line 503: I am bit confused by this change in tls_config, why is this added?

example.yml
```diff
@@ -209,3 +209,18 @@ modules:
...
       preferred_ip_protocol: "ip4" # defaults to "ip6"
       query_name: "www.prometheus.io"
+  websocket_example:
+    prober: websocket
+    websocket:
+      http_config:
+        basic_auth:
+          username: "user"
+          password: "password"
+        bearer_token: "secret_token"
```
electron0zero: does it make sense to configure `bearer_token` and `basic_auth` both? usually you would use one or the other. if you want to show an example of both config, please create two examples instead of putting both in one example.

prober/websocket.go
```diff
@@ -0,0 +1,137 @@
...
+	targetURL, err := url.Parse(target)
+	if err != nil {
+		logger.Error("Could not parse target URL", "err", err)
+		return false
+	}
+
+	logger.Debug("probing websocket", "target", targetURL.String())
+
+	httpStatusCode := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_http_status_code",
```
electron0zero:
```suggestion
		Name: "probe_websocket_status_code",
```

`probe_http_status_code` metric name already exists in http prober.  we have probe_<prober>_status_code metric names, so please follow that.

prober/websocket.go
```diff
@@ -0,0 +1,137 @@
...
+		Help: "Response HTTP status code",
+	})
+	isConnected := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_is_upgraded",
```
electron0zero:
```suggestion
		Name: "probe_websocket_connection_upgraded",
```

please follow `probe_<prober>_` pattern here as well. 

also `is_upgraded` is not very clear on "what" is upgraded, so use `connection_is_upgraded` or `connection_upgraded`

please see https://prometheus.io/docs/practices/naming/ to read more about naming best practices.

config/config.go
```diff
@@ -292,6 +294,27 @@ type DNSRRValidator struct {
...
+	HTTPClientConfig HTTPClientConfig `yaml:"http_config,omitempty"`
+	QueryResponse    []QueryResponse  `yaml:"query_response,omitempty"`
+}
+
+type HTTPClientConfig struct {
+	HTTPHeaders        map[string]interface{} `yaml:"headers,omitempty"`
+	BasicAuth          HTTPBasicAuth          `yaml:"basic_auth,omitempty"`
+	BearerToken        string                 `yaml:"bearer_token,omitempty"`
+	InsecureSkipVerify bool                   `yaml:"insecure_skip_verify,omitempty"`
```
electron0zero: `InsecureSkipVerify` is part of tls config, please follow the example of other probe configs and expose full tls_config, like we do it gRPC and other probes.

config/config.go
```diff
@@ -292,6 +294,27 @@ type DNSRRValidator struct {
 	FailIfNoneMatchesRegexp []string `yaml:"fail_if_none_matches_regexp,omitempty"`
 }
 
+type WebsocketProbe struct {
+	HTTPClientConfig HTTPClientConfig `yaml:"http_config,omitempty"`
+	QueryResponse    []QueryResponse  `yaml:"query_response,omitempty"`
+}
+
+type HTTPClientConfig struct {
```
electron0zero: any reasons to redefine this instead of using `config.HTTPClientConfig` from `github.com/prometheus/common/config` package?

we use it in HTTPProbe so it would be a good idea to use the same struct and keep things common and uniform.

config/config.go
```diff
@@ -292,6 +294,27 @@ type DNSRRValidator struct {
...
+	HTTPHeaders        map[string]interface{} `yaml:"headers,omitempty"`
+	BasicAuth          HTTPBasicAuth          `yaml:"basic_auth,omitempty"`
+	BearerToken        string                 `yaml:"bearer_token,omitempty"`
+	InsecureSkipVerify bool                   `yaml:"insecure_skip_verify,omitempty"`
+}
+
+type HTTPBasicAuth struct {
```
electron0zero: same comment as HTTPClientConfig, the BasicAuth in the github.com/prometheus/common/config has File support and is more rich, any reasons to NOT use it?

if we use github.com/prometheus/common/config's HTTPClientConfig, we don't need this type.

prober/websocket.go
```diff
@@ -0,0 +1,137 @@
...
+		Help: "Indicates if the websocket connection was successfully upgraded",
+	})
+
+	registry.MustRegister(isConnected)
+	registry.MustRegister(httpStatusCode)
+
+	dialer := websocket.Dialer{
+		TLSClientConfig: &tls.Config{
```
electron0zero: once we use the TLSConfig from `github.com/prometheus/common/config`, please configure more fields so end users can control those.

TLS MinVersion and TLS ServerName seems to be something I can think we need to expose but it also make sense to expose CA cert and client cert configs so folks can configure tls for websocket probes.

prober/websocket.go
```diff
@@ -0,0 +1,137 @@
...
+		logger.Error("Error dialing websocket", "err", err)
+		return false
+	}
+	defer connection.Close()
+
+	isConnected.Set(1)
+
+	if len(module.Websocket.QueryResponse) > 0 {
+		probeFailedDueToRegex := prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "probe_failed_due_to_regex",
```
electron0zero: same here, please see metric names in other probers and follow their lead.

prober/websocket.go
```diff
@@ -0,0 +1,137 @@
...
+			Help: "Indicates if probe failed due to regex",
+		})
+		registry.MustRegister(probeFailedDueToRegex)
+
+		queryMatched := true
+		for _, qr := range module.Websocket.QueryResponse {
```
electron0zero, line 129: nit: this block is deeply nested and hard to read. can you simplify it and make it easier to read please.

electron0zero (changes requested):

electron0zero:
please rebase with master and resolve conflicts and update the PR desc to follow the new PR template: https://github.com/prometheus/blackbox_exporter/blob/master/.github/PULL_REQUEST_TEMPLATE.md 🙏🏼

prober/websocket.go
adnull:
```suggestion
			Name: "probe_websocket_failed_due_to_regex",
```

config/config.go
adnull: config.HTTPClientConfig is used to construct new http client using prometheus internal functions while I need to create a custom Dialer with a set of http headers to establish a connection. I still don't see any way to create raw headers from it, so it needs a third-party implementation to provide full support of the config options (including password files etc).  That's why I made my own config structure.

bilkoua:
@adnull Could you bring the pull request up to date so it can be merged into master? Do you need any help with that?

adnull:
@bilkoua Hi! Will do shortly.

bilkoua:
@adnull any updates? Do you need any help with that?

adnull:
@bilkoua Just rebased on master, fixed suggested things, except using standart http and tls config structures. I can't use prometheus HTTPConfig as TLSConfig as they are only used in prometheus functions and probes whereas I need to make Dialer myself and I'm only able to pass headers into it, so it needs additional implementation to construct headers from prometheus config structures if we want to use it.

bilkoua:
@adnull Thank you very much. Could you also fix `ci/circleci: test `and `DCO`? Without that, it’s unlikely we’ll be able to proceed with the pull request.

adnull:
@bilkoua Sure, fixed the things

bilkoua:
@electron0zero Could you take a look at this PR after the changes?

example.yml
```diff
@@ -239,3 +239,33 @@ modules:
         X-API-Key:
           files:
             - /path/to/api-key.txt
+
```
electron0zero:
```suggestion
```

trim this line and add the missing newline at the end of file

CONFIGURATION.md
```diff
@@ -416,9 +417,161 @@ tls_config:
...
+  [ - [ expect: <regex> ],
+        [ send: <string> ],
+        [ starttls: <boolean | default = false> ]
+      ], ...
+  ]
+
+```
+
+**Example configurations:**
+
```
electron0zero, line 451: you should link to example.yaml for examples and add all the examples in one place (example.yml), only add the config options exposed by the websocket probe in this page.

CONFIGURATION.md
electron0zero: can you address this requested change? this looks like a duplicate of the existing `<tls_config>` section

CONFIGURATION.md
```diff
@@ -416,9 +417,161 @@ tls_config:
...
+# Optional parameters to append to the token URL.
+endpoint_params:
+  [ <string>: <string> ... ]
+```
+
 ### `<tls_config>`
 
 ```yml
+[ http_config: <websocket_http_config>]
```
electron0zero: seems like a stray change? can you remove it?

CONFIGURATION.md
```diff
@@ -416,9 +417,161 @@ tls_config:
...
+
+# Maximum acceptable TLS version. Accepted values: TLS10 (TLS 1.0), TLS11 (TLS
+# 1.1), TLS12 (TLS 1.2), TLS13 (TLS 1.3).
+# Can be used to test for the presence of insecure TLS versions.
+# If unset, Prometheus will use Go default maximum version, which is TLS 1.3.
+# See MaxVersion in https://pkg.go.dev/crypto/tls#Config.
+[ max_version: <string> ]
+```
+
+#### `<oauth2>`
```
electron0zero, line 536: can this be removed as well, looks like a copy of existing oauth config section?

config/config.go
```diff
@@ -399,6 +402,27 @@ type DNSRRValidator struct {
...
+type WebsocketProbe struct {
+	WSHTTPClientConfig WSHTTPClientConfig `yaml:"http_config,omitempty"`
+	QueryResponse      []QueryResponse    `yaml:"query_response,omitempty"`
+}
+
+type WSHTTPClientConfig struct {
+	HTTPHeaders map[string]interface{} `yaml:"headers,omitempty"`
+	BasicAuth   HTTPBasicAuth          `yaml:"basic_auth,omitempty"`
+	BearerToken string                 `yaml:"bearer_token,omitempty"`
+	TLSConfig   *tls.Config            `yaml:"tls_config,omitempty"`
```
electron0zero: any reasons to not use `config.TLSConfig` in this case? we use that type across all probes so it's better to stay consistent with TLS config options we expose?

config/config.go
```diff
@@ -399,6 +402,27 @@ type DNSRRValidator struct {
...
 
+type WebsocketProbe struct {
+	WSHTTPClientConfig WSHTTPClientConfig `yaml:"http_config,omitempty"`
+	QueryResponse      []QueryResponse    `yaml:"query_response,omitempty"`
+}
+
+type WSHTTPClientConfig struct {
+	HTTPHeaders map[string]interface{} `yaml:"headers,omitempty"`
+	BasicAuth   HTTPBasicAuth          `yaml:"basic_auth,omitempty"`
+	BearerToken string                 `yaml:"bearer_token,omitempty"`
```
electron0zero: This is a secret so you should be using Secret type from the common/config: https://github.com/prometheus/common/blob/d80d8544703e59a080a204b6f7429ac6561fb24f/config/config.go#L28

if you don't use Secret type in this case, the token will be logged by the logs and will be exposed in the config page.

config/config.go
```diff
@@ -399,6 +402,27 @@ type DNSRRValidator struct {
 	FailIfNoneMatchesRegexp []string `yaml:"fail_if_none_matches_regexp,omitempty"`
 }
 
+type WebsocketProbe struct {
+	WSHTTPClientConfig WSHTTPClientConfig `yaml:"http_config,omitempty"`
+	QueryResponse      []QueryResponse    `yaml:"query_response,omitempty"`
+}
+
+type WSHTTPClientConfig struct {
+	HTTPHeaders map[string]interface{} `yaml:"headers,omitempty"`
```
electron0zero: please use https://github.com/prometheus/common/blob/main/config/headers.go

they have secret type, support for files and validation logic methods that you can reuse those.

see https://github.com/prometheus/common/blob/d80d8544703e59a080a204b6f7429ac6561fb24f/config/headers.go#L116 on how to build headers from all sources.

config/config.go
```diff
@@ -399,6 +402,27 @@ type DNSRRValidator struct {
...
+	BasicAuth   HTTPBasicAuth          `yaml:"basic_auth,omitempty"`
+	BearerToken string                 `yaml:"bearer_token,omitempty"`
+	TLSConfig   *tls.Config            `yaml:"tls_config,omitempty"`
+}
+
+type HTTPBasicAuth struct {
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
```
electron0zero: password should be Secret type.

config/config.go
```diff
@@ -399,6 +402,27 @@ type DNSRRValidator struct {
...
+}
+
+func (c *HTTPBasicAuth) BasicAuthHeader() string {
+	return "Basic " + base64.StdEncoding.EncodeToString([]byte(c.Username+":"+c.Password))
```
electron0zero: since we are converting the basic auth into a Header, do we need to support a basic auth? 

we can just document that they can use Basic Auth with headers and let users set this header by supported methods, including files, etc.

electron0zero (changes requested):

bilkoua:
@electron0zero Thanks
@adnull Could you make these fixes?

bilkoua:
@adnull any updates? Do you need any help with that?

adnull:
Hi, working on that, will update soon

adnull:
Hi, working on that, will update soon

adnull:
> @adnull any updates? 

finally got time to address that. Had to implement my own headers constructor to use with the prometheus.config.HTTPClientConfig. @electron0zero btw just found out the config documentation isn't up to date for http_probe with the real code (it uses HTTPClientConfig but the docs still don't reflect it still using basic_auth, etc.)

bilkoua:
@adnull Thank you very much. Could you also fix tests?

adnull:
> @adnull Thank you very much. Could you also fix tests?

Yep, just've fixed.

bilkoua:
@electron0zero Could you take a look at this PR after the changes?

example.yml
```diff
@@ -239,3 +239,32 @@ modules:
...
+        authorization:
+          credentials: "secret_token"
+        tls_config:
+          insecure_skip_verify: true
+      headers:
+        X-Some-Header: "my_header"
+      query_response:
+        - expect: ^Hello,\s(.+)"
+          send: "Hello server, i'am ${1}"
+        - expect: ^Welcome
```
electron0zero: nit: missing newline, please add.

prober/websocket.go
```diff
@@ -0,0 +1,192 @@
...
+		logger.Debug("message sent", "message", send)
+	}
+	return true
+}
+
+func processWebsocketQueryRegexp(qr *config.QueryResponse, message []byte, logger *slog.Logger) (string, bool) {
+	send := qr.Send
+
+	if qr.Expect.Regexp != nil {
+		match := qr.Expect.Regexp.FindSubmatchIndex(message)
```
electron0zero:
```suggestion
		match := qr.Expect.FindSubmatchIndex(message)
```

prober/websocket.go
```diff
@@ -0,0 +1,192 @@
...
+		if match == nil {
+			logger.Error("Regexp did not match", "regexp", qr.Expect.Regexp, "line", message)
+			return "", false
+		}
+
+		logger.Debug("regexp matched", "regexp", qr.Expect.Regexp, "line", message)
+		send = string(qr.Expect.Regexp.Expand(nil, []byte(send), message, match))
```
electron0zero:
```suggestion
		send = string(qr.Expect.Expand(nil, []byte(send), message, match))
```

prober/websocket.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		logger.Error("Error dialing websocket", "err", err)
+		return false
+	}
+	defer connection.Close()
+
+	isConnected.Set(1)
+
+	if len(module.Websocket.QueryResponse) > 0 {
+		probeFailedDueToRegex := prometheus.NewGauge(prometheus.GaugeOpts{
```
electron0zero: should be move up top with other metrics and should not be under the if condition.

example.yml
adnull: ok np

prober/websocket.go
```diff
@@ -0,0 +1,192 @@
...
+	logger.Debug("probing websocket", "target", targetURL.String())
+
+	httpStatusCode := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_websocket_status_code",
+		Help: "Response HTTP status code",
+	})
+	isConnected := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "probe_websocket_connection_upgraded",
+		Help: "Indicates if the websocket connection was successfully upgraded",
+	})
```
electron0zero, line 50: the prober is missing `probe_websocket_duration_seconds` metric, please see code in http, tcp and other prober and add a metric that matchs those prober.

`duration_seconds` is a basic metric that you expect from your prober to know the latency of the probes you are running. in lots of the cases, you also care about latency along with probe pass/fail.

having phases would be useful as well because then it helps you narrow down where the time is going, so something to consider as well.

prober/websocket.go
```diff
@@ -0,0 +1,192 @@
...
+	registry.MustRegister(isConnected)
+	registry.MustRegister(httpStatusCode)
+
+	tlsConfig, err := promconfig.NewTLSConfig(&module.Websocket.HTTPClientConfig.TLSConfig)
+	if err != nil {
+		logger.Error("Error creating TLS config", "err", err)
+		return false
+	}
+
+	dialer := websocket.Dialer{
```
electron0zero, line 86: we allow users to configure the ip protocal and fallback via `preferred_ip_protocol` and `ip_protocol_fallback` in other probers, please add the same for this prober as well.

users should be able to configure ipv4, ipv6, etc.

config/config.go
```diff
@@ -399,6 +400,12 @@ type DNSRRValidator struct {
 	FailIfNoneMatchesRegexp []string `yaml:"fail_if_none_matches_regexp,omitempty"`
 }
 
+type WebsocketProbe struct {
+	HTTPClientConfig config.HTTPClientConfig `yaml:"http_config,omitempty"`
+	Headers          map[string]string       `yaml:"headers,omitempty"`
```
electron0zero: we should either use Headers type from https://github.com/prometheus/common/blob/28b645dd96866da2c1f6b2ba7acd9c7271732332/config/headers.go#L53 as it has support for paintext and secret headers both.

or you can make it a `map[string]Secret` so we don't leak the secrets in the UI and logs.

we are taking the config and building the headers so in most cases there headers will contain secrets.

prober/websocket.go
```diff
@@ -0,0 +1,192 @@
...
+		TLSClientConfig: tlsConfig,
+	}
+
+	connection, resp, err := dialer.DialContext(ctx, targetURL.String(), constructHeadersFromConfig(module.Websocket, logger))
```
electron0zero, line 106: can we add `probe_websocket_ssl` if the connection is made over ssl, we have similar metrics in other probers so makes sense to follow the lead and expose these metrics.

prober/websocket.go
adnull: implemented

prober/websocket.go
adnull: Maybe we add it in another PR? Would be great to implement it together with the http probe by creating standart approach and not duplicate the code. Like we do smth like that
```
TLSMetrics := registerTLSMetrics(registry)
....
TLSMetrics.setFromConn(tls.Conn)
```

adnull:
@electron0zero could you take a look at the changes?

bilkoua:
@adnull Thanks, great work. Could you also take care of the DCO?

adnull:
@bilkoua thanks. done

prober/websocket.go
electron0zero: cool, I am good with follow up PR.

go.mod
```diff
@@ -21,7 +22,10 @@ require (
 require (
 	github.com/golang-jwt/jwt/v5 v5.3.0 // indirect
 	github.com/google/uuid v1.6.0 // indirect
+	golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842 // indirect
```
electron0zero, line 25: bit confused by this? why are we pulling this? 

asking because https://pkg.go.dev/golang.org/x/exp#section-readme says following:

> This subrepository holds experimental and deprecated (in the old directory) packages.

we don't want both experimental and deprecated packages as dependency.

electron0zero:
we are almost there and ready to merge. 

can you rebase on master and added `websocket` as valid prober in the config.go, we merged https://github.com/prometheus/blackbox_exporter/pull/1521 and it adds validation for prober names.

go.mod
electron0zero: false alarm, I didn't notice that this was already there as an indirect dependency was just reordered.

adnull:
@electron0zero sure, seems completed.

prober/websocket.go
```diff
@@ -0,0 +1,253 @@
...
+
+	isConnected.Set(1)
+
+	if len(module.Websocket.QueryResponse) > 0 {
+		transferStart := time.Now()
+		for _, qr := range module.Websocket.QueryResponse {
+			if !matchQueryResponse(qr, connection, logger) {
+				probeFailedDueToRegex.Set(1)
+				durationGaugeVec.WithLabelValues("transfer").Add(time.Since(transferStart).Seconds())
+				return true
```
electron0zero: we should return false here? or otherwise we will set `probe_success` to 0 instead of setting it to 1.

check `probeSuccessGauge` logic in the handler.go to understand how we use the returned value from the `Probers`

electron0zero: also, we should add a test case to make sure we don't regress. we can check for `probe_success` metric in tests to test this.

adnull: Do we consider the prober to be unsuccessfull in that case? As only a part of prober fails.

config/config.go
```diff
@@ -596,6 +619,16 @@ func (s *HeaderMatch) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (s *WebsocketProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type plain WebsocketProbe
```
electron0zero, line 625:
```suggestion
		*s = DefaultWebsocketProbe
		type plain WebsocketProbe
```

without this our default config won't be unmarshalled. we do the same in other probers as well.

prober/websocket.go
```diff
@@ -0,0 +1,253 @@
...
+			b, err := os.ReadFile(file)
+			if err != nil {
+				logger.Error("Unable to read header file", "file", file, "err", err)
+				continue
+			}
+			headers.Add(headerName, strings.TrimSpace(string(b)))
+		}
+	}
+
+	logger.Debug("Constructed headers", "headers", headers)
```
electron0zero: this log line will leak secrets, `headers` is a plain map here, if we need to log the headers, we should log the websocketConfig.Headers to make sure we don't leak secrets in logs.

prober/websocket.go
electron0zero: yes, because our regex check didn't pass, the probe is considered failed.

think of like this: you expect `abc` from server so you configure a regex to check for it, but the server returned data, but it was not `abc`. so connection was successful, it returned data but not the data we expect.


Think of it like HTTP returning 200 but with wrong body, connection worked, but validation failed, which means that this is a probe failure.

check other probers for reference.

users use `probe_success` for alerting so if we don't return this, folks won't get alerts when the probe failed.

`probe_websocket_failed_due_to_regex` is more of a debugging metric to know why the probe failed.

prober/websocket.go
adnull: we don't need it

prober/websocket.go
adnull: updated the code and tests

electron0zero (approved):
lgtm

electron0zero:
I will wait for few weeks to give time to other maintainers to review it.

bilkoua:
@electron0zero @mem @SuperQ Hi guys
Could you take a look at this PR and merging if all is fine? We’ve been waiting quite a while;)

bilkoua:
@electron0zero @mem @SuperQ ping

bilkoua:
@adnull @electron0zero Thanks, guys, for all your hard work - it’s finally here.
@electron0zero Would it be possible to ship a release so we can start using this functionality?

veso0308:
Is it possible to bump a new version which includes the websocket prober?

PR #1283, opened 2024-08-21
Use yaml.v3 to drop yaml.v2 direct dependency

mikelolasagasti:


electron0zero:
can you rebase and resolve conflicts, and add a meaningful PR description with what? and why? for this change.

prober/handler.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/expfmt"
 	"github.com/prometheus/common/promslog"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
```
electron0zero, line 33: I see @SuperQ's comment about blockers around upgrading to v3. https://github.com/go-yaml/yaml/issues/665#issuecomment-1145736596

so I will hold and wait for @SuperQ's review on this before we merge this.

electron0zero:
I would love to get this in and get the new features like comments, but looks like this is bit more involved then just updating the dependency.

There is a behaviour change, and I am afraid that might break people's config. If and when we do this, we will need to either verify that this is not breaking config for users, and if it's breaking, we will need to give advance notice.

also see:
- https://github.com/prometheus/prometheus/pull/10786#issue-1255209958
- https://github.com/go-yaml/yaml/issues/665#issuecomment-1145736596
- https://github.com/prometheus/prometheus/pull/14886#issuecomment-2500513906

electron0zero:
The upstream of `gopkg.in/yaml.v3` has been marked unmaintained. https://github.com/go-yaml/yaml/commit/944c86a7d29391925ed6ac33bee98a0516f1287a See more discussion at https://github.com/prometheus/prometheus/issues/16415, 

Prometheus projects are [moving](https://github.com/prometheus/prometheus/pull/17151) from `gopkg.in/yaml.v2` to `go.yaml.in/yaml/v2`. `go.yaml.in/yaml/v2` is a drop-in replacement for gopkg.in/yaml.v2.

I will keep PR open, but it needs to be updated because we moved to `go.yaml.in/yaml/*` in https://github.com/prometheus/blackbox_exporter/pull/1470

PR #1284, opened 2024-08-21
Add a way to export labels with content matched by the probe

lapo-luchini:
I wanted to create an exporter to check if all my hosts had an updated `sshd` (due to recent CVEs) then I resorted to using blackbock exporter by matching the "fixed" version and the timeout:

```yaml
  ssh_freebsd_secure:
    prober: tcp
    timeout: 5s
    tcp:
      query_response:
      - expect: "^SSH-2.0-OpenSSH_[0-9.]+ FreeBSD-20240806$"
```

but then again I'd prefer to capture the value as a label and do my checks by coloring the value in Grafana table later on or in alertmanager rules so I did this change instead.

I tried to do it in a generic way so that it can be used also for very different cases when matching part of the protocol chat is useful.

This produces scraping values like these:

```
# HELP probe_content Explicit content matched
# TYPE probe_content gauge
probe_content{ssh_comments="FreeBSD-20240806",ssh_version="OpenSSH_9.7"} 1
```

BTW: I know no Go, except a advent-of-code example, these are the first Go lines that I wrote.

Feel free to close if uninteresting for the project (though I hope it isn't, to avoid the need to always compile my branch) or suggest what I should change in the PR to have it accepted. Thanks!

### DIFF
--- a/prober/tcp.go
+++ b/prober/tcp.go
@@ -158,6 +158,23 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 			}
 			probeFailedDueToRegex.Set(0)
 			send = string(qr.Expect.Regexp.Expand(nil, []byte(send), scanner.Bytes(), match))
+			if qr.Labels != nil {
+				var names []string
+				var values []string
+				for _, s := range qr.Labels {
+					names = append(names, s.Name)
+					values = append(values, string(qr.Expect.Regexp.Expand(nil, []byte(s.Value), scanner.Bytes(), match)))
+				}
+				probeContent := prometheus.NewGaugeVec(
+					prometheus.GaugeOpts{
+						Name: "probe_content",
+						Help: "Explicit content matched",
+					},
+					names,
+				)
+				registry.MustRegister(probeContent)
+				probeContent.WithLabelValues(values...).Set(1)
+			}
 		}
 		if send != "" {
 			level.Debug(logger).Log("msg", "Sending line", "line", send)
--- a/prober/tcp.go
+++ b/prober/tcp.go
@@ -165,15 +165,15 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 					names = append(names, s.Name)
 					values = append(values, string(qr.Expect.Regexp.Expand(nil, []byte(s.Value), scanner.Bytes(), match)))
 				}
-				probeContent := prometheus.NewGaugeVec(
+				probeExpectInfo := prometheus.NewGaugeVec(
 					prometheus.GaugeOpts{
-						Name: "probe_content",
+						Name: "probe_expect_info",
 						Help: "Explicit content matched",
 					},
 					names,
 				)
-				registry.MustRegister(probeContent)
-				probeContent.WithLabelValues(values...).Set(1)
+				registry.MustRegister(probeExpectInfo)
+				probeExpectInfo.WithLabelValues(values...).Set(1)
 			}
 		}
 		if send != "" {
--- a/prober/tcp.go
+++ b/prober/tcp.go
@@ -88,12 +88,12 @@ func dialTCP(ctx context.Context, target string, module config.Module, registry 
 	return tls.DialWithDialer(dialer, dialProtocol, dialTarget, tlsConfig)
 }
 
-func probeExpectInfo(registry *prometheus.Registry, qr *config.QueryResponse, scanner *bufio.Scanner, match []int) {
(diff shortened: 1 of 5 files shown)


lapo-luchini:
I wonder if calling `prometheus.NewGaugeVec` inside the loop is fine, normally in a language that I know I would do it only once during configuration parsing, but I'm not sure how to do that in Go.
Doesn't seems to produce problems anyways.

electron0zero:
Hey, this opens up blackbox_exporter to export data out from the check results, and that feel like something that we should not be doing because it can be used to exfiltrate data using blackbox_exporter.

I share the same concern that @SuperQ [mentioned here]( https://github.com/prometheus/blackbox_exporter/pull/1098#issuecomment-1638828966)


This use-case feel like more like something that the should be solved by the software itself by exporting the version label in `build_info` metric. or with https://github.com/prometheus/node_exporter `processes` or `Textfile Collector` of the node_exporter. I also recommend look into something like osquery for this use-case.

lapo-luchini:
Those are valid concerns. On the other hand, an attacker would need local access to modify the yaml file to exfiltrate anything but what was "explicitly matched" by the template, an attacker with only network access to blackbox_exporter couldn't do much (but it could use remote targets to query server outside the expected domain, that's true).

A node_exporter textfile collector would be another solution, right.
(but it would require a new software to be installed on each and every host, not only on one centralized host)

I didn't know of [osquery](https://www.osquery.io/), looks interesting, thanks!

I'll close this PR for now, maybe I'll write a "security-oriented" more all-around exporter from scratch.

SuperQ:
I'm not too concerned about the data exposure here. Like @lapo-luchini says, the exposure values would be up to the admin's regexp.

The concern #1098 and similar are the ability to modify the behavior of the probe from the URL params.

Since this proposal is entirely contained in the config file, it's safe enough. I think this is actually a reasonably good idea.

electron0zero:
> exposure values would be up to the admin's regexp.

ah I see, I think then it's a non- issue.

> (but it would require a new software to be installed on each and every host, not only on one centralized host)

that's true, I was under the assumption that node_exporter is already running to monitor these nodes. which might not be the case for everyone so my assumption doesn't hold true. :)

let me find some time to give it a through review :)

lapo-luchini:
> that's true, I was under the assumption that node_exporter is already running to monitor these nodes. which might not be the case for everyone so my assumption doesn't hold true. :)

It is my case too, but it does not export the running ssh version, so it would need "something else" to add that to a textfile (to the bare minimum, a shell script in cron).

Anyways… that's just my use-case, returning to the feature proposal of this PR (which tries to be useful to use-cases different than mine too) I'll wait for a review and make adjustments if needed. Thanks!

prober/tcp.go
```diff
@@ -158,6 +158,23 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
...
+			if qr.Labels != nil {
+				var names []string
+				var values []string
+				for _, s := range qr.Labels {
+					names = append(names, s.Name)
+					values = append(values, string(qr.Expect.Regexp.Expand(nil, []byte(s.Value), scanner.Bytes(), match)))
+				}
+				probeContent := prometheus.NewGaugeVec(
+					prometheus.GaugeOpts{
+						Name: "probe_content",
```
SuperQ, line 170: Naming nit:

```suggestion
						Name: "probe_content_info",
```

SuperQ (changes requested):

lapo-luchini:
I renamed the metric as requested, and also added some documentation.

prober/tcp.go
```diff
@@ -158,6 +158,23 @@ func ProbeTCP(ctx context.Context, target string, module config.Module, registry
 			}
 			probeFailedDueToRegex.Set(0)
 			send = string(qr.Expect.Regexp.Expand(nil, []byte(send), scanner.Bytes(), match))
+			if qr.Labels != nil {
```
electron0zero, line 161: can you take this logic out into a function and call that in the ProbeTCP?

please add some tests for this logic (maybe one in ProbeTCP test and then add some unit tests for the extracted func.)

electron0zero (changes requested):
needs some tests, and then we are good to merge.

prober/tcp.go
lapo-luchini: Refactored. I hope I used references correctly.
(I was good at C++ some 20 years ago, but this is basically my first Go `func`)

lapo-luchini:
Will do tests tomorrow.

lapo-luchini:
OK, should be good for review!

electron0zero (approved):

PR #1309, opened 2024-10-16
config: add json tags to config structs

### LINKED ISSUE #1308: Feature: Add json Tags
## Issue 
Currently the blackbox exporter only has yaml tags set for the config fields. I would like to reuse the structs for an internal kubernetes operator where the probes are custom resource definitions. This would allow us to dynamically add new probes to the blackbox config.

Additionally once the json tags are added, it might be possible to also add support for configs in json format.

Jakob3xD:
This adds json tags to config struct fields. This is needed for us to reuse the structs in an internal kubernetes operator.

If just adding the tags is unwanted, I can also look into adding general support for json configs. We would need to add UnmarshalJSON functions like the UnmarshalYAML functions.

Used SED to add the json tags. `sed -i 's/\(yaml:"\(.*\)"\)/\1 json:"\2"/g' config/config.go`

Closes #1308


config/config.go
```diff
@@ -193,104 +193,104 @@ func MustNewRegexp(s string) Regexp {
...
 }
 
 type ICMPProbe struct {
-	IPProtocol         string `yaml:"preferred_ip_protocol,omitempty"` // Defaults to "ip6".
-	IPProtocolFallback bool   `yaml:"ip_protocol_fallback,omitempty"`
-	SourceIPAddress    string `yaml:"source_ip_address,omitempty"`
-	PayloadSize        int    `yaml:"payload_size,omitempty"`
-	DontFragment       bool   `yaml:"dont_fragment,omitempty"`
-	TTL                int    `yaml:"ttl,omitempty"`
+	IPProtocol         string `yaml:"preferred_ip_protocol,omitempty"` // Defaults to "ip6" json:"preferred_ip_protocol,omitempty"` // Defaults to "ip6".
```
electron0zero, line 264: comment is now in between tags, happens at few other places as well.

```suggestion
	IPProtocol         string `yaml:"preferred_ip_protocol,omitempty" json:"preferred_ip_protocol,omitempty"` // Defaults to "ip6".
```

electron0zero:
> If just adding the tags is unwanted, I can also look into adding general support for json configs. We would need to add UnmarshalJSON functions like the UnmarshalYAML functions.

thank you for sending this in, just tags are okay.

I am not sure if we want to add support for json configs, I don't see a big reason to add support for json configs.

config/config.go
Jakob3xD: Fixed and adjusted the SED I used.

Jakob3xD:
> thank you for sending this in, just tags are okay.

Perfect, thank you.

electron0zero (approved):
lgtm, will wait for a week or so to give some time to other maintainers before merging it.

Jakob3xD:
Just adding the json tags won't work as it also needs custom Unmarshal functions. As the json support is not wanted? I might close the PR and do the changes in my fork :/
(Currently still working on the custom Unmarshal functions. Once it is done I would open an PR and you can decide if you want it or not?)

electron0zero:
> also needs custom Unmarshal functions

we don't mind having the JSON Unmarshal functions, it would be useful to folks using Blackbox exporter as a library and build on top of Blackbox exporter like you are doing

What I meant was the full fledged support for configuring blackbox exporter via a json config file instead of YAML, in that side, will stick with YAML only.

Jakob3xD:
First step is done with the custom Unmarshal functions. I might need to add some Marshal functions as well.
Can already tell that the current json lib is a pain as it does not support `inline` fields.

If wished I can also generate the json test cases during the test by converting the original yml test case to a json file.
Would remove the "duplicate" test data.

electron0zero:
> If wished I can also generate the json test cases during the test by converting the original yml test case to a json file.
> Would remove the "duplicate" test data.

yes, that would be better then the duplicate config files.

electron0zero:
@Jakob3xD checking in on this PR, we would be open to merge this if you can update and rebase it.

electron0zero:
@Jakob3xD following up on this, I am open to merge this if this can be rebased and updated to include changes from master. 🙏🏼

Jakob3xD:
@electron0zero sorry for the late response. I´ll try to have a look at this in the next week to get this done.

PR #1322, opened 2024-11-14
fix: set the timeout based on the caller to avoid strange issues

zjj:
although a ctx with deadline is set in most cases, but passing in context.TODO() can lead to some very strange problems.

or would it be more appropriate for the business logic to return an error indicating that the context (ctx) has no deadline if it detects the the ctx without setting deadline?


zjj:
these two commits are on different topic that fixes.

PR #1330, opened 2024-12-03
Document difference between TLS Cert expiry metrics

slrtbtfs:
blackbox-exporter currently offers two metrics to
measure when TLS Certificates will expire.

The difference between those is very subtle, but
using `probe_ssl_earliest_cert_expiry`
for checking whether a certificate is due to
replacement can lead to false positive alerts.

This documents the difference between those two.

Generally, `probe_ssl_last_chain_expiry_timestamp_seconds`
 seems to be what most people would want to use.

### DIFF
--- a/prober/prober.go
+++ b/prober/prober.go
@@ -26,7 +26,7 @@ type ProbeFn func(ctx context.Context, target string, config config.Module, regi
 
 const (
 	helpSSLEarliestCertExpiry     = "Earliest expiry time of any TLS certificate returned by the server"
-	helpSSLChainExpiryInTimeStamp = "Earliest expiry of any TLS certificate that is part of the certificate chain used for validating the servers identity"
+	helpSSLChainExpiryInTimeStamp = "Earliest expiry time in the TLS certificate chain used for validating the servers identity"
 	helpProbeTLSInfo              = "Returns the TLS version used or NaN when unknown"
 	helpProbeTLSCipher            = "Returns the TLS cipher negotiated during handshake"
 )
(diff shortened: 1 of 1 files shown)


slrtbtfs:
Hm, the CircleCI Tests seem to be failing for reasons unrelated to this PR:

```
level=error msg="Resolution with IP protocol failed" target=ipv6.google.com ip_protocol=ip4 err="lookup ipv6.google.com on 10.89.4.1:53: no such host"
```

Looks like the CI Test environment is having some network Issues.

Locally, the tests run fine.

slrtbtfs:
CI is passing now after a rebase, so this PR is ready.

prober/prober.go
```diff
@@ -25,8 +25,8 @@ import (
 type ProbeFn func(ctx context.Context, target string, config config.Module, registry *prometheus.Registry, logger *slog.Logger) bool
 
 const (
-	helpSSLEarliestCertExpiry     = "Returns last SSL chain expiry in unixtime"
-	helpSSLChainExpiryInTimeStamp = "Returns last SSL chain expiry in timestamp"
+	helpSSLEarliestCertExpiry     = "Returns the earliest expiry of any peer certificate returned by the server as an unix timestamp. This can include certificates that are not validated by TLS clients. In rare server configurations this might return a time in the past, even for valid TLS certificate chains."
```
electron0zero: this feels bit much for the help, can you trim it a bit and make it less verbose?

same for the example below.

if you wish to add more details, please add a section about SSL metrics in README.md to document these in details.

slrtbtfs: descriptions have been shortened now

ArthurSens (changes requested):
Please remove the binary that you've included in this change

ArthurSens:
I'm taking the liberty of closing the PR since there have been no responses for a while. Feel free to open a new PR without the binary if you're still interested in making this change

PR #1337, opened 2024-12-22
chore: ignore .idea folder created by JetBrains IDEs

electron0zero:
.idea folder is created by JetBrains IDEs and should be ignored. 

I was using JetBrains GoLand, and saw that it's not ignored yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ dependencies-stamp
 .deps
 *.tar.gz
 /vendor
+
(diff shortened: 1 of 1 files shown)


SuperQ:
Please don't include your local editor ignores in the repo. Use [`git config --global core.excludesFile`](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/).

Issue #1120, opened 2023-09-08
grpc reports probe_ssl_earliest_cert_expiry even with disabled tls


Lusitaniae:
```
uname -a 
Linux blackbox-node1 5.4.0-121-generic #137-Ubuntu SMP Wed Jun 15 13:33:07 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

blackbox_exporter --version
blackbox_exporter, version 0.24.0 (branch: HEAD, revision: 0b0467473916fd9e8526e2635c2a0b1c56011dff)
  build user:       root@e5bbfcc8184e
  build date:       20230516-11:07:25
  go version:       go1.20.4
  platform:         linux/amd64
  tags:             netgo
```

```
  grpc:
    prober: grpc
    timeout: 8s
    grpc:
        preferred_ip_protocol: ip4
        tls: false
```

```
  - job_name: blackbox_grpc
    honor_labels: true
    honor_timestamps: true
    metrics_path: /probe
    scheme: http
    scrape_interval: 30s
    scrape_timeout: 10s
    params:
      module: [grpc]
    static_configs:
      - targets:
          - http://subdomain1.example.com
    relabel_configs:
    - source_labels: [__address__]
      target_label: __param_target
    - source_labels: [__param_target]
      target_label: instance
    - target_label: __address__
      replacement: 1.3.5.9:9115 # blackbox-node1

```

```
curl "localhost:9115/probe?module=grpc&target=http%3A%2F%2Fsubdomain.example.com&debug=true"
Logs for the probe:
ts=2023-09-08T10:01:58.168088079Z caller=main.go:352 module=grpc target=http://subdomain.example.com level=info msg="Beginning probe" probe=grpc timeout_seconds=8
ts=2023-09-08T10:01:58.168346846Z caller=grpc.go:146 module=grpc target=http://subdomain.example.com level=info msg="Resolving target address" target=subdomain.example.com ip_protocol=ip4
ts=2023-09-08T10:01:58.169988786Z caller=grpc.go:146 module=grpc target=http://subdomain.example.com level=info msg="Resolved target address" target=subdomain.example.com ip=3.4.5.6
ts=2023-09-08T10:01:58.170049622Z caller=main.go:144 module=grpc target=http://subdomain.example.com level=debug msg="Dialing GRPC without TLS"
[...]

PR #1338, opened 2024-12-23
Only register grpc TLS metrics on successful handshake

mattb18:
In the current gRPC implementation, TLS metrics are registered before the handshake is complete. In cases where the probe fails, **probeSSLEarliestCertExpiryGauge** is set to the default `time.Time` value (far in the past), resulting in any alerts using `probe_ssl_earliest_cert_expiry` metric firing.

In the case of the TLS handshake failing, it makes more sense for TLS metrics not to be set. This is the way the `http` probe is currently configured: https://github.com/prometheus/blackbox_exporter/blob/7e25c6f88f55dd03a890cab7774340034549f6d7/prober/http.go#L645

This should also solve https://github.com/prometheus/blackbox_exporter/issues/1120.


mattb18:
Seems like the gRPC tests seem to be failing on all recent merge requests? I'm not able to reproduce this failure locally.

prober/grpc.go
```diff
@@ -203,6 +200,7 @@ func ProbeGRPC(ctx context.Context, target string, module config.Module, registr
 	if serverPeer != nil {
 		tlsInfo, tlsOk := serverPeer.AuthInfo.(credentials.TLSInfo)
 		if tlsOk {
+			registry.MustRegister(probeSSLEarliestCertExpiryGauge, probeTLSVersion, probeSSLLastInformation)
```
electron0zero, line 203: can we get a test for this, it will ensure we don't break this in future again.

mattb18: @electron0zero have you got any recommendations on how to achieve this? There doesn't seem to be an equivalent for the http probe.

I could create a new helper function that checks for the absence of registry results? Basically the inverse of this: 
https://github.com/prometheus/blackbox_exporter/blob/master/prober/utils_test.go#L35

electron0zero: maybe checkMetrics can be used here? seehttps://github.com/prometheus/blackbox_exporter/blob/master/prober/utils_test.go#L173

if not, feel free to add a new helper to check for absence of a metric?

mattb18: @electron0zero I've added a helper function and a corresponding test, let me know if there's anything else required.

electron0zero (approved):

PR #1342, opened 2024-12-24
Skip failing IPv6 tests in test CI Pipeline

electron0zero:
These tests have been failing on the `test` pipeline CI for a while, but they run fine locally on the systems that support ipv4 and ipv6

Since we have `test-ipv6` pipeline that runs full test suite in a docker image with IPv6 enabled,
we should be okay to skip these Tests on the `test` pipeline.

The reason these tests are failing is because they are using `net.Listen("tcp", "localhost:0")`,
the use of `tcp` network expects both IPv4 and IPv6.

If we don't want to skip these tests, we can split these tests into a table test with `tcp4` and `tcp6`, and only skip `tcp6` on the `test` pipeline.

this was the first approach that came to my mind, and I am open to other approaches as well. if you can think of something better, please suggest other approaches as well.

### DIFF
--- a/prober/grpc_test.go
+++ b/prober/grpc_test.go
@@ -35,6 +35,9 @@ import (
 )
 
 func TestGRPCConnection(t *testing.T) {
+	if os.Getenv("CI") == "true" {
+		t.Skip("skipping; CI is failing on ipv6 dns requests")
+	}
 
 	ln, err := net.Listen("tcp", "localhost:0")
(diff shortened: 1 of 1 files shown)


SuperQ:
I wonder if we should simply flatten the testing so that we always have ipv6 enabled in CI.

SuperQ (approved):
Let's do this for now and come up with a better solution later.

PR #1359, opened 2025-01-31
Adding MySQL/MariaDB endpoint support

wfelipew:
Adding support for MySQL/MariaDB endpoints.
This is just an initial proposal, if you agree to add database endpoint support I can also develop it for PostgreSQL.

This prober will execute a configured query and check for an expected result to determine if it's up or not.

Here is config example:
```
  mysql:
    prober: mysql
    timeout: 2s
    mysql:
      username: "monitoring"
      password: "123456"
      port: 3306
      query: "SELECT 1"
      sql_query_response:
      - column: "1"
        value: "1"
```


SuperQ:
This is what database server exporters like the myself_exporter are for.

Issue #1377, opened 2025-03-04
Log spam


Jakob3xD:
Hello,
we updated our blackbox exporter and are now getting a lot of log messages due to the current implementation of slog introduced with https://github.com/prometheus/blackbox_exporter/pull/1311 as it removes the default `none` logging option.

We have tens of thousands of probes running each minute. Some of them are expected to fail but will cause the logger to print an error message. We are now producing over 14k of logs per second.

I can see two possible solutions for this where I prefer the first one.
#### 1. Adjust the log level for some message
Currently, a probe failure is logged with level ERROR. IMO this should be level DEBUG (or WARN). My interpretation of level ERROR is that blackbox exporter itself is having an issue. Like failed to read config file, failed to unmarshal config file, failed binding to port. Things that are real issues. A failing probe can have different reasons, but does not affect the blackbox exporter. 

#### 2. Add back the `none` logging option
Add the `none` option by writing logs to `io.Discard` like in `NewNopLogger()`.
Instead, it would also be possible to add a flag to define where to write the logs. For example stderr, stdout, /dev/null.

mateuszdrab:
I just noticed that too, after discovering a shocking 160 million log lines generated (and ingested into Loki) in a mere week.

SuperQ:
CC @tjhop care to look into this one?

tjhop:
@Jakob3xD I agree with your interpretation of where error should be used. I'm hesitant to re-introduce the `none` option again since it was "non-standard" from the accepted level values/promlog package usage in other prometheus projects. I say that in reference to the `--log.level` flag mostly, since the `--log.prober` flag is unique to this exporter. 

By defaulting the flag to an empty string and improving how things are handled when unset, I believe I was able to restore the previously expected behavior. As mentioned in the linked PR:
> TL;DR:
> Leave --log.prober flag unset and it won't do scrape-level logging

Opinions?

SuperQ:
I agree, I don't think we want to re-introduce the `none` mode.

> Leave --log.prober flag unset and it won't do scrape-level logging

I think this isn't working as intended. The default value for `--log.prober` is `info`. But even when I change `--log.prober=debug`, I'm still seeing it log probes as `info` level.

tjhop:
> I think this isn't working as intended. The default value for --log.prober is info. But even when I change --log.prober=debug, I'm still seeing it log probes as info level.

[...]

PR #1381, opened 2025-03-18
    fix: stop scrape logger spam, refactor scrape logger design

tjhop:
Based on discussion in #1377, slack, etc, it seems unclear what the
intended functionality/behavior of the probe scrape logger was supposed to be.
This is an attempt to align behavior with current user expectations. The
improved defaults and updated configs here should address the log spam
as well.

User facing changes:
- the `--log.prober` flag now sets the _level at which prober scrape
  logs will be emitted at_.
- the `--log.prober` flag now defaults to `debug` level. Because the
  default log level is info, this means that in order to see probe
scrape logs, a user must either:
    - increase the level of the logger via `--log.level` to be >= that
      of `--log.prober` (ie, set both to `debug`)
    - decrease the level of the probe logger via `--log.prober` to be <=
      that of `--log.level` (ie, set both to `info`)

Note:
This approach (emitting all logs from the probe logger at the level
specified) involves hijacking calls to the logger and overriding the
level. This comes with the very obvious side effect that the developers
working on code using the probe scrape logger no longer have control
over the control over the level of the log that gets emitted. Regardless
of whether they use `l.Info()` or `l.Debug()`, it'll still get logged at
the configured log prober level. This approach likely would not be
compatible with some of the other discussions around how probe scrape
logs work, like #1401.

Signed-off-by: TJ Hoplock <t.hoplock@gmail.com>


tjhop:
@SuperQ @electron0zero I believe this is ready for review if you find yourselves with some spare time 🙃

SuperQ (approved):
LGTM.

main.go
```diff
@@ -50,7 +50,7 @@ var (
 	configFile     = kingpin.Flag("config.file", "Blackbox exporter configuration file.").Default("blackbox.yml").String()
 	timeoutOffset  = kingpin.Flag("timeout-offset", "Offset to subtract from timeout in seconds.").Default("0.5").Float64()
 	configCheck    = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
-	logLevelProber = kingpin.Flag("log.prober", "Log level from probe requests. One of: [debug, info, warn, error]").Default("info").String()
+	logLevelProber = kingpin.Flag("log.prober", "Log level from probe requests. One of: [debug, info, warn, error]").Default("debug").String()
```
electron0zero, line 53: can we document the behaviour and how to change the log level of probes and scrape logs with few examples so we have this documented.

reasons to document 
1) so we know how it works and can refer to it.
2) point users to that section of docs when they want to know how to change these logs levels.


other then the docs, rest lgtm

tjhop: @electron0zero I added some docs in the last commit, mind taking another look?

electron0zero (approved):

electron0zero:

paketb0te:
@tjhop I feel like I am misunderstanding the `--log.prober` flag, maybe you can help me understand it better? :sweat_smile: 

In the docs, you provide example output for different combinations of `--log.level` and `--log.prober` (which I think is a great way to illustrate this, btw) - to me it looks like the combination `--log.level=info` and `--log.prober=info` is _more_ verbose than the (apparently default) combination `--log.level=info` and `--log.prober=debug`.

My understanding of log levels is that "increasing" the log level (e.g. going from INFO to DEBUG in this case) should only ever "add" log output, meaning that the log lines emitted at level DEBUG should also include all lines that would be emitted at level INFO (and the log level of those messages would still show up as "INFO").

Is my understanding of log levels incorrect, or am I maybe misunderstanding what we are trying to achieve here?

(I have read the discussion in #1377 but am still unsure...)

paketb0te:
related question: Is it (still?) possible to configure log levels in the config file?
The docs mention this, but I see nothing related in the example config and also the actual code (the module structs) looks like this is not possible.

tentakle:
Now when I run the curl command:  
`curl -s 'http://localhost:9115/probe?module=http_2xx&debug=true&target=https://example.com'`  
it outputs empty logs?  
```
Logs for the probe:  
  
```
We did something strange.

tjhop:
@paketb0te no worries, this is a non-standard feature anyway, and the way it "merges" log streams isn't super intuitive (which is why it's taking us so long to work out exactly how this should work). Here's an analogy/mental model that I find helpful when trying to understand what the scrape probe logger is doing now:

Let's imagine logging is like a railroad system, with an individual railway for each of the log levels: info, warn, error, debug. These are the only railways that are routeable to a destination, and the verbosity levels in the `--log.level` flag control how many of these level railways route to the same "destination", ie the user. 
We also have a 5th railway for the probe logger. The `--log.prober` flag acts as a [railway switch](https://en.wikipedia.org/wiki/Railroad_switch) to route this 5th railway onto one of the other existing tracks -- the leveled railway it joins determines where the logs go. 

Here's a decent text diagram I was able to get of the logical flow that I was able to get gemini to make for me:
```
                                      +------------------+
                                      |   --log.prober   |
[probe logger]----------------------->|      SWITCH      |---+
                                      +------------------+   |
                                                             |
   +---------------------------------------------------------+
   |             |                   |                       |
   v             v                   v                       v
[info] track  [warn] track      [error] track           [debug] track
   |             |                   |                       |
   |             |                   |                       |
   v             v                   v                       v
+-----------------------------------------------------------------+
|                          --log.level                            |
|                  (Controls which tracks are active)             |
+-----------------------------------------------------------------+
   |             |                   |                       |
   |             |                   |                       |
   +-------------+-------------------+-----------------------+---> [FINAL OUTPUT]


```

Does that help to clarify things at all?

tjhop:
@tentakle yea, this whole feature is a tad strange :upside_down_face: 

Good catch, though! I was able to replicate, and I'm working on a fix for use with the `debug` param output. I'll open a PR when it's ready and ask for a patch release. bug fix for the bug fix, heh. Here's to hoping we've uncovered the last of the behavioral quirks

PR #1382, opened 2025-03-18
Adding the metric for certificate start

bseenu:
We have a use case to monitor the certificate start time


bseenu:
@electron0zero can you please review this change, Thanks

PR #1388, opened 2025-03-26
Update golangci-lint and fix linter errors

juho9000:


.github/workflows/golangci-lint.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get -y install libsnmp-dev
         if: github.repository == 'prometheus/snmp_exporter'
       - name: Lint
-        uses: golangci/golangci-lint-action@2226d7cb06a077cd73e56eedd38eecad18e5d837 # v6.5.0
+        uses: golangci/golangci-lint-action@dec74fa03096ff515422f71d18d41307cacde373 # v7.0.0
         with:
           args: --verbose
-          version: v1.64.6
+          version: v2.0.2
```
SuperQ, line 39: This file is synced from prometheus/prometheus. We will have to coordinate from there first.

SuperQ (changes requested):

.github/workflows/golangci-lint.yml
juho9000: Ah, gotcha. :)

juho9000: I guess a contribution would be welcome there for migrating to golangci-lint v2? Let's see if I find some time.

SuperQ: This update has been done.

Issue #801, opened 2021-06-18
Unable to scrape Postgres cert metrics


atmanohar:
### What is the blackbox.yml module config.
    modules:
      tls_connect:
        prober: tcp
        timeout: 60s
        tcp:
          preferred_ip_protocol: "ip4"
          ip_protocol_fallback: false
          query_response:
            - expect: "OK.*STARTTLS"
            - send: ". STARTTLS"
            - expect: "OK"
            - starttls: true
            - send: ". capability"
            - expect: "CAPABILITY IMAP4rev1"
          tls: true
          tls_config:
            insecure_skip_verify: true

      ldap:
        prober: tcp
        timeout: 5s
        tcp:
          preferred_ip_protocol: "ip4"
          query_response:
            - send: !!binary "MCICAQFgHQIBAwQIbm90YXVzZXKADm5vdHRoZXBhc3N3b3Jk"
            - expect: !!binary "MAwCAQFhBw=="

      postgres:
        prober: tcp
        tcp:
          query_response:
            - send: !!binary AAAACATSFi8=
            - starttls: true

### What is the prometheus.yml scrape config.
- job_name: postgres-blackbox
    honor_timestamps: true
    scrape_interval: 5m
    scrape_timeout: 5m
    metrics_path: /probe
    params:
      module: [tcp_connect] ##also tried with modules ldap and postgres

    kubernetes_sd_configs:
    - api_server: null
      role: service
      namespaces:
        names: []
    relabel_configs:
    - source_labels: [__meta_kubernetes_namespace]
      target_label: kubernetes_namespace
    - source_labels: [__meta_kubernetes_service_name]
      target_label: kubernetes_name
    - source_labels: [__address__]
      target_label: __param_target
    - source_labels: [__param_target]
      target_label: instance
    - target_label: __address__
      replacement: prometheus-blackbox-exporter.do-prom-athira.svc.cluster.local:9115
   

### What did you expect to see?

# HELP probe_ssl_earliest_cert_expiry Returns earliest SSL cert expiry date
# TYPE probe_ssl_earliest_cert_expiry gauge
probe_ssl_earliest_cert_expiry 1.655444464e+09
# HELP probe_ssl_last_chain_expiry_timestamp_seconds Returns last SSL chain expiry in unixtime
# TYPE probe_ssl_last_chain_expiry_timestamp_seconds gauge
probe_ssl_last_chain_expiry_timestamp_seconds -6.21355968e+10
[...]

PR #1424, opened 2025-06-02
feat(tcp): add support for PostgreSQL SSLRequest in TCP prober

carnerito:
Adds a new `postgres_ssl_request` option to the TCP module configuration. When enabled, the prober sends a PostgreSQL-specific SSLRequest message before initiating a TLS handshake. This allows the TCP prober to connect to PostgreSQL servers that expect protocol negotiation prior to starting TLS.

This change enables monitoring of PostgreSQL TLS sockets using the standard `tcp` prober.

Reference: https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-SSL

Fixes: #801

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -207,6 +207,9 @@ query_response:
 
 # Whether or not TLS is used when the connection is initiated.
 [ tls: <boolean | default = false> ]
+# If set to `true`, the probe will send a PostgreSQL-specific `SSLRequest` message
+# before starting the TLS handshake.
+[ postgres_tls_request: <boolean | default = false> ]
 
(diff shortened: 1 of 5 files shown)


electron0zero:
Hi, we are looking to get https://github.com/prometheus/blackbox_exporter/pull/1441 over finish line get that merged. once  https://github.com/prometheus/blackbox_exporter/pull/1441 is merged, you should be able to configure blackbox exporter to ping PostgreSQL with TCP prober.

I will close this PR, please feel free to comment or reopen if you think https://github.com/prometheus/blackbox_exporter/pull/1441 will not work for this use-case.

Issue #1418, opened 2025-05-27
Supporting HTTP3 Extender Options - QUIC


sebastianfeliciano:
I am exploring different transport solutions, including **QUIC**, for quicker handshakes. However, we currently lack a way to monitor this synthetically. The exporter supports **HTTP1** and **2** but not **HTTP3**.

**I have a few questions and would appreciate your insights:**
1. Are there any existing solutions or suggestions for synthetic traffic monitoring that you could recommend?
2. Is the _quic-go_ library production-ready, and if not, what are the blockers?
3. Any advice on extending the Blackbox Exporter for this purpose?

electron0zero:
> I have a few questions and would appreciate your insights:

> 1. Are there any existing solutions or suggestions for synthetic traffic monitoring that you could recommend?

nothing comes to mind form top of my head.

> 2. Is the quic-go library production-ready, and if not, what are the blockers?

by looking at the the lib, I see it's used in big projects like caddy but I would checking out the [stdlib quic package](https://pkg.go.dev/golang.org/x/net/quic) and seeing if that's enough for us

> 3. Any advice on extending the Blackbox Exporter for this purpose?

we have http prober, you can extend that to add support for HTTP3, feel free to send a PR and we can find some time to review it.

sebastianfeliciano:
> by looking at the the lib, I see it's used in big projects like caddy but I would checking out the [stdlib quic package](https://pkg.go.dev/golang.org/x/net/quic) and seeing if that's enough for us
> 

I saw this on the library that you suggested: 

`This package is a work in progress. It is not ready for production usage. Its API is subject to change without notice.
`

Is that something we would want to consider because the [quic go library](https://pkg.go.dev/golang.org/x/net@v0.40.0/quic) is, I guess, more 'prod ready'?

 What are your thoughts...

electron0zero:
I am good with stdlib or quic-go, if you feel stdlib is not ready, feel free to go with quic-go.

PR #1425, opened 2025-06-10
 Enabling HTTP/3.0 detection for Prober Extender (QUIC)

sebastianfeliciano:
## Summary

**Why**:  

[Link](https://github.com/prometheus/blackbox_exporter/issues/1418)

We need monitoring capabilities to validate QUIC endpoints. 

**Discussion**: 

The standard Prometheus Blackbox Exporter currently supports HTTP/1 and HTTP/2, but not HTTP/3. We've implemented HTTP/3 support by extending the existing HTTP prober rather than creating a separate prober.

We use `quic-go library`

**Preview**: 

Monitor HTTP/3 endpoints with the same reliability as HTTP/1.x and HTTP/2.

Validate QUIC connectivity in our development pipeline.

Collect metrics on HTTP/3 performance using connection and TLS handshake time.

## Testing
local dev host, tbd dev beach hosts

please look at the [original repo](https://github.com/prometheus/blackbox_exporter) for directions how to spin up

## Revert Plan
revert


sebastianfeliciano:
@electron0zero

zhengxiongzhao:
When the certificate expires
handler.go:194 Error for HTTP request CRYPTO_ERROR 0x12a (local): tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2025-06-23T11:32:17+08:00 is after 2025-06-22T02:06:45Z

sebastianfeliciano:
> When the certificate expires handler.go:194 Error for HTTP request CRYPTO_ERROR 0x12a (local): tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2025-06-23T11:32:17+08:00 is after 2025-06-22T02:06:45Z

@zhengxiongzhao in the Go crypto/tls package, tls.Config.ServerName must not be empty, Based on this [issue](https://github.com/quic-go/quic-go/issues/4049), I've moved logic to handle this. Please note this does not have a fallback method where h3 goes to h2. If there's something else please let me know.

sebastianfeliciano:
@bwplotka could I get a review?

zhengxiongzhao:
> > When the certificate expires handler.go:194 Error for HTTP request CRYPTO_ERROR 0x12a (local): tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2025-06-23T11:32:17+08:00 is after 2025-06-22T02:06:45Z
> 
> @zhengxiongzhao in the Go crypto/tls package, tls.Config.ServerName must not be empty, Based on this [issue](https://github.com/quic-go/quic-go/issues/4049), I've moved logic to handle this. Please note this does not have a fallback method where h3 goes to h2. If there's something else please let me know.

ok. thank you for your answer. I will give it a try

sebastianfeliciano:
@SuperQ can I get a review?

prober/http.go
```diff
@@ -415,17 +427,46 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+	var client *http.Client
+	var noServerName http.RoundTripper
 
-	httpClientConfig.TLSConfig.ServerName = ""
-	noServerName, err := pconfig.NewRoundTripperFromConfig(httpClientConfig, "http_probe", pconfig.WithKeepAlivesDisabled())
-	if err != nil {
-		logger.Error("Error generating HTTP client without ServerName", "err", err)
-		return false
+	if httpConfig.UseHTTP3 {
+		logger.Info("Creating HTTP/3 client")
```
jesusvazquez: I dont see the info message for !http3 so perhaps we can get rid of this one.

Dockerfile
```diff
@@ -10,4 +10,4 @@ COPY blackbox.yml       /etc/blackbox_exporter/config.yml
 
 EXPOSE      9115
 ENTRYPOINT  [ "/bin/blackbox_exporter" ]
-CMD         [ "--config.file=/etc/blackbox_exporter/config.yml" ]
+CMD         [ "--config.file=/etc/blackbox_exporter/config.yml" ]
```
electron0zero: let's add that newline back :)

sebastianfeliciano:
> The code is looking great. Left a small nit about a logline that perhaps was a leftover?
> * TestValidHTTPVersion should now perhaps include  HTTP/3

So TestValidHTTPVersion uses a **H1.1** and **H2** server over **TCP**. The structure of the test is starting with the server with **1.1**, it can connect to **2**. Without **1.1** it cannot connect to **H2** because it is an **H1.1** server.

With that being said, this is done over **TCP** so **TestValidHTTP** should not work for **H3**, I've added a **false** case, but added a whole new **ValidHTTPVersionsQUIC** since this is over UDP and does not fallback using TCP.

blackbox.yml
```diff
@@ -60,3 +60,10 @@ modules:
     timeout: 5s
     icmp:
       ttl: 5
+  http_3xx:
+    prober: http
+    http:
+      preferred_ip_protocol: "ip4"
+      enable_http3: true
+      enable_http2: false
```
electron0zero, line 68: do you always need to disable `enable_http2` when using `enable_http3`? or this is the a example of using http3 only?

thinking about it, I can't think of the case on when you need `enable_http3` and `enable_http2` both? but I might be wrong here.

if it doesn't make sense to have both enabled? let's add a validation so folks are not allowed to enable both?

electron0zero: also, let's add the newline back to the file.

prober/http.go
```diff
@@ -380,6 +386,12 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 		target = "http://" + target
 	}
 
+	// For HTTP/3, ensure HTTPS is used
+	if httpConfig.UseHTTP3 && strings.HasPrefix(target, "http://") {
```
electron0zero, line 390: we should validate this at the config time, and warn users on startup when there's an http target with HTTP3.

prober/http.go
```diff
@@ -380,6 +386,12 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
...
+		target = strings.Replace(target, "http://", "https://", 1)
+		logger.Info("Converting HTTP to HTTPS for HTTP/3 compatibility", "target", target)
```
electron0zero: this should be at warn level because we are changing the target and users should be aware of this change. not all users run at info level.

prober/http_test.go
```diff
@@ -1806,3 +1807,60 @@ func TestBody(t *testing.T) {
...
 	}
 }
+
+func TestValidHTTPVersionsQUIC(t *testing.T) {
+	tests := []struct {
+		ValidHTTPVersions []string
+		ShouldSucceed     bool
+	}{
+		{[]string{"HTTP/1.1", "HTTP/2.0"}, false},
+		{[]string{"HTTP/3.0"}, true},
```
electron0zero, line 1832: let's add a case where `ValidHTTPVersions` are all the versions.

prober/http_test.go
```diff
@@ -87,6 +87,7 @@ func TestValidHTTPVersion(t *testing.T) {
 		{[]string{"HTTP/1.1"}, true},
 		{[]string{"HTTP/1.1", "HTTP/2.0"}, true},
 		{[]string{"HTTP/2.0"}, false},
+		{[]string{"HTTP/3.0"}, false},
```
electron0zero, line 96: let's add a case where ValidHTTPVersions are all the versions.

prober/http_test.go
```diff
@@ -1806,3 +1807,60 @@ func TestBody(t *testing.T) {
...
+		{[]string{"HTTP/1.1", "HTTP/2.0"}, false},
+		{[]string{"HTTP/3.0"}, true},
+		{[]string{"HTTP/1.1"}, false},
+		{[]string{"HTTP/2.0"}, false},
+		{[]string{}, true},
+	}
+	for i, test := range tests {
+		recorder := httptest.NewRecorder()
+		registry := prometheus.NewRegistry()
+		result := ProbeHTTP(context.Background(), "https://http3.is",
```
electron0zero: I would like us to NOT depend on a third party website to run our unit tests, I prefer that we don't depend on https://http3.is in our unit tests.

you can setup an simple http3 test server, run it in test and probe that, also we will have a better control over that server and we can remove the dependency on `https://http3.is`

also, dependency on https://http3.is means that unit tests are reaching out on the public internet, and can be ran offline.

prober/http_test.go
```diff
@@ -1806,3 +1807,60 @@ func TestBody(t *testing.T) {
...
+			t.Fatalf("Test %v had unexpected result: %s", i, body)
+		}
+	}
+}
+
+func TestHTTP3ProbeQUIC(t *testing.T) {
+	registry := prometheus.NewRegistry()
+	testCTX, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	result := ProbeHTTP(testCTX, "https://http3.is",
```
electron0zero: same here.

sebastianfeliciano:
@electron0zero, thank you for your help! Just to note, Ive added config checking within the ValidHTTPVersions in additon to config.go, so that others do not get confused on why HTTP3 should not work for the TCP Version (With 1.1 and 2.0) and vice-versa. This is because the Test is called on probe and in main it Reloads a config first, i added the validation inside of the tests, and referenced the file.

prober/http_test.go
```diff
@@ -1806,3 +1820,156 @@ func TestBody(t *testing.T) {
...
+	go func() {
+		if err := server.ListenAndServe(); err != nil {
+			if !strings.Contains(err.Error(), "server closed") {
+				serverStarted <- err
+				return
+			}
+		}
+	}()
+
+	time.Sleep(500 * time.Millisecond)
```
electron0zero: nit: do we need sleep here? sleep results in flaky tests, or if it's high enough, extra CI runtime?

my read is that sleep is used to make sure the sever is running before we return from here, see if https://pkg.go.dev/sync#WaitGroup is better suited here and replace the sleep with it.

if it's not possible then I guess we can live with sleep here :)

config/config.go
```diff
@@ -419,6 +425,28 @@ func (s *HTTPProbe) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 	}
 
+	if !s.UseHTTP3 {
+		for _, version := range s.ValidHTTPVersions {
+			if version == "HTTP/3.0" {
+				return errors.New("HTTP/3 cannot be used as a valid HTTP version when enable_http3 is false")
```
electron0zero, line 431: I see two tests for `error parsing config file: when enable_http3 is true, enable_http2 must be set to false` but no test for this error, can you add that as well?

config/testdata/invalid-no-versions-http3-enabled.yml
```diff
@@ -0,0 +1,8 @@
+modules:
+  http_headers:
+    prober: http
+    timeout: 5s
+    http:
+      enable_http3: true
+      preferred_ip_protocol: "ip4" 
+
```
electron0zero: nit: missing newline

blackbox.yml
```diff
@@ -60,3 +60,12 @@ modules:
...
+      valid_http_versions: ["HTTP/3.0"]
+      
+
```
electron0zero: nit: missing newline :)

prober/http_test.go
sebastianfeliciano: thank you!

jesusvazquez (approved):
This is now looking good to me, awesome work 💪 . 

Waiting for @electron0zero's review before merge

electron0zero:
I am giving time to other maintainers to review before we merge, if there's nothing in a week, I will hit merge. 

feel free to remind me if I don't get to it after a week.

sebastianfeliciano:
@electron0zero, looks like there are no answers after a week 😮

PR #1427, opened 2025-06-13
Experiment: Schematized Metrics

sh0rez:
Uses tooling from OTel (https://github.com/open-telemetry/weaver) to strongly schematize all metrics of blackbox_exporter.

This gives:
- Type-safe `client_golang` constructors for all metrics that make are:
    - hard to mis-use (e.g. each label is a different type, to prevent ordering mismatch using the type system)
    - discoverable (IDE auto-complete, godoc)
- Documentation (`internal/metrics/metrics.md`) that lists all metrics, labels and their respective possible values

I chose `blackbox_exporter` as a sufficiently simple target to try this approach and serve as a ground for future discussions around schematizing Prometheus metrics.


semconv/config/metrics.yaml
```diff
@@ -0,0 +1,17 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/open-telemetry/weaver/refs/heads/main/schemas/semconv.schema.json
+groups:
+  - id: metric.config.last.reload.successful
+    type: metric
+    metric_name: blackbox_exporter_config_last_reload_successful
+    brief: "Blackbox exporter config loaded successfully"
+    instrument: "gauge"
+    unit: "1"
```
bwplotka, line 8: You didn't get any warning from Weaver? This is not the correct Otel unit type I think (just curious)

bwplotka (commented):
Looks good!

One thing I would ensure is to embed/import the generation in the Makefile and CI of this project, so those yamls and Go files don't diverge.

bwplotka:
NOTE: Current schema model is not great for changes later on without https://github.com/open-telemetry/opentelemetry-specification/pull/4492 -- so we might want to be careful with committing to a certain yaml model.

semconv/config/metrics.yaml
tombrk: It is a valid unit!

> Instruments for utilization metrics (that measure the fraction out of a total) are dimensionless and SHOULD use the default unit 1 (the unity).
https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units

However, I still think its the wrong unit here, as it should be seconds lol

PR #1438, opened 2025-07-08
fix: make `debug` param log output work regardless of log level

tjhop:
Fixes issue reported here:
https://github.com/prometheus/blackbox_exporter/pull/1381#issuecomment-3044749631

In aligning behavior for how the `--log.prober` flag is currently
expected to work, I didn't test it's effect on logs when used with the
`debug` URL parameter, as is often done for testing.

This changes the handler to always write to the internal buffer logger
(which is what is used to store the logs for displaying with the `debug`
url param), and it only writes through to the real logger if it's
enabled for the level the prober is set to. This ensures that logs are
always available with the `debug` param, while still correctly filtering
levels for console output at stderr.

As a consequence of thinking about how to make this work, I also
realized it wasn't necessary to manually iterate record attrs or
maintain our own attr list when hijacking the log call -- the level can
be overridden directly from the record, we just need to grab a copy via
Clone() first. So should be more efficient, as well.

Signed-off-by: TJ Hoplock <t.hoplock@gmail.com>

### DIFF
--- a/prober/handler.go
+++ b/prober/handler.go
@@ -16,6 +16,7 @@ package prober
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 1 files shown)


tjhop:
cc: @electron0zero @SuperQ

tjhop:
@electron0zero should we get this fix in now while we gather more input for https://github.com/prometheus/blackbox_exporter/issues/1377#issuecomment-3089558497 ?

electron0zero:
my bad for not getting to it, thanks for the PR 🙇🏼

PR #1112, opened 2023-08-18
feat: Implement byte matching in TCP query responses

tacho:
Currently the exporter only supports lines, which breaks byte-oriented protocols such as the PostgreSQL StartTLS handshake.

We also give a working example for Postgres in the sample configuration.

Fixes: https://github.com/prometheus/blackbox_exporter/issues/801

PR #1441, opened 2025-07-16
feat: Implement byte matching in TCP query responses

bitfehler:
Currently the exporter only supports lines, which breaks byte-oriented protocols such as the PostgreSQL StartTLS handshake.

We also give a working example for Postgres in the sample configuration.

This is an attempt to get #1112 over the finish line, as it has apparently been abandoned.

I would be just as happy with #1424, but maybe this is a bit more generic and flexible?


bitfehler:
I tried to address the review comments from #1112 while rebasing. Let me know if I missed anything...

CONFIGURATION.md
```diff
@@ -195,12 +195,14 @@ regexp: <regex>,
...
 # "expect" matches a regular expression;
+# "expect_bytes" does exact byte-by-byte match.
+# "expect" and "expect_bytes" are mutually exclusive.
 # "labels" can define labels which will be exported on metric "probe_expect_info";
 # "send" sends some content;
 # "send" and "labels.value" can contain values matched by "expect" (such as "${1}");
 # "starttls" upgrades TCP connection to TLS.
 query_response:
-  [ - [ [ expect: <string> ],
+  [ - [ [ expect: <string> | expect_bytes: <string> ],
```
electron0zero: we are using `|` to denote the default value of a config in few places in this file,  so this will break that convention.

let's move it into next line so it's more clear. we do something similar for `body_file` config options.

CONFIGURATION.md
```diff
@@ -195,12 +195,14 @@ regexp: <regex>,
 
 # The query sent in the TCP probe and the expected associated response.
 # "expect" matches a regular expression;
+# "expect_bytes" does exact byte-by-byte match.
+# "expect" and "expect_bytes" are mutually exclusive.
```
electron0zero: let's move this to line above, so each line is for one a config param.


I am thinking something like `# "expect_bytes" does exact byte-by-byte match; "expect" and "expect_bytes" are mutually exclusive.`

blackbox.yml
```diff
@@ -67,3 +67,10 @@ modules:
       enable_http3: true
       enable_http2: false
       valid_http_versions: ["HTTP/3.0"]
+  postgresql:
```
electron0zero, line 70: let's add an example in `example.yml` as well, so folks know they can use `expect_bytes` to do byte matching.

bitfehler:
@electron0zero thanks for the review, I hope this is what you had in mind!

prober/query_response.go
```diff
@@ -63,7 +64,12 @@ func probeQueryResponses(ctx context.Context, target string, conn net.Conn, modu
 		Name: "probe_failed_due_to_regex",
 		Help: "Indicates if probe failed due to regex",
 	})
+	probeFailedDueToBytes := prometheus.NewGauge(prometheus.GaugeOpts{
```
electron0zero, line 67: this metric seems to be unused, I couldn't find it anywhere in the codebase. linter is not warning because of `registry.MustRegister(probeFailedDueToBytes)`

can we clean this up?

prober/query_response.go
```diff
@@ -63,7 +64,12 @@ func probeQueryResponses(ctx context.Context, target string, conn net.Conn, modu
...
+		Name: "probe_failed_due_to_bytes",
+		Help: "Indicates if probe failed due to bytes",
+	})
```
electron0zero, line 70:
```suggestion
```

prober/query_response.go
```diff
@@ -63,7 +64,12 @@ func probeQueryResponses(ctx context.Context, target string, conn net.Conn, modu
...
 	registry.MustRegister(probeFailedDueToRegex)
+	registry.MustRegister(probeFailedDueToBytes)
```
electron0zero, line 72:
```suggestion
```

bitfehler: Ah, sorry, I actually just added this, and my intention was to use it [here](https://github.com/prometheus/blackbox_exporter/pull/1441/files#diff-546aa96ba050a3b227235db1fa1253f1b21ce38fd93a3ea31698741190ed027aR153), which I think would be preferable to dropping it again, WDYT?

Alternatively, we could rename the metric to "failed_due_to_expect" or something, I think that would also work. But the current name is just not very fit for this new case...

electron0zero: ah I see, I was thinking that you are planning to keep using `probe_failed_due_to_regex` for expect_bytes as well.

I think a name like `probe_failed_due_to_expect_bytes` would be good name for it? since we are doing a byte match.

I am hesitant to call it `failed_due_to_expect` because the regexp match is called `expect` in config, and might create some confusion among users.

bitfehler: Sorry, I think I only wrote half my thoughts in the previous comment :innocent: 

So, what I originally had in mind is what I pushed now: a new metric in addition to the existing one, and it's called `probe_failed_due_to_bytes`.

What I _meant_ to say in the last comment is that _another_ option _could_ be to use the same metric for both `expect` and `expect_bytes`, in which case it would need a more generic name, though (like `probe_failed_due_to_expect`). 

So the questions are: 

- Use existing metric and rename it, or use new metric? 
- If new metric, is name (`probe_failed_due_to_bytes`) ok?

I hope this makes sense now, sorry again :sweat_smile:

electron0zero: > Use existing metric and rename it, or use new metric?

let's use a new metric, rename would be a breaking change for most users (_most existing users are and will not using the byte matching_).

> If new metric, is name (probe_failed_due_to_bytes) ok?

yup, `probe_failed_due_to_bytes` OR `probe_failed_due_to_expect_bytes` works. I am okay with any of these names.

prober/tcp_test.go
```diff
@@ -580,6 +580,49 @@ func TestTCPConnectionQueryResponseMatching(t *testing.T) {
...
+	registry := prometheus.NewRegistry()
+	if !ProbeTCP(testCTX, ln.Addr().String(), module, registry, promslog.NewNopLogger()) {
+		t.Fatalf("TCP module failed, expected success.")
+	}
+	mfs, err := registry.Gather()
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedResults := map[string]float64{
+		"probe_failed_due_to_regex": 0,
```
electron0zero, line 621: we should check the `probe_failed_due_to_bytes` metric value has well.

bitfehler: I added a check for it here and also a new test case where it fails.

electron0zero:
thank you for pushing it over finish line

PR #1446, opened 2025-07-31
Make default HTTP User-Agent compliant with RFC9110

Hafpaf:
Looking at an NGINX access log I noticed that the default user-agent doesn't follow HTTP Semantics (RFC9110)

RFC9110 section 10.1.5 (User-Agent) defines that a user-agent consists of tokens (section 5.6.2). The space character is not part of the product token, but reserved for other cases, therefore changing the user-agent to `Blackbox-Exporter/<product-version>`,
 Example: `Blackbox-Exporter/0.27.0`.

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -292,7 +292,7 @@ func (bc *byteCounter) Read(p []byte) (int, error) {
 	return n, err
 }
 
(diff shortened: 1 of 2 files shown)


electron0zero (approved):

electron0zero:

PR #1461, opened 2025-09-04
feat: de-couple --log.prober to be an independent logger

tjhop:
This is the beginning of the changes in architecture/behavior for the
scrape probe logger as described here:
https://github.com/prometheus/blackbox_exporter/issues/1377#issuecomment-3089558497

Currently, the scrape probe logger is a weird transforming-wrapper
around the primary logger, that comes with unconventional behavior -- it
hijacks log calls and forces them to a pre-set level. This both
confusing to users (see #1377, #1411, #1401, potentially others), as
well as very inflexible; by forcing logs to a specific level, we lose
the ability to do more nuanced logging like using debug to increase
verbosity or writing errors at error level instead of $configuredLevel.

This commit converts the scrape probe logger to be it's own
full/independent logger via promslog, same as the primary logger. Rather
than wrapping and munging the log entries, we use the primary
promslog.Config as the basis of the config for the new scrape probe
logger. We override the level field to reflect the level var that
corresponds to the `--log.prober` flag, and then use that config to
create the scrape probe logger.

**NOTES**:
- ~This intentionally leaves the `--log.prober` flag default to `debug`
  for now -- most probers write logs at `info` level, which would result
in more log spamming for users. In order for the blackbox exporter to
truly take advantage of these changes, the probers will need to be
updated to correctly use leveled logging.~ Edit: This has been updated in later commits. The `--log.prober` flag now defaults to `info` like a normal logger, and the probers have been updated to better use leveled logging.
- Since we now use v0.65.0 for prometheus/common, I removed the
  `getSlogLevel` helper func in favor of getting the level directly from
the config.

Example debug output for a scrape:

```
~/go/src/github.com/prometheus/prometheus (main [  ]) -> curl -sL "http://localhost:9115/probe?target=prometheus.io&module=http_2xx&debug=true" | head -20
Logs for the probe:
time=2025-09-04T12:07:05.573-04:00 level=INFO source=handler.go:116 msg="Beginning probe" module=http_2xx target=prometheus.io probe=http timeout_seconds=119.5
time=2025-09-04T12:07:05.574-04:00 level=INFO source=utils.go:61 msg="Resolving target address" module=http_2xx target=prometheus.io target=prometheus.io ip_protocol=ip4
time=2025-09-04T12:07:05.574-04:00 level=INFO source=utils.go:96 msg="Resolved target address" module=http_2xx target=prometheus.io target=prometheus.io ip=172.67.201.240
[...]

PR #1464, opened 2025-09-12
Support reload config automatically

### LINKED ISSUE #1416: Support automatically reload config
Related to https://github.com/prometheus/prometheus/issues/9783

It would be really useful to have this feature available in the nodes (at least the official ones). It would avoid having to set up mechanisms that vary depending on the installation method during deployments.

endesapt:
Introduces automatic reloading of the Blackbox exporter configuration file at a specified interval.

- `--config.auto-reload-interval`: Interval to check for config changes in seconds (default: `30`).
- `--config.enable-auto-reload`: Enables the auto-reload functionality.

Uses checksum-based detection like in Prometheus

Fixes #1416

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Changes:
 * [FEATURE]
 * [ENHANCEMENT]
 * [BUGFIX]
+* [FEATURE] Support config reload automatically - feature flag `config.enable-auto-reload`, `config.auto-reload-interval`
+* [CHANGE] Config is not reloaded if the file content didn't change when using all reload methods
 
 ## 0.27.0 / 2025-06-26
 


main.go
```diff
@@ -47,14 +48,16 @@ import (
...
-	routePrefix    = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").PlaceHolder("<path>").String()
-	toolkitFlags   = webflag.AddFlags(kingpin.CommandLine, ":9115")
+	configFile         = kingpin.Flag("config.file", "Blackbox exporter configuration file.").Default("blackbox.yml").String()
+	timeoutOffset      = kingpin.Flag("timeout-offset", "Offset to subtract from timeout in seconds.").Default("0.5").Float64()
+	configCheck        = kingpin.Flag("config.check", "If true validate the config file and then exit.").Default().Bool()
+	enableAutoReload   = kingpin.Flag("config.enable-auto-reload", "When enabled, Blackbox exporter will automatically reload its configuration file at a specified interval. The interval is defined by the `--config.auto-reload-interval` flag, which defaults to `30s`").Default().Bool()
```
electron0zero, line 55: can you document this in the https://github.com/prometheus/blackbox_exporter#configuration section?

endesapt: yes, sure. I added description to README.md as you said, is it enough?

main.go
```diff
@@ -139,6 +150,12 @@ func run() int {
 				if err := sc.ReloadConfig(*configFile, logger); err != nil {
 					logger.Error("Error reloading config", "err", err)
 					continue
+				} else if *enableAutoReload {
```
electron0zero: this deep nesting of if else within a case feels little untidy, we should clean this up.

one way I was thinking was having the checksum in the SafeConfig, and then Reload Config takes care of computing the latest checksum and saving it in the SafeConfig, if we are doing automatic reloads.

that way this doesn't need to be done next to each call to `.ReloadConfig.`, and code would be lot cleaner.

endesapt: Okay, i can do that. I Can just move checksum and logic inside ReloadConfig function (it will always recalculate checksum). But as i see it i will work like that:
1. it checks for checksum
2. if it is equal to previois checksum -> skip
3. reload config
So if the config didn't change, all the reload methods will not reload configuration
I think its great and will also optimize reloading using hup signal and web, but it will change behaviour that we have now, previously it would reload config anyways (for hup and web) and AFTER THAT recalculate checksum.

I dont see it as a problem, but i think i need to write about it in case my way of implementing this is not ideal.

electron0zero: we should able to gate the checksum and auto-reload logic with `if *enableAutoReload` to compute checksum only when autoreload is enabled and leave the existing reload logic unchanged?

electron0zero: also, I see we are logging that "config not changed" message, so as long as we add that behaviour change into the change log, I am good with the behaviour change as well.

endesapt: > we should able to gate the checksum and auto-reload logic with `if *enableAutoReload` to compute checksum only when autoreload is enabled and leave the existing reload logic unchanged?

I can do that, but then i need to pass enableAutoReload to every configReload function to gate the logic accordingly, including editing tests. I guess that it is better to make as few changes as possible



> also, I see we are logging that "config not changed" message, so as long as we add that behavior change into the change log, I am good with the behavior change as well.

So you are satisfacted with the logic we have now? I dont understand what i should do now?
Do you want me to gate the logic as you said in your first message, or should i add something to acknowledge that the behavior of  configReload has changed?

In case you want my opinion, as i said earlier, i think that the logic we have now is clean, it is transparently adding the logic that is really should have been existed for all the reload methods, and it doesnt ruin or change drastically any interactions users have.

electron0zero: yeah, I thought about it a bit and logic is fine. reloading a config with no changes is same as not reloading it so behaviour change is not an issue.

Let's just add a line in the `CHANGELOG.md` about the change and behaviour change, so we have it in the release notes when we cut a new release, and then we are good.

endesapt: srry i am a little bit confused, should i add anything to CHANGELOG.md, or will you do it when u will be creating MR for new release? as i can see you have Changelog  sorted per release... This is my first time contributing, so i dont know how it works actually.

endesapt: @electron0zero sorry to bother you, but it would be sad if this wont make it to next release because i did smth wrong(

electron0zero: @Endesapt It be great if you can add it, see https://github.com/prometheus/blackbox_exporter/pull/1461/ as an example.

this is first PR where we started updating https://github.com/prometheus/blackbox_exporter/blob/master/CHANGELOG.md

electron0zero: also, can you resolve the conflicts please, I merged #1461 and now there's a conflict on main.go

endesapt: I synced my branch to current master, resolved conflicts, and wrote notes to changelog.

PR #1468, opened 2025-09-23
repo: create pull request template

electron0zero:
Add a pull request template with guidelines for contributors.

with a template, it makes it more clear and produce more uniform PR descriptions, that helps us with producing better changelog for end users when we do the release.

inspired by https://github.com/prometheus/prometheus/pull/16904 but I am skipping the CI Check.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,47 @@
+<!--
+    - Please give your PR a title in the form "area: short description".  For example "tsdb: reduce disk usage by 95%"
```
SuperQ: The example should be more relevant. Maybe something like `probe_http: ....`

electron0zero: thanks, updated the example and link to change log.

PR #1469, opened 2025-09-24
Update dns.go

jamala0309:
Declaring and registering the probeFailedDueToRegex gauge.

Passing it into validRRs and setting it to 1 whenever a regex validation fails.

### DIFF
--- a/prober/dns.go
+++ b/prober/dns.go
@@ -3,7 +3,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-// http://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


prober/dns.go
```diff
@@ -145,6 +167,11 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 		Name: "probe_dns_query_succeeded",
 		Help: "Displays whether or not the query was executed successfully",
 	})
+	// NEW METRIC
```
SuperQ, line 170:
```suggestion
```

prober/dns.go
```diff
@@ -263,10 +288,6 @@ func ProbeDNS(ctx context.Context, target string, module config.Module, registry
 	client.Timeout = time.Until(timeoutDeadline)
 	requestStart := time.Now()
 	response, rtt, err := client.Exchange(msg, targetIP)
-	// The rtt value returned from client.Exchange includes only the time to
-	// exchange messages with the server _after_ the connection is created.
-	// We compute the connection time as the total time for the operation
-	// minus the time for the actual request rtt.
```
SuperQ, line 269: Why are you removing existing comments?

prober/dns.go
```diff
@@ -28,17 +28,26 @@ import (
...
 // validRRs checks a slice of RRs received from the server against a DNSRRValidator.
-func validRRs(rrs *[]dns.RR, v *config.DNSRRValidator, logger *slog.Logger) bool {
+func validRRs(
+	rrs *[]dns.RR,
+	v *config.DNSRRValidator,
+	logger *slog.Logger,
+	probeFailedDueToRegex prometheus.Gauge, // <-- new parameter
+) bool {
+	// reset metric at the start of each validation
+	probeFailedDueToRegex.Set(0)
```
SuperQ, line 38: This would be a lot less fragile as a defer.

Something like:
```suggestion
) (valid bool) {
	defer func() {
		if valid {
			probeFailedDueToRegex.Set(0)
		} else {
			probeFailedDueToRegex.Set(1)
		}
	}()
```

ArthurSens:
Closing due to inactivity

PR #1480, opened 2025-10-23
prober: added unix socket probes

edevil:
Added another type of probe, unix probes, similar to the tcp probes but which connect via a unix domain socket.

```release-notes
[FEATURE] Add a new type of probe using unix domain sockets
```

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/prober/unix.go
+++ b/prober/unix.go
@@ -16,8 +16,11 @@ package prober
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"log/slog"
 	"net"
+	"os"
+	"syscall"
 
 	"github.com/prometheus/client_golang/prometheus"
 	pconfig "github.com/prometheus/common/config"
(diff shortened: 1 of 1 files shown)


config/testdata/invalid-unix-query-response-regexp.yml
```diff
@@ -0,0 +1,9 @@
+modules:
+  unix_socket_bad_regex:
+    prober: unix
+    timeout: 5s
+    unix:
+      query_response:
+        - send: "PING"
+          expect: ":["
+
```
SuperQ:
```suggestion
```

prober/query_response.go
```diff
@@ -0,0 +1,170 @@
+// Copyright 2023 The Prometheus Authors
```
SuperQ: We're removing the date from these headers.

```suggestion
// Copyright The Prometheus Authors
```

config/testdata/invalid-unix-query-response-regexp.yml
edevil: Done

prober/query_response.go
edevil: Done.

example.yml
```diff
@@ -215,3 +215,10 @@ modules:
       transport_protocol: "tcp" # defaults to "udp"
       preferred_ip_protocol: "ip4" # defaults to "ip6"
       query_name: "www.prometheus.io"
+  unix_socket_example:
```
SuperQ:
```suggestion
  unix_socket_ping:
```

edevil: Done.

prober/unix.go
```diff
@@ -0,0 +1,59 @@
...
+	"net"
+
+	"github.com/prometheus/client_golang/prometheus"
+	pconfig "github.com/prometheus/common/config"
+
+	"github.com/prometheus/blackbox_exporter/config"
+)
+
+func dialUnix(ctx context.Context, target string, module config.Module, registry *prometheus.Registry, logger *slog.Logger) (net.Conn, error) {
+	dialer := &net.Dialer{}
```
electron0zero, line 32: any reasons to not use `net.DialUnix`?

prober/unix_test.go
```diff
@@ -0,0 +1,76 @@
...
+	tmpfile, err := os.CreateTemp("", "unix-socket-test")
+	if err != nil {
+		t.Fatalf("Error creating temp file: %s", err)
+	}
+	socketPath := tmpfile.Name()
+	// Close and remove the file so we can use the path for the socket.
+	tmpfile.Close()
+	os.Remove(socketPath)
+
+	ln, err := net.Listen("unix", socketPath)
```
electron0zero, line 41: same as above, does it make sense to use https://pkg.go.dev/net#ListenUnix here?

CONFIGURATION.md
```diff
@@ -221,6 +222,37 @@ tls_config:
 
 ```
 
+### `<unix_probe>`
+
+```yml
+
+# The query sent in the unix socket probe and the expected associated response.
```
electron0zero, line 229: can you update the readme to mention unix_probe, and also add a line or two about permissions.

My understanding is that you need to be part of the same group (_for example: probing docker socket_), otherwise you will get permissions errors.

so imo, it's better to document the permissions, otherwise we will have users trying to use, and then fail.

prober/unix.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		logger.Error("Error creating TLS configuration", "err", err)
+		return nil, err
+	}
+
+	timeoutDeadline, _ := ctx.Deadline()
+	dialer.Deadline = timeoutDeadline
+
+	logger.Debug("Dialing unix with TLS")
+	return tls.DialWithDialer(dialer, "unix", target, tlsConfig)
```
electron0zero: does it make sense to handle socket permissions errors and log a detailed log message for those errors?

prober/unix.go
edevil: It doesn't allow passing in a Context. I see no advantage in using it tbh.

prober/unix_test.go
edevil: I kind of prefer the generic functions. In this case both approaches seem very similar. ListenUnix does require a UnixAddr instead of a generic string, but since we're building the path ourselves it shouldn't be an issue. This way the code is similar to the tcp path.

CONFIGURATION.md
edevil: Updated readme.

prober/unix.go
edevil: Done.

electron0zero:
Thanks for the PR 🙏🏼

I will for wait for few days to give more time for other maintainers to chime in. If we don't get more comments, I will merge it.

PR #1483, opened 2025-10-30
dns: make query_name optional; fallback to target hostname when omitted

serrrios:
Makes dns.query_name optional. When omitted the prober uses the hostname part of target. Rationale: simplifies common configs and reduces duplication. Changes: remove mandatory check in DNSProbe.UnmarshalYAML; derive qName in prober/dns.go; update documentation. Backward compatibility: existing configs continue to work. Tests: keep explicit query_name behavior and add fallback coverage.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -241,7 +241,8 @@ tls_config:
 tls_config:
   [ <tls_config> ]
 
-query_name: <string>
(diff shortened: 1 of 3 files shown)


ArthurSens:
Hi, I'm taking the liberty of closing this PR because it's been inactive. Also, #1648 seems closer to the right implementation, so I'll close this one so we can focus on a single PR.

Thanks for your efforts!

PR #1484, opened 2025-10-30
metrics: add probe_ip_info gauge with labels {target, ip, protocol}

serrrios:
Adds probe_ip_info=1 to expose the resolved IP and protocol (ip4 or ip6) during resolution including fallback paths. Rationale: improves observability for dual-stack behavior and IP changes. Changes: register and set probe_ip_info in chooseProtocol across all resolution branches. Backward compatibility: additive metric. Tests: assert presence and correct labels for IPv4 IPv6 and fallback.

### DIFF
--- a/prober/utils.go
+++ b/prober/utils.go
@@ -46,9 +46,14 @@ func chooseProtocol(ctx context.Context, IPProtocol string, fallbackIPProtocol b
 		Name: "probe_ip_addr_hash",
 		Help: "Specifies the hash of IP address. It's useful to detect if the IP address changes.",
 	})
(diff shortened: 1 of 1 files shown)


SuperQ:
The original change to add the IP address hash (as a value) was intentionally done that way to avoid the cardinality issues with exposing IP addresses.

See the original PR: https://github.com/prometheus/blackbox_exporter/pull/584

ArthurSens:
As mentioned above, adding the metric unconditionally introduces serious cardinality problems. 

On #1502, we've been discussing adding this feature behind a configuration flag. Nothing has been decided yet, but if we proceed, that's the way we're gonna go.

Since this PR has been inactive for a long time, I'm taking the liberty to close it.

Issue #819, opened 2021-08-14
ip_protocol_fallback when IPv6 target is blackholed


candlerb:
### Host operating system: output of `uname -a`

```
Linux prometheus 5.4.0-80-generic #90~18.04.1-Ubuntu SMP Tue Jul 13 19:40:02 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
### blackbox_exporter version: output of `blackbox_exporter --version`
```
blackbox_exporter, version 0.19.0 (branch: HEAD, revision: 5d575b88eb12c65720862e8ad2c5890ba33d1ed0)
  build user:       root@2b0258d5a55a
  build date:       20210510-12:56:44
  go version:       go1.16.4
  platform:         linux/amd64
```

### What is the blackbox.yml module config.
```
modules:
  certificate:
    prober: tcp
    timeout: 300s
    tcp:
      tls: true
      tls_config: {}
```
### What is the prometheus.yml scrape config.
n/a

### What logging output did you get from adding `&debug=true` to the probe URL?
```
# time curl -g 'localhost:9115/probe?module=certificate&target=prometheus.example.com:443&debug=true'
Logs for the probe:
ts=2021-08-14T11:21:58.27193929Z caller=main.go:320 module=certificate target=prometheus.example.com:443 level=info msg="Beginning probe" probe=tcp timeout_seconds=119.5
ts=2021-08-14T11:21:58.272163494Z caller=tcp.go:40 module=certificate target=prometheus.example.com:443 level=info msg="Resolving target address" ip_protocol=ip6
ts=2021-08-14T11:21:58.272451208Z caller=tcp.go:40 module=certificate target=prometheus.example.com:443 level=info msg="Resolved target address" ip=2001::1
ts=2021-08-14T11:21:58.272496013Z caller=tcp.go:121 module=certificate target=prometheus.example.com:443 level=info msg="Dialing TCP with TLS"
ts=2021-08-14T11:23:57.773216549Z caller=main.go:130 module=certificate target=prometheus.example.com:443 level=error msg="Error dialing TCP" err="dial tcp6 [2001::1]:443: i/o timeout"
ts=2021-08-14T11:23:57.773399293Z caller=main.go:320 module=certificate target=prometheus.example.com:443 level=error msg="Probe failed" duration_seconds=119.501374766



Metrics that would have been returned:
# HELP probe_dns_lookup_time_seconds Returns the time taken for probe dns lookup in seconds
[...]

PR #1485, opened 2025-10-30
Expand the documentation of address selection and `ip_protocol_fallback`

candlerb:
docs: add description of `ip_protocol_fallback` logic

Fixes: #819

```release-notes
NONE
```

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -25,6 +25,32 @@ modules:
 
 ```
 
+### Address selection
+
+The blackbox_exporter is a testing tool.  Because of this requirement it
+does not behave like a typical end-user client.  Only a single connection
+attempt is made to the target under test.  There is no "happy eyeballs"
+style of connection parallelization or retry.
+
+When you run a test, you want to know what you're testing.  If a host has
+both AAAA and A records then you should be running two separate tests: you
+don't want the service being "half up" to be considered as "working".  (Such
+issues are the bane of the dual stack world, where nobody notices IPv6 going
+down).
+
+Therefore, if the target is specified using a hostname, and that hostname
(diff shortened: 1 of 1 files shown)


SuperQ:
You wrote this in the in the linked issue. I actually think including part of this in the first paragraph of your change here would be helpful to users. Specifically talking about "you need two tests".

> When you run a test, you want to know what you're testing. If a host has both AAAA and A records then you should be running two separate tests: you don't want the service being "half up" to be considered as "working". (Such issues are the bane of the dual stack world, where nobody notices IPv6 going down).

CONFIGURATION.md
```diff
@@ -25,6 +25,23 @@ modules:
 
 ```
 
+### Address selection
+
+Only a single connection attempt is made to the target under test - there is
+never any "happy eyeballs" style of connection retry.  If the target is
```
SuperQ: How about this phrasing change:

```suggestion
The blackbox_exporter is a testing tool. Because of this requirement it does not behave like
a typical end-user client. Only a single connection attempt is made to the target under test.
There is no "happy eyeballs" style of connection parallelization or retry.  If the target is
```

CONFIGURATION.md
```diff
@@ -25,6 +25,23 @@ modules:
...
+specified using a hostname, and that hostname resolves to multiple
+addresses, then a single address is selected to test according to the
```
SuperQ: This is also true if an host lookup returns multiple A or AAAA records. We've had similar support issues in the past where users expect different behavior when multiple A records are returned.

```suggestion
addresses or IP protocols, then a single address is selected to test according to the
```

candlerb:
I have altered the wording in a way which hopefully incorporates your comments.

Issue #740, opened 2021-01-20
Feature Request: Return the resolved IP address 


stwellman:
For modules that return an IP address hash, it would be useful to also return the resolved IP address.  The code is already resolving the IP and could return it as a label similar to the below examples: 

probe_ip_addr_hash{ip_addr="127.0.0.1"} 5.98285105e+08
probe_resolved_ip_addr_info{ip_addr="127.0.0.1"} 1

A specific use-case would be for companies monitoring servers in multiple data centers since the data center can often be derived from the network part of the address.   The presentation layer (like Grafana) could possibly translate the IP to a data center name using regex for easier isolation of root problem.

brian-brazil:
This is on purpose, as it could lead to cardinality issues as there's effectively no bound on how many different IP addresses might be resolved over time especially for v6. Thus it's only returned as a hashed value.

stwellman:
Thanks for the reply.  I think I understand what you are saying.  Basically, each time the address changes, it would be a new time series since the IP is essentially part of the "key".  That would totally break the functionality of the hash.  For my purposes, though, I still think the benefits outweigh the drawbacks.  I am going to modify the code locally, but I can understand why it can't be added here.

brian-brazil:
> Basically, each time the address changes, it would be a new time series since the IP is essentially part of the "key".

Yes, and each of those costs RAM in Prometheus. Thousands are not out of the question, more if IPv6 privacy stuff that continuously rotated addresses is involved.

> That would totally break the functionality of the hash.

That's separate. There's no way to expose this as a label that avoids the cardinality issue.

PR #1502, opened 2025-11-25
prober: added ip address to metric label

fluzzykitten:
Adds IP address as label in returned metrics. Having the hash as a metric is lovely and all, but often you need the actual IP, which is now included as a label on metrics returned from probes.

Previously, metrics had the hash:
```release-notes
probe_ip_addr_hash 3.521344457e+09
probe_ip_protocol 4
# TYPE probe_success gauge
probe_success 1
```
Now it also includes the actual IP:
```release-notes
# HELP probe_ip_addr IP address. It's useful.
# TYPE probe_ip_addr gauge
probe_ip_addr{ip="10.42.49.95"} 1
```

**Checklist**
- [X] Tests updated
- [X] Documentation added

Screenshot 2025-11-25 at 9 51 37 AM


fluzzykitten:
Hi @brian-brazil @SuperQ , are you guys the codeowners/maintainers for the blackbox exporter?

This PR adds the resolved IP address as an additional label on probe metrics, which can be very helpful for debugging when DNS resolution varies or when monitoring anycast/CDN endpoints.

Would it be possible to get a review when you have a moment? Happy to make any adjustments needed.

Thanks!

prober/utils.go
```diff
@@ -46,9 +46,18 @@ func chooseProtocol(ctx context.Context, IPProtocol string, fallbackIPProtocol b
 		Name: "probe_ip_addr_hash",
 		Help: "Specifies the hash of IP address. It's useful to detect if the IP address changes.",
 	})
+
+	probeIPAddr := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "probe_ip_addr",
+		Help: "IP address. It's useful.",
```
ArthurSens, line 52: This help text looks unhelpful

ArthurSens (changes requested):
IP addresses can become unbounded, which is the reason why this feature was already rejected at #740 

I'm not saying we should accept this now, but IF we accept, I believe it needs to be disabled by default and there's a configuration option that enables it.

Also, if you run blackbox exporter with debug log level, you do get the ip addresses of every probe in the logs. I'd argue that log search engines are better suited for high-cardinality operations than prometheus 🤔

prober/utils.go
fluzzykitten: Disabled by default makes sense. This change has been huge for our metrics. Blackbox exporter is fantastic, but with this change, we can both verify with labels that the IP is what we're expecting, and having the IP as a string attached to the metrics allows us to do pivot tables with this data to combine it with other data in Prometheus that has the IP. Would you like me to change the MR to include a disabled by default with a flag to toggle?

fluzzykitten: As far as the logs comment, unfortunately the logging is not helpful in this case that we need to create metric pivot tables in Grafana using the prometheus metric data. Sadly, we have ~4000 servers we're getting metrics on, and the logs are piped out to a completely different system (elasticserach) and are unavailable to use in our metrics stack (prometheus/thanos). The IP needs to exist on the metircs for us to use this information in our alerting and dashboards in Grafana with the Prometheus data.

ArthurSens: I believe we still need to validate the idea with other maintainers, but if we move forward it will probably be by making this behavior configurable and disabled by default

electron0zero: I can see the use-case but it should be disabled by default so folks who wish to have this can enable it.

please update the metric address, and name to follow the style of existing metrics, and add docs as well.

PR #1503, opened 2025-11-28
feat: http 'source_interface' parameter

raboof:

Add a 'source_interface' parameter to the 'http' probe

```release-notes
[FEATURE] allow restricting the HTTP probe to a given network interface
```

Sadly I didn't see a clear way to meaningfully test this feature.

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,21 @@ scrape_configs:
         target_label: vhost  # and store it in 'vhost' label
 ```
 
+HTTP probes can also limit themselves to a particular
+interface with the `source_interface` parameter. This
+feature works only for 'regular' (non-HTTP3) HTTP
+probes, is likely Linux-specific and may require you
+to disable return-path (rp) filtering in the network
+configuration of the host on which blackbox runs:
+
+```
+modules:
+  http_2xx:
+    prober: http
+    http:
+      source_interface: wlan0
+```
+
 ## Permissions
 
 The ICMP probe requires elevated privileges to function:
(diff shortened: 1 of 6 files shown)


electron0zero:
I did some digging into the topic and I think that this is not something that should be supported by blackbox exporter,  because it falls outside the core responsibility of the blackbox exporter.

You can use [network namespaces](https://man7.org/linux/man-pages/man7/network_namespaces.7.html) to bind your process to a network namespace that's attached to your interface, if that doesn't work for you, checkout [few examples in here](https://unix.stackexchange.com/questions/210982/bind-unix-program-to-specific-network-interface) and [here](https://forums.docker.com/t/could-a-container-bind-a-specific-network-interface-to-reach-external-network/39722/5)

raboof:
> I did some digging into the topic 

Thank you!

> I think that this is not something that should be supported by blackbox exporter, because it falls outside the core responsibility of the blackbox exporter.
> 
> You can use [network namespaces](https://man7.org/linux/man-pages/man7/network_namespaces.7.html) to bind your process to a network namespace that's attached to your interface
>
> if that doesn't work for you, checkout [few examples in here](https://unix.stackexchange.com/questions/210982/bind-unix-program-to-specific-network-interface) and [here](https://forums.docker.com/t/could-a-container-bind-a-specific-network-interface-to-reach-external-network/39722/5)

I see I did not do a good job explaining what the use case for the feature in this PR is: I have a node that has both a wired ethernet connection and a wifi card. I want to run the blackbox exporter on this node to monitor the connectivity of a target host both via ethernet and via wifi. Perhaps that was clear to you but I wanted to make sure to describe it in the PR now :)

It looks like your suggestion is to wrap blackbox-exporter using the techniques you link, to 'convince' it to bind to the 'right' interface, right? tbh that looks quite a bit more complex than adding the feature to blackbox-exporter :) . Of course if you decide it's out-of-scope for blackbox-exporter we can close this PR, but I'll likely just keep using this patch.

electron0zero:
I see your use-case and I think that it's outside the scope of the core responsibility of the blackbox exporter. I am going to close this PR because it's out of scope.

Issue #1509, opened 2025-12-06
Unable to do unprivileged listen on socket,


datamuc:
The latest docker image logs a lot of these when doing ICMP checks.
```
time=2025-12-06T22:08:50.901Z level=ERROR source=icmp.go:178 msg="Unable to do unprivileged listen on socket, will attempt privileged" module=ping4 target=... err="socket: permission denied"
time=2025-12-06T22:08:50.990Z level=ERROR source=icmp.go:126 msg="Unable to do unprivileged listen on socket, will attempt privileged" module=ping6 target=... err="socket: permission denied"
```

The level of that message was changed from DEBUG to ERROR recently. Is that supposed to be so? To me the [code](https://github.com/prometheus/blackbox_exporter/blob/master/prober/icmp.go#L122-L130) reads like it is trying to figure out how to do things and it should continue to be DEBUG.

electron0zero:
we updated the log levels and changed the logging behaviour in v0.28.0 to be more accurate, please see release notes: [v0.28.0 (release)](https://github.com/prometheus/blackbox_exporter/releases/tag/v0.28.0)

see more: https://github.com/prometheus/blackbox_exporter#controlling-log-level-for-probe-logs

if you still think this error is unexpected, please share the config and more details on the setup?

datamuc:
it's just the default config:
```
docker run --network=host  --rm quay.io/prometheus/blackbox-exporter:v0.28.0 \
  --web.listen-address=:19115 --config.file=/etc/blackbox_exporter/config.yml
```

```
$ curl -s localhost:19115/probe -G -d module=icmp -d target=localhost | grep probe_success
# HELP probe_success Displays whether or not the probe was a success
# TYPE probe_success gauge
probe_success 1
```

and it logs ERRORS even though the probe was successful

```
time=2025-12-16T13:50:27.351Z level=ERROR source=icmp.go:126 msg="Unable to do unprivileged listen on socket, will attempt privileged" module=icmp target=localhost err="socket: permission denied"
```


The code in question:
```
               /* We try it unprivileged */
		if tryUnprivileged {
			// "udp" here means unprivileged -- not the protocol "udp".
			icmpConn, err = icmp.ListenPacket("udp6", srcIP.String())
			if err != nil {
				logger.Error("Unable to do unprivileged listen on socket, will attempt privileged", "err", err) /* i still think these should not be errors */
			} else {
				privileged = false
			}
		}

[...]

PR #1517, opened 2026-01-01
Update log level from error to debug and info to reduce log noise

electron0zero:

Undo some of the log level changes made in https://github.com/prometheus/blackbox_exporter/pull/1461/

Fixes https://github.com/prometheus/blackbox_exporter/issues/1509

```release-notes
fix: Update log level from error to debug and info to reduce log noise
```

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -440,6 +440,7 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 
 		// HTTP/3 requires TLS 1.3 minimum
 		if tlsConfig.MinVersion < tls.VersionTLS13 {
+			logger.Debug("Setting TLS Version to 1.3 because HTTP/3 requires TLS 1.3 minimum")
 			tlsConfig.MinVersion = tls.VersionTLS13
 		}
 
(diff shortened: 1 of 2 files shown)


SuperQ:
Shouldn't we make these `Warn` or maybe `Debug`?

:eyeglasses: @tjhop

electron0zero:
> Shouldn't we make these `Warn` or maybe `Debug`?

not sure if Warn would address the log noise concerns folks are reporting, I am fine with both Debug or Info. @tjhop what do you say?

datamuc:
> > Shouldn't we make these `Warn` or maybe `Debug`?
> 
> not sure if Warn would address the log noise concerns folks are reporting, I am fine with both Debug or Info. @tjhop what do you say?

I still think it should be debug.

tjhop:
For these error check cases where we're continuing on, I think going with debug logging makes sense in context here -- we would only really care about them if we needed to troubleshoot the entire flow of the probe. For the error checks where we actually stop and return, I think those should remain at error level, though.

wat-org:
hello,

I also suggest to switch from 'warn' to 'debug' two logs in the http prober concerning redirection:
* https://github.com/prometheus/blackbox_exporter/blob/master/prober/http.go#L490
* https://github.com/prometheus/blackbox_exporter/blob/master/prober/http.go#L493

nb:
it was switched from 'info' to 'warn' 4 month ago and generates log noise too

regards

Lusitaniae:
When could we expect to get this released?

PR #1138, opened 2023-10-19
enhancement: config file probe type validation

fgouteroux:
Hello,

I want to enhance the validation of config file for probe type when using `--config.check` option.

```
$ blackbox_exporter  --config.check --config.file config.yml 
ts=2023-10-19T08:46:20.927Z caller=main.go:86 level=info msg="Starting blackbox_exporter" version="(version=0.24.0, branch=fgx-0.24, revision=2c699796ff6809c45027247f142bc7fea13469c1)"
ts=2023-10-19T08:46:20.927Z caller=main.go:87 level=info build_context="(go=go1.20.5, platform=linux/amd64, user=fgouteroux, date=20231019-08:45:34, tags=netgo)"
ts=2023-10-19T08:46:20.953Z caller=main.go:98 level=error err="Unknown probe type 'http2' for module name 'fgx-online"
ts=2023-10-19T08:46:20.953Z caller=main.go:98 level=error err="Unknown probe type 'http1' for module name 'fgx-online-service"
```

PR #1521, opened 2026-01-13
config: fail if field 'prober' contains an invalid value

wat-org:
The configuration of the blackbox_exporter accepts invalid prober value like this:
```
modules:
  http_post_2xx:
    prober: invalid     # invalid value here. should be either 'http', 'tcp', 'icmp', 'dns', 'grpc' or 'unix'
    timeout: 5s
    http:
      method: POST
      valid_http_versions: ["HTTP/1.1", "HTTP/2.0"]
```

This fix will refuse such configuration.: blackbox_exporter will fail to start/reload with:
```
msg="Error loading config" err="error parsing config file: prober 'invalid' is not valid"
```

NB: This PR also fixes a comment and a test case (the test file 'blackbox-good.yml' was containing a 'prober: grpc_probe', which is an invalid value)

Yes. In the previous version, blackbox_exporter successfully starts and an http error was returned when the invalid prober was called.

```release-notes
fix: validate prober names and fail in case of invalid prober names
```

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -418,6 +418,7 @@ func (s *Module) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	switch s.Prober {
 	case "http", "tcp", "icmp", "dns", "grpc", "unix":
 		// valid
+		return nil
 	default:
 		// invalid
 		return fmt.Errorf("prober '%s' is not valid", s.Prober)


wat-org:
This PR duplicates https://github.com/prometheus/blackbox_exporter/pull/1138, as they both raise an error when the 'prober' field contains an invalid value.

but the implementations differ:
The PR 1138 detects the error once the yaml configuration file is fully parsed, using a new function.
This PR detects the error during the parsing, as it is done for the other validation. So it fit better with the current logic of the code.

wat-org:
hello @electron0zero 
I've seen you have merged a PR few days ago.
Could you have a look at this one ?
Regards

config/config.go
```diff
@@ -415,6 +415,13 @@ func (s *Module) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := unmarshal((*plain)(s)); err != nil {
 		return err
 	}
+	switch s.Prober {
+	case "http", "tcp", "icmp", "dns", "grpc", "unix":
+		// valid
```
electron0zero, line 420:
```suggestion
		// valid
		return nil
```

I think we should return here because prober is valid

electron0zero (approved):

wat-org:
many thanks @electron0zero

PR #1525, opened 2026-01-22
prober: handle single and double encoded target parameter

### LINKED ISSUE #1520: Add support to specify port as a parameter
Currently, blackbox uses 'target' to specify port, however, when trying to use that in prometheus/alloy, that portion will get url-encoded, breaking things:

> Jan 08 13:46:03 briantest4 alloy[46804]: ts=2026-01-08T18:46:03.87421331Z level=debug msg="Scrape failed" component_path=/ component_id=prometheus.scrape.blackbox_ssh scrape_pool=prometheus.scrape.blackbox_ssh target="http://127.0.0.1:9115/probe?module=blackbox_ssh&target=127.0.0.1%3A22" err="server returned HTTP status 400 Bad Request"

That is sensible on their end for various reasons.
By adding support for a 'port' parameter, we can avoid issues and make it obvious when targeting services running on non-standard ports

AnuragEkkati:
This feature addresses the target URL encoding issues by decoding it before application

Fixes #1520
This PR decodes the target url incase if it is encoded

#### Test Results
Screenshot 2026-02-04 at 12 06 04 PM

```release-notes
[CHANGE] prober: handle single and double encoded target parameter
```
#### AI Usage 
Used Cursor IDE to validate the test-cases.

**Checklist**
- [X] Tests updated
- [X] Documentation added

### DIFF
--- a/prober/handler_test.go
+++ b/prober/handler_test.go
@@ -304,3 +304,90 @@ func TestTCPHostnameParam(t *testing.T) {
 	}
 
 }
+
+func TestURLDecoding(t *testing.T) {
+	c := &config.Config{
+		Modules: map[string]config.Module{
+			"http_2xx": {
+				Prober:  "http",
+				Timeout: 10 * time.Second,
+				HTTP: config.HTTPProbe{
+					IPProtocolFallback: true,
+				},
+			},
+		},
+	}
+
+	// Create a test server that echoes back request details
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer ts.Close()
+
+	tsHost, tsPort, _ := net.SplitHostPort(strings.TrimPrefix(ts.URL, "http://"))
+
+	tests := []struct {
+		name          string
+		targetParam   string
+		shouldSucceed bool
+	}{
+		{
+			// Normal target without any encoding
+			name:          "normal_target",
+			targetParam:   ts.URL,
+			shouldSucceed: true,
+		},
+		{
+			// Double-encoded colon in port: %253A -> %3A -> :
+			// This happens when clients like Prometheus/Alloy encode an already-encoded URL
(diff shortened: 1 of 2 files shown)


AnuragEkkati:
Hello Team, 

Can someone please review these changes :). We need this fix in official release. It will be really helpful 

@electron0zero @jesusvazquez

electron0zero:
> avoiding URL encoding issues when the colon in target=host:port gets encoded as %3A by some HTTP clients like grafana/alloy.

 @AnuragEkkati I think it makes sense to handle and fix the url encoding instead of adding a special case for port, supporting the url encoding would support more use-cases then just adding port?

AnuragEkkati:
That makes sense.

electron0zero:
> That makes sense.

cool, in that I would say that we should add support for that in blackbox exporter, and that will solve the url encoded port's issue as well.

I am happy to review if you can put a new PR or update the current PR for it.

AnuragEkkati:
@electron0zero, updated the code to just handle the target URL decoding. Also added a test case to validate the same.

prober/handler_test.go
```diff
@@ -304,3 +304,70 @@ func TestTCPHostnameParam(t *testing.T) {
...
+			expectedTarget: fmt.Sprintf("http://%s:%s", tsHost, tsPort),
+			shouldSucceed:  true,
+		},
+		{
+			name:           "normal_target",
+			targetParam:    ts.URL,
+			expectedTarget: ts.URL,
+			shouldSucceed:  true,
+		},
+	}
```
electron0zero, line 362: let's add few more cases for better coverage

- Double encoded - `http://example.com%253A8080/path`, should fail because it's double encoded
- Basic auth with encoded colon - `http://user%3Apass@example.com%3A8080/`, should work
- Fully encoded URL - `http%3A%2F%2Fuser%3Apass@example.com%3A8080%2Fpath`, should work
- Edge case: %3A in path - `http://example.com/api/v1%3Alatest`, I think it should work and probe `http://example.com/api/v1:latest`

AnuragEkkati: added few more testcases for better coverage

PR #1537, opened 2026-02-06
prober: randomize ICMP echo identifier to avoid SNAT session collision

6ylqq:
This PR randomizes the ICMP Echo Identifier used by the blackbox exporter when sending ICMP probes.

In the current implementation, ICMP Echo Requests are sent with a fixed identifier. Since ICMP does not have a port concept, some SNAT implementations rely on the ICMP identifier to build session mappings.

When multiple clients behind the same SNAT send ICMP probes with the same ICMP identifier to the same destination:

- The SNAT device may create colliding sessions.

- ICMP Echo Replies from the server can be matched to the wrong client.

- One client may receive no reply, while another client may receive multiple replies and drop those with mismatched destinations.

By randomizing the ICMP Echo Identifier per probe, each ICMP flow becomes distinguishable at the SNAT layer, avoiding session collisions and ensuring that replies are correctly delivered to the originating client.

This change fixes ICMP probe failures observed in SNAT environments with multiple concurrent clients.

```release-notes
[BUGFIX] Randomize ICMP Echo Identifier to avoid SNAT session collisions when multiple clients send ICMP probes concurrently.
```

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ BREAKING CHANGES:
 * [FEATURE]
 * [ENHANCEMENT]
 * [BUGFIX]
+* [BUGFIX] Randomize ICMP Echo ID in probes to avoid SNAT session collisions that could drop replies for some clients.
 
 ## 0.28.0 / 2025-12-04
 
--- a/prober/icmp.go
+++ b/prober/icmp.go
@@ -19,7 +19,6 @@ import (
 	"log/slog"
 	"math/rand"
 	"net"
-	"os"
 	"runtime"
 	"sync"
 	"time"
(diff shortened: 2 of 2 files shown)


SuperQ (approved):
Yea, per-request random IDs seems like a good idea to me.

6ylqq:
Hi maintainers,

Thanks for merging this fix.

We noticed that #1537 was merged after the latest release v0.28.0, so the current published release artifacts do not include this change yet.

This fix is important for our environment because ICMP probes can run behind SNAT, and reusing the same ICMP echo identifier may cause session collision / incorrect probe behavior. This can lead to incorrect probe results in production-like deployments.

Would it be possible to cut a new release that includes #1537, or share the expected timeline for the next blackbox_exporter release?

If a patch release is acceptable, v0.28.1 including this fix would be very helpful for downstream users affected by SNAT-based ICMP probing.

Thanks!

PR #1557, opened 2026-03-22
fix: return on grpc.NewClient error and propagate probe context

Aprazor:
Two fixes in ProbeGRPC:
1. When grpc.NewClient returns an error, execution continued into
   a nil conn, causing a panic. Add 'return false' after the error.
2. client.Check used context.Background() instead of the probe ctx,
   bypassing the scrape timeout. Use ctx to honor the deadline.

Signed-off-by: Aprazors <Aprazors@gmail.com>


ArthurSens (approved):
LGTM, thanks :)

PR #1559, opened 2026-03-22
fix: replace time.Tick with time.NewTicker to prevent goroutine leak

Aprazor:
time.Tick inside a select creates a new ticker on every loop iteration
that is never stopped. Replace with a single NewTicker created before
the loop with defer Stop() for proper cleanup.

Signed-off-by: Aprazors <Aprazors@gmail.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -143,6 +143,8 @@ func run() int {
 	reloadCh := make(chan chan error)
 	signal.Notify(hup, syscall.SIGHUP)
 	go func() {
+		autoReloadTicker := time.NewTicker(time.Duration(*autoReloadInterval) * time.Second)
+		defer autoReloadTicker.Stop()
 		for {
 			select {
 			case <-hup:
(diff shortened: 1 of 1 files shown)


ArthurSens:
Hey, since Go 1.23+, [tickers who are unreacheable by code can be garbage collected](https://go.dev/doc/go1.23#timer-changes:~:text=First%2C%20Timers%20and%20Tickers%20that%20are%20no%20longer%20referred%20to%20by%20the%20program%20become%20eligible%20for%20garbage%20collection%20immediately%2C%20even%20if%20their%20Stop%20methods%20have%20not%20been%20called.%20Earlier%20versions%20of%20Go%20did%20not%20collect%20unstopped%20Timers%20until%20after%20they%20had%20fired%20and%20never%20collected%20unstopped%20Tickers.), so the leak doesn't happen in reality :)

Additionally, `time.Tick(0)` returns a nil pointer, while `time.NewTicker(0)` panics. So the only thing we're adding here is a panic that didn't exist before if `--config.auto-reload-interval` was set to 0 during startup 😬 

With all that said, I'm closing this PR. Please reopen it if you think I'm misreading things!

PR #1574, opened 2026-04-27
[ENHANCEMENT] Unified SSL metrics registry

adnull:
SSL metrics: unified way to register/populate from raw connection

This change consolidates how TLS-related Prometheus metrics are created and populated so HTTP, gRPC, and WebSocket probes share the same implementation instead of duplicating metric definitions in each prober.

What changed

- Shared TLS helpers (prober/tls.go): Added registerTLSMetrics and setTLSMetrics, backed by a TLSMetrics struct, to register and set the standard SSL/TLS gauges in one place: earliest cert expiry, last chain expiry timestamp, leaf chain info (probe_ssl_last_chain_info), TLS version, and cipher.
- HTTP and gRPC (prober/http.go, prober/grpc.go): Removed per-prober copies of those metrics and call the shared helpers when a TLS connection is present.
- Naming (prober/prober.go): Centralized options for probe_ssl_last_chain_info via probeSSLLastInformationGaugeOpts so the help text and name stay consistent.
- Added probe_websocket_ssl to indicate whether the final connection used TLS.

gRPC previously exposed only a subset of TLS metrics (e.g. no chain-expiry or cipher series in the same way as HTTP). After this refactor, gRPC uses the same setTLSMetrics path, so it now emits the full set of shared TLS metrics, including probe_ssl_last_chain_expiry_timestamp_seconds and probe_tls_cipher_info, aligned with the HTTP and WebSocket behavior. 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/prober/grpc.go
+++ b/prober/grpc.go
@@ -99,21 +99,6 @@ func ProbeGRPC(ctx context.Context, target string, module config.Module, registr
 			Name: "probe_grpc_healthcheck_response",
 			Help: "Response HealthCheck response",
 		}, []string{"serving_status"})
-
-		probeSSLEarliestCertExpiryGauge = prometheus.NewGauge(sslEarliestCertExpiryGaugeOpts)
-
-		probeTLSVersion = prometheus.NewGaugeVec(
-			probeTLSInfoGaugeOpts,
-			[]string{"version"},
-		)
-
-		probeSSLLastInformation = prometheus.NewGaugeVec(
-			prometheus.GaugeOpts{
-				Name: "probe_ssl_last_chain_info",
-				Help: "Contains SSL leaf certificate information",
-			},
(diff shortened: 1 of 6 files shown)


ArthurSens (commented):
Hey @adnull, sorry for the delay here! I think this idea of unifying how SSL metrics are generated is great, but there are a few things that still need to be worked on before we proceed with the merge :)

First we'll need to rebase on latest master, and you'll notice that we're now checking CRL before we create the SSL metrics and we should preserve this behavior.

Tests added only cover websocket, and it might be good to add similar tests new metrics that are being created here.

In #1579, we're also adding SSL metrics to TCP/Unix probes, we should probably wait this one get merged and also include the refactoring being done here!

PR #1579, opened 2026-05-07
add tcp and unix prober metric for tls_cipher_info

nathwill:
adds tls_cipher_info metric to tcp prober

```release-notes
[FEATURE] add tls_cipher_info metric to tcp and unix probers
```

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/prober/query_response.go
+++ b/prober/query_response.go
@@ -60,6 +60,10 @@ func probeQueryResponses(ctx context.Context, target string, conn net.Conn, modu
 		probeTLSInfoGaugeOpts,
 		[]string{"version"},
 	)
+	probeTLSCipher := prometheus.NewGaugeVec(
+		probeTLSCipherGaugeOpts,
+		[]string{"cipher"},
+	)
 	probeFailedDueToRegex := prometheus.NewGauge(prometheus.GaugeOpts{
 		Name: "probe_failed_due_to_regex",
 		Help: "Indicates if probe failed due to regex",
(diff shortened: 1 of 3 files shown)


nathwill:
i have no idea why circleci couldn't find the ref, it seems like perhaps github was backlogged when the PR was opened or something and circleci beat github to the punch, as it took several minutes to show up in the PR list UI, by which time the tests had already failed.

ArthurSens:
Hey, a rebase should solve the CircleCI problem

but before we get into that, could you tell us the motivation behind this metric? There's no linked issue, the PR description also doesn't mention any problems you're trying to solve, so it sounds a bit weird to add a metric just for the sake of it

nathwill:
@ArthurSens yes it's to provide parity with the http/s prober so we can use the tls prober to determine not just reachability, but also the in-use ciphers, so we can better understand the impact of insecure ciphers from a client perspective as CVEs are announced.

nathwill:
note that this may be duplicative of the other open cipher-label related PR https://github.com/prometheus/blackbox_exporter/pull/1574; i didn't see engagement on that issue when i opened this so opened this one focused more narrowly on our specific need since i wasn't sure what the disposition of that issue may end up being.

ArthurSens:
> note that this may be duplicative of the other open cipher-label related PR #1574; i didn't see engagement on that issue when i opened this so opened this one focused more narrowly on our specific need since i wasn't sure what the disposition of that issue may end up being.

It doesn't seem like duplication to me; you're covering mostly TCP/Unix probes, and #1574 is consolidating HTTP/gRPC/WebSocket.

I think we should proceed with your PR ffirst and then provide guidance on #1574 to also include your work there :)

prober/query_response.go
ArthurSens, line 49: This function is also used by Unix probes, which means we're also adding a new metric to this path. Could we at least have a test for that too?

ArthurSens (commented):
Just one small test and we should be good to go

prober/query_response.go
nathwill: good catch, test added.

nathwill:
sounds good, thanks @ArthurSens :)

ArthurSens (approved):
Thanks!

PR #1583, opened 2026-05-19
feat(prober): add CRL certificate revocation checking

sebastiangaiser:
Check the full certificate chain against CRLs across the HTTP, TCP, gRPC, and Unix probes. For each cert with CRLDistributionPoints, fetch the CRL, verify its signature against the issuer, reject CRLs that are stale or not yet valid, and look up the cert serial in the revocation list. Enabled per probe via the opt-in `check_revoked` flag (default false); the CRL fetch is bounded by the probe timeout, no separate knob. DNS over TLS is not covered yet.

```release-notes
CRL certificate revocation checking
```

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -153,6 +153,12 @@ then a single address is selected to test, using the following logic:
   tls_config:
     [ <tls_config> ]
 
+  # Check each certificate in the TLS chain against its CRL and emit probe_ssl_crl_* metrics.
+  # This only reports revocation state, it never fails the probe: a revoked certificate or
+  # an unreachable CRL leaves probe_success untouched. Alert on the metrics instead.
+  # CRLs are fetched using the proxy settings configured below.
+  [ check_revoked: <boolean> | default = false ]
+
   # The HTTP basic authentication credentials.
   basic_auth:
     [ username: <string> ]
@@ -259,6 +265,12 @@ query_response:
 tls_config:
   [ <tls_config> ]
 
+# Check each certificate in the TLS chain against its CRL and emit probe_ssl_crl_* metrics.
+# This only reports revocation state, it never fails the probe: a revoked certificate or
+# an unreachable CRL leaves probe_success untouched. Alert on the metrics instead.
+# Requires tls to be true, or a query_response step using starttls.
+[ check_revoked: <boolean> | default = false ]
+
 ```
 
 ### `<unix_probe>`
@@ -289,6 +301,12 @@ query_response:
 # Configuration for TLS protocol of unix socket probe.
 tls_config:
   [ <tls_config> ]
+
+# Check each certificate in the TLS chain against its CRL and emit probe_ssl_crl_* metrics.
+# This only reports revocation state, it never fails the probe: a revoked certificate or
+# an unreachable CRL leaves probe_success untouched. Alert on the metrics instead.
+# Requires tls to be true, or a query_response step using starttls.
+[ check_revoked: <boolean> | default = false ]
 
 ```
 
@@ -415,6 +433,12 @@ metadata:
 # Configuration for TLS protocol of gRPC probe.
 tls_config:
   [ <tls_config> ]
+
+# Check each certificate in the TLS chain against its CRL and emit probe_ssl_crl_* metrics.
+# This only reports revocation state, it never fails the probe: a revoked certificate or
+# an unreachable CRL leaves probe_success untouched. Alert on the metrics instead.
+# Requires tls to be true.
+[ check_revoked: <boolean> | default = false ]
 ```
 
 ### `<websocket_probe>`
--- a/config/config.go
+++ b/config/config.go
@@ -315,7 +315,7 @@ type TLSConfigWithCRL struct {
 	CRLCheck         CRLCheckConfig `yaml:"crl_check,omitempty"`
 }
 
-func (t *TLSConfigWithCRL) UnmarshalYAML(unmarshal func(interface{}) error) error {
+func (t *TLSConfigWithCRL) UnmarshalYAML(unmarshal func(any) error) error {
 	type tlsPlain config.TLSConfig
 	type wrapper struct {
 		tlsPlain `yaml:",inline"`
@@ -354,7 +354,7 @@ type HTTPProbe struct {
 	Compression                  string                  `yaml:"compression,omitempty" json:"compression,omitempty"`
 	BodySizeLimit                units.Base2Bytes        `yaml:"body_size_limit,omitempty" json:"body_size_limit,omitempty"`
 	UseHTTP3                     bool                    `yaml:"enable_http3,omitempty" json:"enable_http3,omitempty"`
-	TLSCRLCheck                  CRLCheckConfig          `yaml:"tls_crl_check,omitempty" json:"tls_crl_check,omitempty"`
+	TLSCRLCheck                  CRLCheckConfig          `yaml:"tls_crl_check,omitempty" json:"tls_crl_check,omitzero"`
 }
 
 type GRPCProbe struct {
--- a/config/config.go
+++ b/config/config.go
@@ -327,7 +327,7 @@ func (t *TLSConfigWithCRL) UnmarshalYAML(unmarshal func(interface{}) error) erro
 	}
 	t.TLSConfig = config.TLSConfig(w.tlsPlain)
 	t.CRLCheck = w.CRLCheck
-	return t.TLSConfig.Validate()
+	return t.Validate()
 }
 
 type HTTPProbe struct {
--- a/config/config.go
+++ b/config/config.go
@@ -293,6 +293,11 @@ func MustNewRegexp(s string) Regexp {
 	return re
 }
 
+type CRLCheckConfig struct {
+	Enabled      bool          `yaml:"enabled,omitempty"`
+	FetchTimeout time.Duration `yaml:"fetch_timeout,omitempty"`
+}
+
 type Module struct {
 	Prober    string         `yaml:"prober,omitempty" json:"prober,omitempty"`
 	Timeout   time.Duration  `yaml:"timeout,omitempty" json:"timeout,omitempty"`
@@ -303,6 +308,26 @@ type Module struct {
 	GRPC      GRPCProbe      `yaml:"grpc,omitempty" json:"grpc,omitzero"`
 	Unix      UnixProbe      `yaml:"unix,omitempty" json:"unix,omitzero"`
 	Websocket WebsocketProbe `yaml:"websocket,omitempty" json:"websocket,omitzero"`
+}
+
+type TLSConfigWithCRL struct {
+	config.TLSConfig `yaml:",inline"`
+	CRLCheck         CRLCheckConfig `yaml:"crl_check,omitempty"`
+}
+
+func (t *TLSConfigWithCRL) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type tlsPlain config.TLSConfig
+	type wrapper struct {
+		tlsPlain `yaml:",inline"`
+		CRLCheck CRLCheckConfig `yaml:"crl_check,omitempty"`
+	}
+	var w wrapper
+	if err := unmarshal(&w); err != nil {
+		return err
+	}
+	t.TLSConfig = config.TLSConfig(w.tlsPlain)
+	t.CRLCheck = w.CRLCheck
+	return t.TLSConfig.Validate()
 }
 
 type HTTPProbe struct {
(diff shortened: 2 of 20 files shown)


prober/crl_test.go
```diff
@@ -0,0 +1,521 @@
...
+	"math/big"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/promslog"
+)
```
electron0zero, line 29: tests feels quite different from the rest of the test suite.

it introduces a lot of bespoke certificate/CRL fixture helpers and uses panic inside test helpers. Could we refactor these helpers to use t.Helper()/t.Fatalf() and reduce the duplicated setup so the tests stay closer to the repo’s usual style?

config/config.go
```diff
@@ -305,6 +310,26 @@ type Module struct {
 	Websocket WebsocketProbe `yaml:"websocket,omitempty" json:"websocket,omitzero"`
 }
 
+type TLSConfigWithCRL struct {
```
electron0zero: I think this would be simpler as a single opt-in flag like `check_revoked: true` in each check type instead of a separate CRL specific nested config. 

That will also match the existing style of short behavior flags in this repo and would let HTTP/TCP/gRPC/Unix/DNS-over-TLS share the same revocation switch without introducing extra TLS wrapper types. 


also, I don’t think CRL fetching needs its own timeout here. it should just use the probe/check timeout, so we avoid another config knob and keep the behavior aligned with the existing request lifecycle.

so just add a config called `check_revoked` or `check_cert_revoked` to perform the CRL and publish the CRL metrics, and leave the flag default value to false. 

we have defaults for each check so configure it there.

electron0zero (changes requested):

sebastiangaiser:
@electron0zero thanks for the review. I've reworked the PR along your suggestions:

- Added a single opt-in flag. Dropped the `CRLCheckConfig` type and the `TLSConfigWithCRL` wrapper entirely. Each probe now has a plain `check_revoked: <bool>` (default `false`) next to `tls`/`tls_config`; `tls_config` is back to the plain `config.TLSConfig`. HTTP/TCP/gRPC/Unix share the same switch.
- Removed the separate CRL fetch timeout, the fetch is now bounded by the probe timeout.
- Regarding DNS-over-TLS. I reverted DNS to `config.TLSConfig` and left DoT out for now (the TLS `ConnectionState` isn't readily accessible via the `miekg/dns` client), so there's no silently-ignored knob. Not sure how we should follow-up on that one.

I also addressed the Copilot findings: `probe_ssl_crl_fetch_time_seconds` now accumulates across distribution points and measures through the body read, and not-yet-valid CRLs (`thisUpdate` in the future) are rejected. Tests were refactored to `t.Helper()`/`t.Fatalf()` and I added integration tests that actually enable CRL and assert the metrics for TCP (direct TLS + StartTLS), Unix, and gRPC. `check_revoked` is now documented in `CONFIGURATION.md`.

prober/crl_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+// createTestCA generates a self-signed CA certificate and key.
+func createTestCA(t *testing.T) (*x509.Certificate, *rsa.PrivateKey) {
```
electron0zero: can we use helpers from prober/utils_test.go, we have generateCertificateTemplate, and other helpers that are using across the test suite.

prober/crl_test.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+	ca, err := x509.ParseCertificate(caDER)
+	if err != nil {
+		t.Fatalf("parsing CA certificate: %v", err)
+	}
+	return ca, caKey
+}
+
+// createTestLeafCert generates a leaf certificate signed by the CA.
+func createTestLeafCert(t *testing.T, ca *x509.Certificate, caKey *rsa.PrivateKey, serial *big.Int, crlURL string) (*x509.Certificate, *rsa.PrivateKey) {
```
electron0zero: same here, maybe we can use the helpers from prober/utils_test.go.

prober/crl_test.go
```diff
@@ -0,0 +1,455 @@
...
+			}
+			if match {
+				return m.GetGauge().GetValue(), true
+			}
+		}
+	}
+	return 0, false
+}
+
+func TestCheckChainCRL_ValidCert(t *testing.T) {
```
electron0zero: can we turn the test cases of the TestCheckChainCRL in a table test, these tests have a lot of similar code in them and might be a good fit for table driven test.

prober/crl_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+// serverTLSCert builds a tls.Certificate presenting the leaf and its CA.
+func serverTLSCert(leaf, ca *x509.Certificate, leafKey *rsa.PrivateKey) tls.Certificate {
+	return tls.Certificate{
+		Certificate: [][]byte{leaf.Raw, ca.Raw},
+		PrivateKey:  leafKey,
+	}
+}
+
+func getMetricValue(mfs []*dto.MetricFamily, name string) (float64, bool) {
```
electron0zero: can we use the existing helpers or if we need these new helpers, can we move them to utils_test.go.

prober/crl_test.go
```diff
@@ -0,0 +1,455 @@
...
+		if mf.GetName() == name {
+			if len(mf.GetMetric()) > 0 {
+				return mf.GetMetric()[0].GetGauge().GetValue(), true
+			}
+		}
+	}
+	return 0, false
+}
+
+func getMetricWithLabels(mfs []*dto.MetricFamily, name string, labels map[string]string) (float64, bool) {
```
electron0zero: same here

prober/crl.go
```diff
@@ -0,0 +1,269 @@
...
+func checkCRL(ctx context.Context, state *tls.ConnectionState, enabled bool, registry *prometheus.Registry, logger *slog.Logger) {
+	if !enabled {
+		return
+	}
+	crlResult := checkChainCRL(ctx, state, logger)
+	registerCRLMetrics(registry, crlResult)
+}
+
+// CRLCertResult holds the CRL check result for a single certificate in the chain.
+type CRLCertResult struct {
```
electron0zero: this is exported but I only see this being used in the same package, we can make it private and not export it if we don't need it outside?

prober/crl_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+	leafResult := result[0]
+	if !leafResult.Available {
+		t.Error("Expected CRL to be available")
+	}
+	if !leafResult.Revoked {
+		t.Error("Expected leaf cert to be revoked")
+	}
+}
+
```
electron0zero, line 206: can we add a test to ensure that we check CRL for the Intermediate certs as well?

config/config.go
electron0zero, line 489: we should do a similar validation to not allow check_revoked when tls: false.

if tls is false, allowing check_revoked: true doesn't make sense so we should validate and reject this config pair to ensure check_revoked: true is only allowed when tls is not false.

prober/http_test.go
```diff
@@ -1571,6 +1571,63 @@ func TestHTTPUsesTargetAsTLSServerName(t *testing.T) {
...
+			IPProtocol:         "ip4",
+			IPProtocolFallback: true,
+			CheckRevoked:       true,
+			HTTPClientConfig: pconfig.HTTPClientConfig{
+				TLSConfig: pconfig.TLSConfig{InsecureSkipVerify: true},
+			},
+		},
+	}
+
+	// Deadline shorter than the default 10s CRL fetch timeout, so the probe
```
electron0zero: is this still true, we removed CRL fetch timeout?

prober/crl.go
```diff
@@ -0,0 +1,269 @@
...
+		break
+	}
+
+	return result
+}
+
+// fetchCRL downloads a CRL from the given URL and parses it.
+// The request is bounded by the probe context deadline carried in ctx.
+func fetchCRL(ctx context.Context, url string) (*x509.RevocationList, float64, error) {
+	client := &http.Client{}
```
electron0zero: each probe can have proxy config, we should support that as well.

check for ProxyConfig and ensure that CRL check will also use the configured proxy for the checks.

if we don't add this, checks will pass but the CRL might fail because CRL checks are not using the proxy config.

prober/crl.go
```diff
@@ -0,0 +1,269 @@
...
+	)
+	crlFetchTime := prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "probe_ssl_crl_fetch_time_seconds",
+			Help: "Total time spent fetching CRLs in seconds",
+		},
+	)
+	crlNextPublish := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "probe_ssl_crl_next_publish",
```
electron0zero:
```suggestion
			Name: "probe_ssl_crl_next_update_timestamp_seconds",
```

to better match the existing metric names.

see https://prometheus.io/docs/practices/naming/#metric-names and  https://prometheus.io/docs/practices/instrumentation/#timestamps-not-time-since

CONFIGURATION.md
```diff
@@ -153,6 +153,9 @@ then a single address is selected to test, using the following logic:
   tls_config:
     [ <tls_config> ]
 
+  # Check each certificate in the TLS chain against its CRL and emit probe_ssl_crl_* metrics.
```
electron0zero, line 156: thinking about this, one thing come to my mind. When someone sets check_revoked to true, do they just expect the metrics or they also want the probe to fail when the Cert is revoked?

What is `check_revoked` meant to promise? A cert confirmed on its issuer's CRL still gives `probe_success=1`, but the config name reads like enforcement.

we have flags like `fail_if_ssl` / `fail_if_not_ssl` /  `fail_if_body_matches_regexp` for enforcement and failing the probe.

if your use-case and intent is to fail the probe, then we should add support for a fail_if_cert_revoked flag as well, and when that's set, we should fail the probe like we do for the fail_if_ssl and many other flags.

if you just want the metrics, then current setup is fine as is, and we can keep publishing the metrics and probe_success is left as is.

sebastiangaiser: Our use case is alerting on the metrics, with the revocation signal deliberately kept separate from `probe_success`:

- `probe_ssl_crl_revoked == 1` -> critical after 5m
- `probe_ssl_crl_available == 0` (restricted to `crl_url!=""`) -> warning after 15m
- `probe_ssl_crl_stale == 1` -> warning after 1h

`probe_success` then keeps its own meaning: the target is reachable and answered as expected. That separation is the useful part. A CRL distribution point being down is not the same outage as the service being down, and the two want different severities and different runbooks. If `check_revoked` also failed the probe, every dead or slow CRL responder would look like a total outage and we'd lose the ability to grade "revoked" (act now) differently from "revocation status unknown" (look into it).

For the same reason we alert on the per-phase `probe_http_duration_seconds` rather than the total `probe_duration_seconds` for latency, since the total includes the CRL fetch and a slow responder would otherwise inflate it.

You're right that the name alone doesn't convey that, so I've spelled the semantics out in `CONFIGURATION.md` for all four probes: the flag only reports revocation state and never fails the probe, a revoked certificate or an unreachable CRL leaves `probe_success` untouched, and alerting is expected to happen on the metrics. If you'd rather have that stated somewhere more prominent than the per-probe options, I'm happy to add a short section on the CRL metrics and how to alert on them.

And if hard enforcement turns out to be wanted later, a separate `fail_if_cert_revoked` can be added on top without changing this behaviour.

sebastiangaiser:
Thanks for the detailed second pass @electron0zero. I've pushed changes for all of it.

- Metric renamed to `probe_ssl_crl_next_update_timestamp_seconds` as suggested. I also renamed the corresponding struct field to `NextUpdate` so it lines up with `x509.RevocationList.NextUpdate`.
- `crlCertResult` is no longer exported, it was only ever used inside the package.
- `check_revoked` is now rejected when TLS is off, so the pair can't be configured in a way that silently does nothing.
- CRL fetches use the probe's proxy configuration, including `proxy_connect_header`.
- Test helpers moved to `prober/utils_test.go` and rebuilt on top of `generateCertificateTemplate`. `crl_test.go` no longer carries any fixture helpers of its own.
- `TestCheckChainCRL` and `TestRegisterCRLMetrics` are now table driven, which removed most of the duplication you pointed at.
- Added a root -> intermediate -> leaf test that gives each level its own CRL and asserts the intermediate is checked against the root's CRL.
- Dropped the stale comment about the CRL fetch timeout, that was left over from removing the timeout.

Three things where I deviated from the review, all for reasons worth flagging:

Proxy config only exists on the HTTP probe: Only `HTTPProbe` carries an `HTTPClientConfig` (which embeds `ProxyConfig`); TCP, Unix and gRPC only have a `TLSConfig`, so there is no proxy to honour there and they pass `nil`. Worth noting this changes behaviour slightly for HTTP: the fetch previously used Go's default transport, which picks up `HTTP_PROXY` from the environment. It now follows the probe's proxy configuration, so getting the environment proxy needs `proxy_from_environment: true`, the same as for the probe request itself. Let me know if you'd rather keep the old implicit behaviour.

TCP and Unix accept `check_revoked` with `tls: false` if a `query_response` step uses `starttls`: CRL checking runs on the STARTTLS-upgraded connection too, so rejecting purely on `tls: false` would have broken valid STARTTLS setups. gRPC has no STARTTLS path and is validated on `tls` alone.

The new certificate helper doesn't reuse `generateSignedCertificate`: That helper takes a `parentCert` but discards it (`generateCertificate` signs `template` with `template` as parent), so certificates come out with self-referential issuer names. That's fine for the existing tests but not for a chain test where each certificate has to be checked against its actual issuer's CRL, so `generateCRLTestCert` honours the parent. Happy to fix `generateCertificate` instead if you'd prefer that, though it touches tests beyond this PR.

CONFIGURATION.md
electron0zero: makes sense, I am good with current setup and we can always add `fail_if_cert_revoked` flag in the future.

electron0zero (approved):
lgtm, thanks for the PR :shipit:

PR #1599, opened 2026-06-13
refactor: replace interface{} with any (Go 1.18+)

MD-Mushfiqur123:
Replaces all remaining occurrences of `interface{}` with the `any` alias.


ArthurSens:
Hey, seems like this problem is already fixed on main. Thanks!

PR #1602, opened 2026-06-22
feat: support TCP responses that do not contain CR or LF markers

mkl262:
if a TCP response is matched by regex continue to next step, dont wait for end-of-line markers
Add basic test to query_response.go

```release-notes
[ENHANCEMENT] Handle TCP responses without end-of-line markers, continue on first regex match.
```

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/prober/query_response.go
+++ b/prober/query_response.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"io"
 	"log/slog"
 	"net"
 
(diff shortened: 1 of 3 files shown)


mkl262:
@bwplotka Are you ok with the changes copilot recommended? and are there any other changes you want me to make?

bwplotka:
Yes, those looks correct

mkl262:
I've committed the changes.

mkl262:
@bwplotka Are there any further changes that you think are needed?

dgl:
I believe this is a breaking change without a configuration to enable this option. For example with SMTP some servers split 220 greeting lines into multiple packets as an anti-spam measure, with this change the probe may match incorrectly.

prober/query_response.go
```diff
@@ -46,6 +47,62 @@ func probeExpectInfo(registry *prometheus.Registry, qr *config.QueryResponse, by
...
+					if match := re.FindSubmatchIndex(line); match != nil {
+						return line, match, nil
+					}
+					continue
+				}
+				line = append(line, data[:idx]...)
+				if len(line) > bufio.MaxScanTokenSize {
+					return line, nil, fmt.Errorf("read line exceeds %d bytes", bufio.MaxScanTokenSize)
+				}
+				data = data[idx+1:]
```
dgl, line 82: More fundamentally this is throwing away data if a read returns more than one line.

dgl (changes requested):
This just can't work as written. One way to make this work is for each `QueryResponse` have a flag that says whether this is binary or line based.

Using regexps to parse binary data is very https://xkcd.com/1313/ -- the main problem is a regexp can be greedy, but you don't have a delimiter here, so you don't know where the binary data ends. There's not an issue associated with this, not sure if discussions happened elsewhere; one way to handle this would be to only allow a binary regexp match as the final item in `query_responses`, but given the lack of an issue I'm not sure if that works for the use case here.

mkl262:
Hi @dgl, sorry for the delay, what is your recommendation? Should I create a new field in the queryResponse config to differentiate the two options?

PR #1624, opened 2026-07-20
Fix go vet ./... findings

randomizedcoder:
`go test`'s built-in vet subset does not run the `composites` or `unreachable` analyzers, so these `go vet ./...` findings are not caught by the normal test pipeline:

- **config**: drop the unreachable `return nil` after the exhaustive switch in `Module.UnmarshalYAML` (both the valid case and the default already return).
- **prober/dns**: use keyed fields for the `dns.Question` composite literal, so it is robust to field reordering in `github.com/miekg/dns`.
- **prober**: use keyed fields for the `config.HeaderMatch` literals in the HTTP header-match tests (17 occurrences).

No behavior change. `go vet ./...` is clean afterwards; `go build`, `go test ./...` pass.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -438,7 +438,6 @@ func (s *Module) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		// invalid
 		return fmt.Errorf("prober '%s' is not valid", s.Prober)
 	}
-	return nil
 }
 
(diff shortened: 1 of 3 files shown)


SuperQ (approved):
Thanks. I think I will followup with some additional linters in our golangci-lint config.

PR #1626, opened 2026-07-20
Set ReadHeaderTimeout on the HTTP server

randomizedcoder:
## Security: bound HTTP request-header reads (Slowloris / gosec G112)

The `&http.Server{}` handed to `web.ListenAndServe` left `ReadHeaderTimeout`
unset, so request-header reads were **unbounded** — a Slowloris-style
connection-exhaustion denial-of-service vector, and why `gosec` flags G112 here.
The exporter-toolkit does not set any server timeouts either, so nothing else
bounds it.

Set `ReadHeaderTimeout` to one minute via a small `newServer` constructor. It
bounds only header reading, **not** the probe handler, so legitimate probe
requests (which can run for the full scrape timeout) are unaffected. This closes
the DoS vector and lowers the exporter's network-facing attack surface. Happy to
adjust the value if a different bound is preferred.

A regression test drives a stalled-header client at the server and asserts the
connection is closed once the timeout elapses, while a well-formed request still
succeeds.

`go build`, `go vet`, `go test ./...` pass; `gosec` G112 clears.

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,6 +69,18 @@ func init() {
 	prometheus.MustRegister(versioncollector.NewCollector("blackbox_exporter"))
 }
 
+// serverReadHeaderTimeout bounds how long the server waits to read request
+// headers, mitigating Slowloris-style connection exhaustion. It covers only
+// header reading, not the probe handler, so a generous value is safe. It is a
+// package var so tests can shorten it.
+var serverReadHeaderTimeout = time.Minute
+
+// newServer builds the exporter's HTTP server with the hardening options the
+// exporter-toolkit does not set itself.
+func newServer() *http.Server {
+	return &http.Server{ReadHeaderTimeout: serverReadHeaderTimeout}
+}
+
 func main() {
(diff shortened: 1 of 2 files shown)


randomizedcoder:
Hi @ArthurSens — friendly ping on this one when you get a chance. It's a small hardening change: the `&http.Server{}` handed to `web.ListenAndServe` leaves `ReadHeaderTimeout` unset, so header reads are unbounded — a Slowloris-style connection-exhaustion vector (gosec G112). Bounding it to one minute covers only header reading (not the probe handler, which can run for the full scrape timeout), so legitimate probes are unaffected while the exporter's attack surface drops.

It's part of a small cluster of hardening PRs I've just rebased onto the latest master (all green in CI): #1632 enables `nilerr`, which caught the swallowed TLS-config error fixed in #1629, and #1625 replaces the predictable time-seeded ICMP IDs with `math/rand/v2`. Happy to tune the timeout or split anything out if you'd prefer. Thanks!

PR #1631, opened 2026-07-20
Enable govet and additional correctness linters in golangci-lint

randomizedcoder:
Follow-up to #1624. As noted there — "I think I will followup with some
additional linters in our golangci-lint config" — here is a proposed set.

### Why

The config runs only `misspell`, `revive`, `sloglint` and `staticcheck`.
`govet` is not enabled, and `go test`'s built-in vet subset skips the
`composites` and `unreachable` analyzers — which is exactly why the findings
fixed in #1624 were not caught by CI. (Verified: reintroducing those issues
makes this config flag both `unreachable` and `composites`.)

### What

Enable `govet` (with the `composites` analyzer) plus a curated, low-noise
correctness set: `copyloopvar`, `durationcheck`, `ineffassign`, `makezero`,
`unconvert`, `wastedassign`. The `wastedassign` (main) and `unconvert`
(http_test) findings these surface are fixed in the first commit.

`go build`, `go test ./...`, and `golangci-lint run` (pinned v2.11.4) all pass.

Happy to trim/extend the set to your preference. A follow-up will add `nilerr`
(it depends on a separate dropped-error fix, #1629).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,10 +2,24 @@ version: "2"
 linters:
   default: none
   enable:
+    - copyloopvar
+    - durationcheck
+    - govet
+    - ineffassign
+    - makezero
     - misspell
     - revive
     - sloglint
     - staticcheck
+    - unconvert
+    - wastedassign
+  settings:
+    govet:
+      # Enable the composites analyzer (unkeyed struct literals) on top of the
+      # default vet analyzers (which include unreachable). Neither is part of
+      # `go test`'s built-in vet subset, so they otherwise go uncaught in CI.
+      enable:
+        - composites
   exclusions:
     generated: lax
     presets:
--- a/main.go
+++ b/main.go
@@ -258,7 +258,7 @@ func run() int {
 			http.Error(w, "Probe module filter only works in conjunction with target parameter", http.StatusBadRequest)
 			return
 		}
-		result := new(prober.Result)
+		var result *prober.Result
 		if target != "" {
 			result = rh.GetByTargetAndModule(target, module)
 			if result == nil {
--- a/prober/http_test.go
+++ b/prober/http_test.go
@@ -186,7 +186,7 @@ func TestContentLength(t *testing.T) {
 			msg := testmsg
 			var buf bytes.Buffer
 			fw := brotli.NewWriter(&buf)
-			fw.Write([]byte(msg))
+			fw.Write(msg)
 			fw.Close()
 			return testdata{
 				msg:                    msg,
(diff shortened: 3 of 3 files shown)


SuperQ (commented):
I'm not a fan of enabling linters for the sake of enabling them. Especially if an LLM just does it with no thought or reasoning behind it.

randomizedcoder:
@SuperQ 

Nobody wants false positives

In this case, it seems like they are not false positives

https://github.com/prometheus/blackbox_exporter/pull/1626
https://github.com/prometheus/blackbox_exporter/pull/1629

Please correct me if I'm mistaken

randomizedcoder:
@SuperQ 

Sorry - I want to apologize

With this PR. I was hoping to help contribute, and so this PR goes to an extreme from the static analysis

We definitely don't need to go this far and it's just a suggestion

Happy to adjust the level up/down based on feedback

ArthurSens:
Hey @randomizedcoder, it looks like none of the 6 linters added with govet are doing anything in the codebase, so I'm not sure what benefit each one provides.

To unblock this PR, I would reduce the config to only govet, and we can discuss the benefits of all other linters in an issue or in [slack](https://cloud-native.slack.com/archives/C08CJ0A471D)

randomizedcoder:
Thanks @ArthurSens — that's a fair point. I've reduced the config to just `govet`, keeping the `composites` analyzer since that's the piece that actually adds value over `go test`'s built-in vet subset (it's what would have caught the #1624 findings that slipped through). The six extra correctness linters are dropped for now — happy to discuss their merits separately in an issue or in Slack, as you suggested.

I've also rebased the branch onto the latest master, so it's fully up to date. `go build`, `go test ./...`, and `golangci-lint run` all pass.

randomizedcoder:
@ArthurSens 
I've also updated the other security related PRs :pray: 
https://github.com/prometheus/blackbox_exporter/pulls?q=is%3Apr+state%3Aopen+author%3Arandomizedcoder

ArthurSens:
There are more changes in the PR than just the golangcilint fixes, could you make sure we only have related changes in a single PR?

.golangci.yml
```diff
@@ -2,11 +2,19 @@ version: "2"
...
     - revive
     - sloglint
     - staticcheck
+  settings:
+    govet:
+      # Enable the composites analyzer (unkeyed struct literals) on top of the
+      # default vet analyzers (which include unreachable). Neither is part of
+      # `go test`'s built-in vet subset, so they otherwise go uncaught in CI.
+      enable:
+        - composites
```
ArthurSens, line 17: Looks like composite is already enabled by default
https://github.com/golangci/golangci-lint/blob/6d2288e072e6f9c9bca28180cae9ce58a049c912/pkg/golinters/govet/govet.go#L124

Issue #1429, opened 2025-06-23
Next release could use a CA bundle update


ntavares:
Hi,

I just came across this, Sectigo has gone through a root CA migration:
https://www.sectigo.com/sectigo-public-root-cas-migration

This was done in a timely manner, but it seems that the blackbox image didn't catch up: the CA bundle is too old. Looking at the [mozilla ca bundle|https://curl.se/docs/caextract.html] I can see the new CA's are there, but they're not found in the latest blackbox-exporter image.

```
# Mozilla CA Bundle
$ openssl storeutl -noout -text -certs cacert-2025-05-20.pem  | grep Subject: | grep -i Sectigo

$ openssl storeutl -noout -text -certs cacert-2025-05-20.pem  | grep Subject: | grep -i Sectigo
        Subject: C=GB, O=Sectigo Limited, CN=Sectigo Public Server Authentication Root E46
        Subject: C=GB, O=Sectigo Limited, CN=Sectigo Public Server Authentication Root R46
```

```
# CA Bundle on the latest image:
$ openssl storeutl -noout -text -certs ca-certificates.crt  | grep Subject: | grep -i Sectigo
$ 
```

This is an easy one to fix,  not sure how I could help here?

ianhinder:
According to that Sectigo page,

> How will you ensure backwards compatibility with legacy systems?
> All our new Root CAs, have been cross-signed by both of our long standing Root CAs:
> 
> AAA Certificate Services
> USERTrust RSA Certification Authority (For RSA)
> USERTrust ECC Certification Authority (For ECC)
> Through these cross-signings, we extend the ubiquity of the new Root CAs, so they are also trusted on legacy systems that may not know about these new CA certificates, but do know about the long standing Root CAs mentioned above.

So it should still be possible to validate a cert signed up to the new root certs, since the new root certs are cross-signed with other existing root certs. Unless the library used in blackbox doesn't support this cross-signing?

winem:
Hi,

I noticed that newer Sectigo certificates no longer seem to be cross-signed which causes the blackbox-exporter to report the SSL handshake as failed:
```
level=ERROR source=http.go:474 msg="Error for HTTP request" module=http_get_with_accessToken target=https://www.neh.gov.ie/ err="Get \"https://54.230.114.27/\": tls: failed to verify certificate: x509: certificate signed by unknown authority"
```

An example is www.neh.gov.ie:
```
[...]

PR #1634, opened 2026-07-22
Update CA certificates in Docker image from Debian

locker95:
### Description

The busybox base image ships `/etc/ssl/certs/ca-certificates.crt`, but that bundle can lag behind current public roots. Probes against hosts using newer CAs (for example Sectigo Public Server Authentication Root R46) fail with `x509: certificate signed by unknown authority` even when desktop clients succeed.

This multi-stage Dockerfile installs `ca-certificates` from `debian:trixie-slim` and copies a fresh `ca-certificates.crt` into the image at build time, so each blackbox image build picks up current roots without depending on a busybox base rebuild cycle.

Related: #1429

Fixes #1547

### How to verify

```bash
# After building the image:
# docker run --rm --entrypoint cat <image> /etc/ssl/certs/ca-certificates.crt | openssl storeutl -noout -text -certs /dev/stdin | grep -i 'Sectigo Public Server Authentication Root'
```

```release-notes
[BUGFIX] Refresh CA certificates in the Docker image from Debian at build time #1547
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,14 @@
 ARG ARCH="amd64"
 ARG OS="linux"
+
+# Install a current CA bundle independently of the busybox base image rebuild
+# cycle. Blackbox probes remote TLS endpoints, so stale roots cause false
+# probe failures (see #1429, #1547).
+FROM debian:trixie-slim AS certs
+RUN apt-get update \
+	&& apt-get install -y --no-install-recommends ca-certificates \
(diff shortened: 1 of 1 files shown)


ArthurSens:
Hey, I believe this problem is already solved in the upstream prometheus/busybox image, but if not, that's where we would like to solve it, since it would fix the same problem for all exporters that depend on it.

That said, I'm closing this PR, but please open a new one in prometheus/busybox if this problem exists there.

PR #1105, opened 2023-08-02
allow specification of query_name by probe param for dns prober

Fe8N:
related issues: [#1104](https://github.com/prometheus/blackbox_exporter/issues/1104)
allow specification of query_name by probe param for dns prober

PR #1648, opened 2026-08-05
DNS: Make query_name module attribute optional

### LINKED ISSUE #1104: allow specification of query_name by probe param for dns prober
Currently dns prober only specify query_name by blackbox_exporter configuration file.

In some scenes,  I want to specify query_name by prometheus targets config.

I found in tcp and http prober, user can specify hostname by 'hostname' query param.

I think dns prober can reuse hostname query param  to specify query_name .

UiP9AV6Y:
this is similar to the PRs below with the following differences:

* added test coverage for new logic
* the query name is *not* blindly accepted from the *hostname* parameter, but only applied if the configuration does not define one. this is similar to how the HTTP and TCP probers treat the *hostname* parameter.

Closes #1483
Closes #1105

Operators can provide a DNS probe module config without an explicit `query_name` for the probe. This allows scrapers/consumer to provide the value via the `hostname` parameter, similar to the HTTP and TCP probers. Should no value be defined anywhere, the DNS query will use the target hostname instead.

Closes #1104

```release-notes
[ENHANCEMENT] Queryname for DNS probe can be provided via scrape URL parameter
```

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -312,7 +312,10 @@ tls_config:
 tls_config:
   [ <tls_config> ]
 
-query_name: <string>
+# Question to ask the DNS server. If empty, scrapers can provide
+# their own value via the *hostname* parameter, otherwise the
+# target hostname is used as default.
+[ query_name: <string> ]
 
 [ query_type: <string> | default = "ANY" ]
 [ query_class: <string> | default = "IN" ]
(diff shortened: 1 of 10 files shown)


ArthurSens (commented):
I support the idea, with one caveat. For HTTP/TCP probes, I think it's fine to fall back to the target address if the hostname is missing, since that HTTP/TCP address is almost always the host being probed, but that's not the common scenario for DNS servers.

For DNS, I'd rather fail the scrape if neither the module nor the URL param specifies the DNS query name. Probing a DNS server and a DNS target are completely different things and could be a silent failure for folks that aren't paying attention

PR #1652, opened 2026-08-10
prober: add subjectkeyid label to probe_ssl_last_chain_info

### LINKED ISSUE #1581: Add SKI as a label to HTTPS certificate labels and allow to disable fingerprint and serialnumber
Today the **fingerprint** and the **serialnumber** of the certificate are used as labels.
Given that certificate lifetime will be reduced soon (it already started) those two labels are changing more often.
Today already with ACME issued certs (e.g. Let's Encrypt) the certificate lifetime is 90 days or less.

What would be a more stable identifier is the subject public key (SKI). 
It only changes when the key changes and from security point of view this would be the more interesting change.

From security point of view a certificate renewal isn't that important. 
For a certificate mainly the expiration is interesting.

IMHO it would be more important to track key changes then certificate changes.
Specially because those changes will happen more often in future and more often generates stale labels.

Because some integrations probably rely on those labels, the idea would be to make it configurable and not change what is available today.
This would be an enhancement request. Not a bug. What is there works perfectly fine. But could be improved.

Example:

```
probe_ssl_last_chain_info{fingerprint_sha256="ff0e64257497f17ba6a8060b261a1376b1a2d92ae3fcf6b3cddf95ab57aa3bbc",issuer="CN=E8,O=Let's Encrypt,C=US",serialnumber="05d8a38847e7691982f2a1c41e31bc319b84",subject="CN=*.example.com",subjectalternative="*.example.com,example.com"} 1
```

locker95:
Leaf cert fingerprint and serial change every reissue (often every 90 days with ACME). The Subject Key Identifier only moves when the key changes, so it is a better long-lived join key for cert inventory-style dashboards.

Adds a `subjectkeyid` label (hex of `SubjectKeyId`, empty if the extension is missing) on `probe_ssl_last_chain_info` for HTTP, TCP/unix, and gRPC probes.

Leaving fingerprint/serial as-is for now so existing rules keep working; happy to follow up with knobs to drop them if that is still useful.

Fixes #1581

### DIFF
--- a/prober/grpc.go
+++ b/prober/grpc.go
@@ -112,7 +112,7 @@ func ProbeGRPC(ctx context.Context, target string, module config.Module, registr
 				Name: "probe_ssl_last_chain_info",
 				Help: "Contains SSL leaf certificate information",
 			},
-			[]string{"fingerprint_sha256", "subject", "issuer", "subjectalternative", "serialnumber"},
+			[]string{"fingerprint_sha256", "subject", "issuer", "subjectalternative", "serialnumber", "subjectkeyid"},
 		)
 	)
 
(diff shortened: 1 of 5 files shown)


prober/grpc.go
```diff
@@ -112,7 +112,7 @@ func ProbeGRPC(ctx context.Context, target string, module config.Module, registr
 				Name: "probe_ssl_last_chain_info",
 				Help: "Contains SSL leaf certificate information",
 			},
-			[]string{"fingerprint_sha256", "subject", "issuer", "subjectalternative", "serialnumber"},
+			[]string{"fingerprint_sha256", "subject", "issuer", "subjectalternative", "serialnumber", "subjectkeyid"},
```
electron0zero, line 115: can you explain the use cases where `subject` is not enough and you need the subject key id? just trying to see what is the end goal here

also please keep in mind that Blackbox exporter is an monitoring tool and not a security tool.

locker95:
yeah this isn't meant as a security control — just a more stable identity for dashboards.

the case I keep hitting is ACME/LE rotation: fingerprint and serial change every ~90 days even when the same key is reused. subject usually stays put too, but it isn't unique (two hosts with the same CN collide). SKI only moves when the key changes, so you can group those renewals as one series instead of a new one every quarter.

empty string if the cert has no SKI, so existing probes don't break. if that still feels too inventory-ish I can drop it.

electron0zero:
thanks, I think it's better to leave it out for now because I don't see a strong use case of this.

locker95:
fair enough — i'll close this then.

Issue #295, opened 2018-02-22
Feature request: toggle to suppress newlines in tcp query_response mode


h3po:
### Host operating system:
docker prom/blackbox-exporter:latest

### blackbox_exporter version:
```
blackbox_exporter, version 0.11.0 (branch: HEAD, revision: 200181ba84e595570eeed8cb1848491d3d4f85f0)
  build user:       root@3db6abb4b9fb
  build date:       20171128-13:47:58
  go version:       go1.9.2
```
### What is the blackbox.yml module config.
```
modules:
  ldap:
    prober: tcp
    timeout: 5s
    tcp:
      preferred_ip_protocol: "ip4"
      query_response:
        - send: !!binary "ME4CAQFgSQIBAwQyQ049cHJvbWV0aGV1cyxPVT1Vc2VycywtY2Vuc29yZWQtgBAtY2Vuc29yZWQt"
        - expect: !!binary "MAwCAQFhBwoBAAQABAA="
        - send: !!binary "MB4CAQJ3GYAXMS4zLjYuMS40LjEuNDIwMy4xLjExLjM="
        - expect: !!binary "MEMCAQJ4PgoBAAQABACLNWRuOmNuPXByb21ldGhldXMsb3U9VXNlcnMsLWNlbnNvcmVkLQ=="
        - send: !!binary "MAUCAQNCAA=="

```
### What did you do that produced an error?
I am aware the send/response feature was not intended for binary data. I'm trying here to do a simple bind with an ldap server and then issue a "whoami" to verify that the server is working. It looks to me like this could work, but the query_response mode is appending a newline character (\x0a) to each send: line which breaks the protocol

### What did you expect to see?
I'd like to be able to toggle the automatic newline off.

SuperQ:
This seems reasonable to me.. I'd almost prefer that we don't mangle the `send` string at all, and require that configurations include the `\n` in them explicitly.

brian-brazil:
This feature is only intended for line-based protocols, and your requested feature does not make it support binary protocols - it'll just break at the next step.

You'll want to write a exporter for this.

h3po:
@brian-brazil 
>it'll just break at the next step.
What do you think exactly would break? I'm fine with writing my own exporter, but I don't see why it couldn't be handled by blackbox. The only difference between a line protocol like smtp and ldap in this example is that ldap uses asn1 encoding instead of ascii.

brian-brazil:
> What do you think exactly would break?

[...]

PR #1659, opened 2026-08-27
prober: add `skip_newline_on_send` option

### LINKED ISSUE #1657: TCP: Support for send_bytes
## Current status

`expect_bytes` as been implemented into #1441 enabling to monitor postgresql database (see #801).

default config now propose the following config to check if a postgresql is accessible (introduce by #1441):

```yaml
postgresql:
    prober: tcp
    tcp:
      query_response:
      - send: !!binary AAAACATSFi8= # 0x00, 0x00, 0x00, 0x08, 0x04, 0xD2, 0x16, 0x2F - PostgreSQL SSLRequest
      - expect_bytes: S # 0x53 - Reply will be 'S' if SSL is enabled, and 'N' if it is not.
      - starttls: true
```

## The Issue

This test is buggy as blackbox-exporter send a `/n` on each `send` instruction.

therefore blackbox-exporter send `0x00`, `0x00`, `0x00`, `0x08`, `0x04`, `0xD2`, `0x16`, `0x2F`, **`0x0A`** instated of `0x00`, `0x00`, `0x00`, `0x08`, `0x04`, `0xD2`, `0x16`, `0x2F`.

Making postgresql failed after TLS check (has it's have an extra byte that he can't parse) and that is actually not encrypted).

### `send_bytes`

Introduce the `send_bytes` config that send the given bytes without any added `\n`.

This option will be mutually exclusive with the `send` config (like the `expect_bytes` and the `expect` configuration)

### `add_trailing_new_line`

Introduce the `add_trailing_new_line` configuration.

`add_trailing_new_line` configuration will be a boolean.

`add_trailing_new_line` will be `true` by default for backward compatibility.

This option will control if a new line will be added at the end of the `send` payload.

Note: An old issue #295 already proposed more or less that.

JulesdeCube:

Fixes #1657 and #295

```release-notes
[FEATURE] Add support to skip adding new line at end of send data
```

**Checklist**
- [X] Tests updated
- [X] Documentation added

cc: @electron0zero

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -227,6 +227,33 @@ regexp: <regex>,
 [ allow_missing: <boolean> | default = false ]
 ```
 
+### `<query_response>`
+
+```yml
+
+# The query sent in the TCP, WebSocket, or Unix socket probe and the expected associated response.
+# "expect" matches a regular expression; it is mutually exclusive with "expect_bytes".
+# "expect_bytes" does exact byte-by-byte match; it is mutually exclusive with "expect".
+# "labels" can define labels which will be exported on metric "probe_expect_info";
+# "send" sends some content;
+# "skip_newline_on_send" controls whether a '\n' byte is appended to the "send" content, omitting the newline when set to true.
+# "send" and "labels.value" can contain values matched by "expect" (such as "${1}");
+# "starttls" upgrades TCP or Unix socket connection to TLS.
+query_response:
+  [ - [ [ expect: <string> ],
+        [ expect_bytes: <string> ],
+        [ skip_newline_on_send: <boolean | default = false>]
+        [ labels:
+          - [ name: <string>
+              value: <string>
+            ], ...
+        ],
+        [ send: <string> ],
+        [ starttls: <boolean | default = false> ]
+      ], ...
+  ]
+```
+
 ### `<tcp_probe>`
 
 ```yml
@@ -239,24 +266,7 @@ regexp: <regex>,
 [ source_ip_address: <string> ]
 
 # The query sent in the TCP probe and the expected associated response.
-# "expect" matches a regular expression; it is mutually exclusive with "expect_bytes".
-# "expect_bytes" does exact byte-by-byte match; it is mutually exclusive with "expect".
-# "labels" can define labels which will be exported on metric "probe_expect_info";
-# "send" sends some content;
-# "send" and "labels.value" can contain values matched by "expect" (such as "${1}");
-# "starttls" upgrades TCP connection to TLS.
-query_response:
-  [ - [ [ expect: <string> ],
-        [ expect_bytes: <string> ],
-        [ labels:
-          - [ name: <string>
-              value: <string>
-            ], ...
-        ],
-        [ send: <string> ],
-        [ starttls: <boolean | default = false> ]
-      ], ...
-  ]
+query_response: <query_response>
 
 # Whether or not TLS is used when the connection is initiated.
 [ tls: <boolean | default = false> ]
@@ -278,22 +288,7 @@ tls_config:
 ```yml
 
 # The query sent in the unix socket probe and the expected associated response.
-# "expect" matches a regular expression;
-# "labels" can define labels which will be exported on metric "probe_expect_info";
-# "send" sends some content;
-# "send" and "labels.value" can contain values matched by "expect" (such as "${1}");
-# "starttls" upgrades connection to TLS.
-query_response:
-  [ - [ [ expect: <string> ],
-        [ labels:
-          - [ name: <string>
-              value: <string>
-            ], ...
-        ],
-        [ send: <string> ],
-        [ starttls: <boolean | default = false> ]
-      ], ...
-  ]
+query_response: <query_response>
 
 # Whether or not TLS is used when the connection is initiated.
 [ tls: <boolean | default = false> ]
@@ -423,8 +418,8 @@ validate_additional_rrs:
 
 # gRPC client metadata
 metadata:
-  [ <string>: 
-    [ - <string> ], ...  
+  [ <string>:
+    [ - <string> ], ...
   ], ...
 
 # Whether to connect to the endpoint with TLS.
@@ -458,14 +453,7 @@ headers:
   ], ...
 
 # The query sent after connection upgrade and the expected associated response.
-# "expect" matches a regular expression against incoming messages;
-# "send" sends a message (can use values matched by "expect" such as "${1}").
-query_response:
-  [ - [ expect: <regex> ],
-        [ send: <string> ],
-        [ starttls: <boolean | default = false> ]
-      ], ...
-  ]
+query_response: <query_response>
 
   # The IP protocol of the Websocket probe (ip4, ip6).
   [ preferred_ip_protocol: <string> | default = "ip6" ]
--- a/blackbox.yml
+++ b/blackbox.yml
@@ -74,5 +74,6 @@ modules:
     tcp:
       query_response:
       - send: !!binary AAAACATSFi8= # 0x00, 0x00, 0x00, 0x08, 0x04, 0xD2, 0x16, 0x2F - PostgreSQL SSLRequest
+        skip_newline_on_send: true # Don't send extra new line to prevent received unencrypted data after SSL request
       - expect_bytes: S # 0x53 - Reply will be 'S' if SSL is enabled, and 'N' if it is not.
       - starttls: true
--- a/config/config.go
+++ b/config/config.go
@@ -354,11 +354,12 @@ type Label struct {
 }
 
 type QueryResponse struct {
-	Expect      Regexp  `yaml:"expect,omitempty" json:"expect,omitzero"`
-	ExpectBytes string  `yaml:"expect_bytes,omitempty" json:"expect_bytes,omitempty"`
-	Labels      []Label `yaml:"labels,omitempty" json:"labels,omitempty"`
-	Send        string  `yaml:"send,omitempty" json:"send,omitempty"`
-	StartTLS    bool    `yaml:"starttls,omitempty" json:"starttls,omitempty"`
+	Expect            Regexp  `yaml:"expect,omitempty" json:"expect,omitzero"`
+	ExpectBytes       string  `yaml:"expect_bytes,omitempty" json:"expect_bytes,omitempty"`
+	Labels            []Label `yaml:"labels,omitempty" json:"labels,omitempty"`
+	Send              string  `yaml:"send,omitempty" json:"send,omitempty"`
+	SkipNewlineOnSend bool    `yaml:"skip_newline_on_send,omitempty" json:"skip_newline_on_send,omitempty"`
+	StartTLS          bool    `yaml:"starttls,omitempty" json:"starttls,omitempty"`
 }
 
 type TCPProbe struct {
--- a/example.yml
+++ b/example.yml
@@ -271,7 +271,9 @@ modules:
         tls_config:
           insecure_skip_verify: true
       headers:
-        X-Some-Header: "my_header"
+        X-Some-Header:
+          values:
+          - "my_header"
       query_response:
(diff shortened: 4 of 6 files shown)


prober/query_response.go
```diff
@@ -166,9 +165,14 @@ func probeQueryResponses(ctx context.Context, target string, conn net.Conn, modu
...
 			probeFailedDueToBytes.Set(0)
 		}
-		if send != "" {
-			logger.Debug("Sending line", "line", send)
-			if _, err := fmt.Fprintf(conn, "%s\n", send); err != nil {
+		if len(send) > 0 {
+			if qr.SkipNewlineOnSend {
+				logger.Debug("Sending data", "data", send)
+			} else {
+				logger.Debug("Sending line", "line", send)
```
electron0zero: this logline is now wrong, you are sending on line 175, this code is just appending newline to payload.

I think you can just log "skipping newline on send if the config is enabled" or just log the value of `skip_newline_on_send` next to line 175 and write a better log message.

blackbox.yml
```diff
@@ -74,5 +74,6 @@ modules:
     tcp:
       query_response:
       - send: !!binary AAAACATSFi8= # 0x00, 0x00, 0x00, 0x08, 0x04, 0xD2, 0x16, 0x2F - PostgreSQL SSLRequest
+        skip_newline_on_send: true # Don't send extra new line to prevent received unencrypted data after SSL request
```
electron0zero, line 77: add a new example with this enabled in examples.yml file as well.

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 BREAKING CHANGES:
 
 * [CHANGE]
-* [FEATURE]
+* [FEATURE] Add support to skip adding new line at end of send data #1659
```
electron0zero: only mention it in PR desc, it can be skipped here.

prober/query_response.go
JulesdeCube: i keep the "Sending data" and move it at line 175

prober/tcp_test.go
```diff
@@ -60,6 +61,128 @@ func TestTCPConnection(t *testing.T) {
 	<-ch
 }
 
+func TestTCPConnectionSendLines(t *testing.T) {
```
electron0zero: there is a massive overlap in both tests, I think these two tests can be merged into a single table driven test so it's easy to read and maintain

blackbox.yml
JulesdeCube: i add the `tcp_mqtt_health` check (tested against `test.mosquitto.org:1883`).

also in bonus a small fix on websocket example, headers value should be an object with values, secrets or files (see https://github.com/prometheus/common/blob/main/config/headers.go#L83)

CHANGELOG.md
JulesdeCube: removed

prober/tcp_test.go
JulesdeCube: fix (merge both test into a unique `TestTCPConnectionSend` test)

example.yml
```diff
@@ -249,7 +271,9 @@ modules:
         tls_config:
           insecure_skip_verify: true
       headers:
-        X-Some-Header: "my_header"
+        X-Some-Header:
+          values:
+          - "my_header"
```
ArthurSens, line 276: This is a good fix, but it looks unrelated to the PR. Could we split this into a separate one?

prober/tcp_test.go
```diff
@@ -60,6 +61,83 @@ func TestTCPConnection(t *testing.T) {
...
+				{Send: "test line.", SkipNewlineOnSend: true},
+				{Send: "test other line.", SkipNewlineOnSend: true},
+			},
+			sentenceDelimiter: '.',
+			answers:           []string{"test line.", "test other line."},
+		},
+	}
+
+	for _, testCase := range tests_cases {
+		ln, err := net.Listen("tcp", "localhost:0")
```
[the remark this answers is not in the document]
JulesdeCube: fixed

CONFIGURATION.md
```diff
@@ -227,27 +227,21 @@ regexp: <regex>,
...
 # "labels" can define labels which will be exported on metric "probe_expect_info";
 # "send" sends some content;
+# "skip_newline_on_send" control if a '\n' byte is append to the "send" content.
 # "send" and "labels.value" can contain values matched by "expect" (such as "${1}");
-# "starttls" upgrades TCP connection to TLS.
+# "starttls" upgrades TCP or unix socket connection to TLS.
 query_response:
   [ - [ [ expect: <string> ],
         [ expect_bytes: <string> ],
+        [ skip_newline_on_send: <boolean | default = false>]
```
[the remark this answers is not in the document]
JulesdeCube: disambiguated + added that it can also be used for websocket

prober/tcp_test.go
```diff
@@ -60,6 +61,83 @@ func TestTCPConnection(t *testing.T) {
...
+					t.Errorf("Failed to read from connection (answer %d): %s", i, err)
+					return
+				}
+				if receivedFirstLine != answer {
+					t.Errorf("Expected %q, got %q", answer, receivedFirstLine)
+				}
+			}
+
+			ch <- struct{}{}
+		}()
```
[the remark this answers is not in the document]
JulesdeCube: fix (use sub case)

example.yml
JulesdeCube: Hello,
the fix seems so small that i thought that to split into an other MR will must make pointless noisy.
I can remove it from this MR if needed (waiting for your answer).

JulesdeCube:
hello,
sorry for the message but i think i answer all your questions.
What prevent this PR to been merge ?

JulesdeCube:
Hello @electron0zero what is blocking this MR ?

JulesdeCube:
hello @electron0zero, i know that this project is opensource and without any warranties and that you have total right to refuse this MR but can i have an "official" statement about this PR ? will it ever been merged ?

CONFIGURATION.md
```diff
@@ -227,27 +227,22 @@ regexp: <regex>,
...
-# The source IP address.
-[ source_ip_address: <string> ]
-
-# The query sent in the TCP probe and the expected associated response.
+# The query sent in the TCP, WebSocket, or Unix socket probe and the expected associated response.
 # "expect" matches a regular expression; it is mutually exclusive with "expect_bytes".
 # "expect_bytes" does exact byte-by-byte match; it is mutually exclusive with "expect".
 # "labels" can define labels which will be exported on metric "probe_expect_info";
 # "send" sends some content;
+# "skip_newline_on_send" controls whether a '\n' byte is appended to the "send" content, omitting the newline when set to true.
```
electron0zero, line 239: The websocket section now points at the shared block. That block documents expect_bytes, labels (exported as probe_expect_info) and skip_newline_on_send, and the websocket prober ignores all 3 (see: prober/websocket.go:142-184). 

A websocket step with only a mismatching expect_bytes returns success, so the probe passes without reading anything. expect also behaves differently. Websocket checks only the next message, while TCP and Unix skip lines until one matches. 

undo the shared `query_response` so each check type has it's own block like before, and we don't make similar mistake when documenting again.

example.yml
electron0zero, line 253: thi example needs `skip_newline_on_send` to pass the test, so add it here as well.

example.yml
```diff
@@ -178,6 +178,28 @@ modules:
...
+        # Sends a minimal MQTT v3.1.1 CONNECT packet (Client ID: "test")
+        # \x10 = CONNECT command
+        # \x10 = Remaining length (16 bytes)
+        # \x00\x04MQTT = Protocol length (4) + Name
+        # \x04 = Protocol Level
+        # \x02 = Connect flags (Clean Session)
+        # \x00\x3C = Keep Alive (60s)
+        # \x00\x04test = Client ID length (4) + ID
+        - send: "\x10\x10\x00\x04MQTT\x04\x02\x00\x3C\x00\x04test"
+          # Prevents the \n injection that would break the MQTT frame length
```
electron0zero, line 195: can you explain this comment? how and why does it break the MQTT fram length?

example.yml
```diff
@@ -178,6 +178,28 @@ modules:
...
+      query_response:
+        # Sends a minimal MQTT v3.1.1 CONNECT packet (Client ID: "test")
+        # \x10 = CONNECT command
+        # \x10 = Remaining length (16 bytes)
+        # \x00\x04MQTT = Protocol length (4) + Name
+        # \x04 = Protocol Level
+        # \x02 = Connect flags (Clean Session)
+        # \x00\x3C = Keep Alive (60s)
+        # \x00\x04test = Client ID length (4) + ID
+        - send: "\x10\x10\x00\x04MQTT\x04\x02\x00\x3C\x00\x04test"
```
electron0zero, line 194: use `!!binary` to correctly configure the bytes you want to send, you see that in the postgresql example.

YAML \x escapes give raw bytes only below 0x80. With go.yaml.in/yaml/v3, "\x80" decodes to the 2 bytes c2 80.

so this example is not a good example, it works here because we are not sending anything abobe \x80. so using !!binary is the safe and best way.

example.yml
```diff
@@ -178,6 +178,28 @@ modules:
       tls_config:
         insecure_skip_verify: true
         ca_file: "/etc/blackbox_exporter/CA_cert.crt"
+  tcp_mqtt_health:
+    prober: tcp
+    timeout: 5s
+    tcp:
+      query_response:
+        # Sends a minimal MQTT v3.1.1 CONNECT packet (Client ID: "test")
```
electron0zero, line 186: can this example comment be trimmed? we don't need to explain the MQTT packets here.

prober/tcp_test.go
```diff
@@ -60,6 +61,90 @@ func TestTCPConnection(t *testing.T) {
 	<-ch
 }
 
+func TestTCPConnectionSend(t *testing.T) {
+	testsCases := []struct {
```
electron0zero, line 65:
```suggestion
	tests := []struct {
```

nit: match the existing tests.

prober/tcp_test.go
```diff
@@ -60,6 +61,90 @@ func TestTCPConnection(t *testing.T) {
...
+			sentenceDelimiter: '\n',
+			answers:           []string{"test line\n", "test other line\n"},
+		},
+		{
+			name: "dot delimiter",
+			query: []config.QueryResponse{
+				{Send: "test line.", SkipNewlineOnSend: true},
+				{Send: "test other line.", SkipNewlineOnSend: true},
+			},
+			sentenceDelimiter: '.',
```
electron0zero, line 86: why are we doing this? I am unable to make sense of this?

prober/tcp_test.go
```diff
@@ -60,6 +61,90 @@ func TestTCPConnection(t *testing.T) {
...
+				conn, err := ln.Accept()
+				if err != nil {
+					errCh <- fmt.Errorf("error accepting on socket: %w", err)
+					return
+				}
+				defer conn.Close()
+
+				reader := bufio.NewReader(conn)
+				for i, answer := range tc.answers {
+					receivedFirstLine, err := reader.ReadString(tc.sentenceDelimiter)
```
electron0zero, line 123: same, why can't we read the response and then compare it with expected?

config/config.go
electron0zero, line 615: Let's disallow `skip_newline_on_send` on a step that has no `send`. The flag only affects the `send` in its own step, so on a step by itself it does nothing. One extra dash is enough to cause this:

```go
if s.SkipNewlineOnSend && s.Send == "" {
	return errors.New("skip_newline_on_send cannot be used without send")
}
```

This matches how `check_revoked` without TLS is rejected. 

The field is new, so no existing config breaks. 

Please also add a case to `TestLoadBadConfigs`, like `invalid-tcp-check-revoked-without-tls.yml`.

electron0zero (changes requested):
few more questions and changed are needed.

also, we prefer that you ask for reviews in the slack (CNCF slack, in #prometheus-exporters-dev channel) instead of tagging maintainers in the PRs.

PR #1660, opened 2026-08-28
Fix Basic Auth being stripped on same-host redirects

### LINKED ISSUE #1656: Basic Auth is stripped on same-host redirects after the target hostname is rewritten to an IP
### Summary

Basic Auth is stripped from same-host redirects after Blackbox Exporter adopted the cross-host credential protection from `prometheus/common`.

Blackbox Exporter rewrites the initial request URL host to the resolved IP address. `prometheus/common` then compares that IP address with the hostname in the redirect URL, incorrectly classifies the redirect as cross-host, and removes the credentials.

### Deployment

Blackbox Exporter runs in Docker using an image built from `master` after prometheus/blackbox_exporter#1608 upgraded `github.com/prometheus/common` from 0.68.0 to 0.70.0.

### Module configuration

```yaml
modules:
  http_2xx_auth:
    prober: http
    timeout: 10s
    http:
      preferred_ip_protocol: ip4
      valid_http_versions:
        - HTTP/1.1
        - HTTP/2.0
      method: GET
      fail_if_not_ssl: true
      basic_auth:
        username: probe-user
        password: <redacted>
```

### Example target

```text
https://service.example.com/old-path
```

The target returns a same-host redirect:

```text
HTTP/2 302
Location: https://service.example.com/new-path
```

The redirected URL returns HTTP 200 when the same Basic Auth credentials are sent.

### Observed behavior

Blackbox Exporter authenticates the initial request and follows one redirect. The redirected request does not contain the Basic Auth credentials and returns HTTP 401.

Relevant debug output:

```text
level=ERROR msg="Invalid HTTP response status code, wanted 2xx"
module=http_2xx_auth
target=https://service.example.com/old-path
status_code=401

probe_http_redirects 1
probe_http_status_code 401
probe_success 0
```

The module output confirms that redirects are enabled:

```yaml
basic_auth:
  username: probe-user
  password: <secret>
follow_redirects: true
```

An equivalent curl request succeeds:

```console
curl --location --user 'probe-user:<redacted>' \
  https://service.example.com/old-path
```

Result:

```text
redirects=1
final_status=200
```

### Expected behavior

Blackbox Exporter should retain credentials when a redirect stays on the same logical host. It should strip credentials when a redirect points to another host.

### Root cause analysis

Blackbox Exporter resolves the target itself to select an address according to `preferred_ip_protocol` and report DNS and IP probe metrics.

After resolution, the HTTP prober replaces `targetURL.Host` with the selected IP address. It preserves the original hostname in `request.Host` and TLS SNI.

The first request is therefore represented internally as:

```text
URL host: 192.0.2.10
Host header: service.example.com
```

The server returns this redirect:

```text
https://service.example.com/new-path
```

`prometheus/common` 0.69 added `isCrossHostRedirect` to prevent credentials from reaching another host. The function compares `req.URL.Hostname()` values in the redirect chain. It does not consider the original `request.Host`.

The comparison therefore becomes:

```text
192.0.2.10 != service.example.com
```

The code incorrectly classifies the same-host redirect as cross-host. The Basic Auth round tripper then omits the credentials from the redirected request.

This regression results from the interaction of two behaviors:

1. Blackbox Exporter replaces the hostname in the initial URL with the selected IP address.
2. `prometheus/common` treats a change in the URL hostname as a cross-host redirect.

### Regression range

The behavior changed after these upstream changes:

- prometheus/common#901 added credential stripping for cross-host redirects.
- prometheus/common#920 made cross-host detection sticky across the redirect chain.
- prometheus/blackbox_exporter#1608 upgraded `prometheus/common` from 0.68.0 to 0.70.0.

Related issue prometheus/blackbox_exporter#618 describes problems caused by comparing the resolved IP address with the redirect hostname. That issue predates the credential-stripping change and does not cover this authentication regression.

### Version comparison

The same module and scenario work with Blackbox Exporter `v0.28.0`:

```text
probe_http_redirects 1
probe_http_status_code 200
probe_success 1
```

### Suggested fix

Keep the logical hostname in `request.URL` and use a custom `DialContext` to connect to the selected IP address.

Alternatively, pass the original logical hostname to the redirect security logic. The redirect check should compare the destination with the original target hostname or `request.Host`, not the rewritten IP address.

The fix must continue to strip credentials from genuine cross-host redirects.

### Related issues and changes

- https://github.com/prometheus/blackbox_exporter/issues/618
- https://github.com/prometheus/common/pull/901
- https://github.com/prometheus/common/pull/920
- https://github.com/prometheus/common/pull/921
- https://github.com/prometheus/blackbox_exporter/pull/1608

AI Assisted by openai/gpt-5.6-sol/OpenCode

neoLsH:
## What

Fixes #1656.

Blackbox rewrites the probe request URL host to the resolved IP address. When
the target then issues a same-host redirect (e.g. `https://host/old` ->
`https://host/new`), the redirect's hostname no longer matches the rewritten
IP, so the request is classified as a cross-host redirect and the Basic Auth
credentials are dropped, resulting in a 401.

## How

In `CheckRedirect`, when Basic Auth is configured and the redirect stays on the
same logical host (the original target hostname, or the IP it resolved to),
re-apply the credentials to the redirect request. Cross-host redirects still
never receive credentials, preserving the security behavior.

## Testing

Added `TestBasicAuthPreservedOnSameHostRedirect`, which fails on `master`
(401 after the redirect) and passes with this change. Existing HTTP/auth tests
are unaffected.

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -485,6 +485,17 @@ func ProbeHTTP(ctx context.Context, target string, module config.Module, registr
 			logger.Info("Not following redirect")
 			return errors.New("don't follow redirects")
 		}
+
+		// The request URL host is rewritten to the resolved IP below, so a
+		// redirect back to the original hostname is seen as cross-host and the
+		// credentials are dropped. Re-apply Basic Auth when the redirect stays
+		// on the same logical host.
+		if ba := httpConfig.HTTPClientConfig.BasicAuth; ba != nil {
+			rh := strings.ToLower(r.URL.Hostname())
+			if rh == strings.ToLower(targetHost) || (ip != nil && rh == strings.ToLower(ip.String())) {
+				r.SetBasicAuth(ba.Username, string(ba.Password))
+			}
+		}
 		return nil
 	}
 
--- a/prober/http_test.go
+++ b/prober/http_test.go
@@ -806,6 +806,48 @@ func TestBasicAuth(t *testing.T) {
 	}
 }
 
+func TestBasicAuthPreservedOnSameHostRedirect(t *testing.T) {
+	// Reproduces https://github.com/prometheus/blackbox_exporter/issues/1656:
+	// the probe dials the resolved IP but must keep the hostname in the request
+	// URL, otherwise a same-host redirect is classified as cross-host and the
+	// Basic Auth credentials are stripped, yielding a 401.
+	mux := http.NewServeMux()
+	ts := httptest.NewServer(mux)
+	defer ts.Close()
+
+	u, err := url.Parse(ts.URL)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	mux.HandleFunc("/old", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, "http://localhost:"+u.Port()+"/new", http.StatusFound)
+	})
+	mux.HandleFunc("/new", func(w http.ResponseWriter, r *http.Request) {
+		user, pass, ok := r.BasicAuth()
+		if !ok || user != "username" || pass != "password" {
+			http.Error(w, "unauthorized", http.StatusUnauthorized)
+			return
+		}
+	})
+
+	registry := prometheus.NewRegistry()
+	testCTX, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	result := ProbeHTTP(testCTX, "http://localhost:"+u.Port()+"/old",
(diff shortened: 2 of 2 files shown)


pgassmann:
This fix is ready without feedback for 3 weeks. Can a maintainer have a look at this regression-fix and plan a new release? Thanks.

neoLsH:
Hi @SuperQ , do you have time to review this PR? Thanks a lot!

PR #1675, opened 2026-09-04
Add the 'valid_tls_alert_codes' configuration option to the 'http' an…

### LINKED ISSUE #1674: Feature Request: Add capacity to consider a list of TLS Alerts as expected and to therefore succeed if and only if they happen
It is not so uncommon to protect a site behind a mutually authenticated TLS authentication _(mTLS for short)_ that requires a client authentication.

In most cases this is implemented by a reverse proxy and depending on the tool used, misconfiguration can rapidly and silently open your otherwise well protected site into an open spot for anyone to crawl into.

When I monitor such services or hosts, I want to make sure that:
1. They are indeed up and are responding
2. Their server certificates are indeed valid, not expired, etc... _(I want to use all the perks that the blackbox exporter already offers)_
3. They reject my HTTPS connection with a proper/expected TLS alert.

As far as I understand currently blackbox exporter fails short in the latter bullet point.

I understand that I can negate the result of a query _(i.e. consider it is successful when it fails)_ but this is not what I want there.

Indeed, since a mutually authenticated TLS starts by receiving and verifying the server certificate chain, I don't see why blackbox exporter would have to fail here and not report other useful information like the server certificates expiry information etc...

I propose to add a `valid_tls_alert_codes` array, both in the TCP and HTTP prober configurations and define there a list of acceptable TLS alerts that we would consider successful responses:

```yaml
modules:
  mtls_enforcement_tcp:
    prober: tcp
    tcp:
      tls: true
      valid_tls_alert_codes: [40, 42, 116]
      preferred_ip_protocol: ip4
  mtls_enforcement_http:
    prober: http
    http:
      valid_tls_alert_codes: [40, 42, 116]
      preferred_ip_protocol: ip4
~

When the connection fails with one of the acceptable TLS alerts the probe will succeed and can still try to validate other fields of the TLS connection such like the server certificate chain _(and add this data to the response)_
```

tchinchow:
Fixes prometheus/blackbox_exporter#1674
- **`tcp prober`**: add acceptable tls alert codes configuration parameter to TLS connections
- **` http prober`**: add acceptable tls alert codes configuration parameter to TLS connections

[FEATURE] Add a new configuration parameter (`valid_tls_alert_codes`) to the TCP and HTTP probes to list expected TLS alert codes when probing TLS connections. The probe then only succeeds if one of these alerts is received.

**Checklist**
- [x] Tests updated
- [x] Documentation added


tchinchow:
- Rebased on `master` branch
- Added Tests,
- Fixed bugs found when building the tests,
- Added documentation _(changelog + `example.yml`)_
- Updated CHANGELOG in the pull request text.

tchinchow:
Rebased on `upstream/master` branch

tchinchow:
Add test for inconsistent configurations and code to reject them _(tchinchow/blackbox_exporter#8)_.

